PR #7, opened 2024-08-01
change: ETCM-8051 setup-main-chain-state checks

AmbientTea:
# Description

* ~Display a better error message for Kupo and Ogmios connection error~ removed checks for Kupo and Ogmios connection. it will better be handled by the smart contracts CLI
* Check for existence of `sidechain-main-cli` before running

# Checklist

- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/partner-chains-cli/src/setup_main_chain_state/mod.rs
+++ b/partner-chains-cli/src/setup_main_chain_state/mod.rs
@@ -1,7 +1,7 @@
 use crate::config::config_fields::{CARDANO_PAYMENT_SIGNING_KEY_FILE, POSTGRES_CONNECTION_STRING};
 use crate::config::{
 	config_fields, ChainConfig, ConfigFieldDefinition, SidechainParams, CHAIN_CONFIG_FILE_PATH,
-	PARTNER_CHAINS_CLI_PATH,
+	SIDECHAIN_MAIN_CLI_PATH,
 };
 use crate::io::IOContext;
 use crate::permissioned_candidates::{ParsedPermissionedCandidatesKeys, PermissionedCandidateKeys};
@@ -95,9 +95,9 @@ impl CmdRun for SetupMainChainStateCmd {
 		context.print(
 			"This wizard will set or update D-Parameter and Permissioned Candidates on the main chain. Setting either of these costs ADA!",
 		);
-		if !context.file_exists(PARTNER_CHAINS_CLI_PATH) {
+		if !context.file_exists(SIDECHAIN_MAIN_CLI_PATH) {
 			return Err(anyhow!(
-				"Partner Chains CLI executable file ({PARTNER_CHAINS_CLI_PATH}) is missing."
+				"Partner Chains CLI executable file ({SIDECHAIN_MAIN_CLI_PATH}) is missing."
 			));
 		}
 		let config_initial_authorities =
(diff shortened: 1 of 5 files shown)


partner-chains-cli/src/setup_main_chain_state/mod.rs
```diff
@@ -94,6 +96,11 @@ impl CmdRun for SetupMainChainStateCmd {
 		context.print(
 			"This wizard will set or update D-Parameter and Permissioned Candidates on the main chain. Setting either of these costs ADA!",
 		);
+		if !context.file_exists(SIDECHAIN_MAIN_CLI_PATH) {
+			return Err(anyhow!(
+				"Partner Chains CLI executable file ({SIDECHAIN_MAIN_CLI_PATH}) is missing."
```
LGLO, line 101: Partner Chains CLI is partner-chains-cli. What is missing is smart contracts CLI. This message is confusing.

partner-chains-cli/src/smart_contracts.rs
```diff
@@ -24,3 +25,65 @@ pub fn runtime_config_arguments(
 			payment_signing_key_path
 	)
 }
+
+pub fn check_for_kupo_ogmios_connection_error(
+	response: &str,
+	sidechain_main_cli_resources: &SidechainMainCliResources,
+) -> anyhow::Result<()> {
+	if response.contains("ECONNREFUSED") {
```
LGLO, line 33: Unfortunately, this is wrong as well. Please test locally with recent sidechain-main-cli.

LGLO: This is an attempt to fix the error in a wrong place. Instead of us, it should be smart contracts cli to not throw out cryptic errors

partner-chains-cli/src/setup_main_chain_state/mod.rs
AmbientTea: fixed that

partner-chains-cli/src/smart_contracts.rs
AmbientTea: I've tested this earlier and retested it now and it correctly produces:
```
Error: Failed to connect to Ogmios at http://localhost:1337. Please check connection configuration and try again.
```
As for where this should be fixes - I agree that smart cotracts cli should improve its errors, but in the meantime we've had the requirement to provide a work-around for that. We can clean this up when the CLI improves its messaging

LGLO: I've tested prepare-configuration command with wrong ogmios port and this has not been triggered. sidechain-main-cli error does not contain ECONNREFUSED.

partner-chains-cli/src/smart_contracts.rs
```diff
@@ -24,3 +25,65 @@ pub fn runtime_config_arguments(
...
+		let kupo_port_str = &sidechain_main_cli_resources.kupo.port.to_string();
+		let (target, connection) = if response.contains(kupo_port_str) {
```
LGLO, line 35: Edge-case when it doesn't work: both services are at same port but different hosts.

AmbientTea: You're right. Unfortunately, there's nothing we can do, since the host printed by the CLI is not necessarily the same as the user provided (it looks like it resolves the host name to IP).

partner-chains-cli/src/smart_contracts.rs
AmbientTea: @LGLO Sorry, I can't reproduce it on my machine, I get the correct error for both prepare-configuration and setup-main-chain-state. But I do know that sometimes the error message from CLI is different (because the error returned is wrapped somehow), so maybe that's not determinate?

partner-chains-cli/src/smart_contracts.rs
AmbientTea: We've decided to scrap this check alltogether

partner-chains-cli/src/smart_contracts.rs
AmbientTea: We've decided to scrap this check alltogether

mpskowron (approved):

LGLO (approved):

PR #15, opened 2024-08-06
refactor: make cli commands independent of RPC

LGLO:
# Description

I planned to move as much code from `node` as possible, because it requires copy-pasting by users - it is under GPLv3 and we cannot use it in Apache2 libraries. I'm not happy with the amount of code moved, I would happily receive any help to be able to move more.

The other objective for this PR was to not depend on RPC in commands, so refactoring was made to extract "query" API to sp-session-validator-management, but unfortunately I had to make additional crate, due to circular dependencies with authority-selection-inputs.

RPC layer is now very thin and trivial. I've moved all tests to "query" create, leaving this very thin layer w/o tests now.

Manually tested two commands and two RPC methods that are subjects of this refactoring.

# Checklist

- [x] Self-reviewed the diff

# Note on CI
for you.


node/src/command/mod.rs
```diff
@@ -300,3 +311,12 @@ fn initialize_chain_spec(runner: &Runner<Cli>) -> Result<Box<dyn sc_service::Cha
 		})
 		.map_err(|_| "Initialization process terminated by user".into())
 }
+
+async fn print_result<FIn>(command_future: FIn) -> Result<(), sc_cli::Error>
```
LGLO, line 315: This replaces former `run_ignore_result` and `println!` in `run`.

primitives/session-validator-management/query/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+	C::Api: SessionValidatorManagementApi<
+		Block,
+		SessionKeys,
+		CrossChainPublic,
+		AuthoritySelectionInputs,
+		ScEpochNumber,
+	>,
+	C::Api: GetSidechainParams<Block, SidechainParams>,
+	C::Api: CandidateValidationApi<Block>,
+{
```
LGLO, line 92: Methods in this implementation are like methods that were in RPC, only errors are mapped to strings instead of jsonrpsee ErrorObjects.

primitives/session-validator-management/query/src/lib.rs
```diff
@@ -0,0 +1,211 @@
...
+use sp_runtime::traits::NumberFor;
+use sp_runtime::traits::{Block as BlockT, Zero};
+use sp_session_validator_management::SessionValidatorManagementApi;
+use sp_sidechain::{GetSidechainParams, GetSidechainStatus};
+use std::sync::Arc;
+use types::*;
+
+pub type QueryResult<T> = Result<T, String>;
+
+#[async_trait]
```
mpskowron, line 26: If you have time and want to play with it, since rust 1.75 async fn can be used in trait without the annotation. There are some differences though and I haven't played with it yet: https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html

mpskowron (approved):

vvviiimmm (approved):

primitives/session-validator-management/query/src/lib.rs
LGLO: I am able to replace:
```
#[async_trait]
pub trait SessionValidatorManagementQueryApi {
```
with
```
#[trait_variant::make(SessionValidatorManagementQueryApi: Send)]
pub trait LocalSessionValidatorManagementQueryApi {
```
I'm not sure if it worth doing now. Pros: we'll use "more modern way" and will have some example for future generations in the code. Cons: we need to stick with `#[async_trait]` in the next layer, because traits defined with `#[rpc(...)]` require implementations to use `#[async_trait]` annotation anyway.

mpskowron: I think that having such example in the code would be nice, up to you though.

LGLO: I'll do it in a separate PR in few minutes!

PR #17, opened 2024-08-06
change: change McBlockHash inner type from Vec to an array

AmbientTea:
# Description

Changed the inner type of `McBlockHash` from variable-length `Vec<u8>` to constant length array `[u8; 32]`.

# Checklist

- [x] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/primitives/sidechain-mc-hash/src/lib.rs
+++ b/primitives/sidechain-mc-hash/src/lib.rs
@@ -191,7 +191,7 @@ impl InherentDigest for McHashInherentDigest {
 			if let DigestItem::PreRuntime(id, data) = item {
 				if *id == MC_HASH_DIGEST_ID {
 					let data = data.clone().try_into().map_err(|_| {
-						format!("Invalid MC hash in digest: {:?}", ByteString(data.to_vec()))
+						format!("Invalid MC hash referenced by block author in digest: {:?}\nMC hash must be exactly 32 bytes long.", ByteString(data.to_vec()))
 					})?;
 					return Ok(McBlockHash(data));
 				}
--- a/primitives/sidechain-mc-hash/src/lib.rs
+++ b/primitives/sidechain-mc-hash/src/lib.rs
@@ -2,7 +2,7 @@ use crate::McHashInherentError::StableBlockNotFound;
 use main_chain_follower_api::{
 	block::MainchainBlock, common::Timestamp as McTimestamp, BlockDataSource, DataSourceError,
 };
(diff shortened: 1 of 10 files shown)


LGLO:
Please describe this PR

LGLO (changes requested):
Doesn't work for me.
Have you actually tested it by running at least two local nodes?

primitives/sidechain-mc-hash/src/lib.rs
```diff
@@ -189,8 +189,10 @@ impl InherentDigest for McHashInherentDigest {
 	) -> Result<Self::Value, Box<dyn Error + Send + Sync>> {
 		for item in digest {
 			if let DigestItem::PreRuntime(id, data) = item {
+				let data =
+					data.clone().try_into().map_err(|_| "Invalid MC hash in digest: {data}")?;
```
LGLO, line 193: The error message then contains: `Invalid MC hash in digest: {data}`

AmbientTea:
> Doesn't work for me. Have you actually tested it by running at least two local nodes?

Thanks for noticing, I only run a single node. I fixed the issue and covered it with a test

primitives/sidechain-mc-hash/src/lib.rs
AmbientTea: fixed

primitives/sidechain-mc-hash/src/lib.rs
```diff
@@ -190,7 +190,11 @@ impl InherentDigest for McHashInherentDigest {
 		for item in digest {
 			if let DigestItem::PreRuntime(id, data) = item {
 				if *id == MC_HASH_DIGEST_ID {
-					return Ok(McBlockHash(data.clone()));
+					let data = data
+						.clone()
+						.try_into()
+						.map_err(|_| format!("Invalid MC hash in digest: {data:?}"))?;
```
LGLO, line 196: This time it looks like this (if digest is wrong):
`Invalid MC hash in digest: [91, 223, 29, 17, 0, 0, 0, 0]`. Why not hex, like we should do for all bytes printing?

LGLO: I think it's also a programmatic error, to have something else than 32 bytes here. It is not exceptional situation. It would be nice to add that `mcsh` has to contain 32 bytes and block creator put syntactically wrong value there.

LGLO (approved):

mpskowron (approved):

PR #22, opened 2024-08-08
change: ETCM-7811 native token movement observability

AmbientTea:
# Description

‼️ This PR is here to show all the changes together and is not intended to be merged. Please review the partial PRs (when undrafted):
1. [data source](https://github.com/input-output-hk/partner-chains/pull/40)
2. [inherent data provider](https://github.com/input-output-hk/partner-chains/pull/37)
3. [pallet](https://github.com/input-output-hk/partner-chains/pull/36)

A working implementation of [the native token observability design](https://miro.com/app/board/uXjVKs3q69w=/?moveToWidget=3458764596872450048&cot=14) that I shared a couple weeks ago. It:
1. observes all transactions that put utxos containing native tokens (identified by policy and asset name) into the illiquid supply validator
 (identified by an address)
2. creates an inherent data with the total count of newly "locked" tokens since the last mc hash (mc hash referenced by the parent block)
3. consumes the inherent data in the `native-token-management` pallet and runs an inherent that triggers a hook provided by the PC dev as part of the pallet's configuration.
4. The current implementation of the hook credits a dummy account (`0xaa...aa`) with the normal substrate native token, so there's some state that can be observed by SDET tests etc.

I tested this implementation manually and it works correctly, but I've used addresses provided by Radek, which unfortunately have only one utxo with the native tokens.

# Checklist

- [x] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,8 +31,9 @@ members = [
 	"primitives/session-validator-management/query",
     "primitives/session-manager",
     "primitives/sidechain",
-	"partner-chains-cli"
-]
+	"partner-chains-cli",
+    "pallets/native-token-management",
+    "primitives/native-token-management"]
 resolver = "2"
 
 [profile.release]
@@ -188,3 +189,5 @@ authority-selection-inherents = { path = "primitives/authority-selection-inheren
 session-manager = { path = "primitives/session-manager", default-features = false }
 sp-sidechain = { path = "primitives/sidechain", default-features = false }
 chain-params = { path = "primitives/chain-params", default-features = false }
+pallet-native-token-management = { path = "pallets/native-token-management", default-features = false }
+sp-native-token-management = { path = "primitives/native-token-management", default-features = false }
--- a/mainchain-follower/db-sync-follower/src/db_model.rs
+++ b/mainchain-follower/db-sync-follower/src/db_model.rs
@@ -82,6 +82,12 @@ impl Asset {
 #[derive(Debug, Clone, sqlx::FromRow, PartialEq)]
 pub(crate) struct AssetName(pub Vec<u8>);
 
+impl From<sidechain_domain::AssetName> for AssetName {
+	fn from(name: sidechain_domain::AssetName) -> Self {
+		Self(name.0.to_vec())
+	}
+}
+
 #[derive(Debug, Clone, sqlx::FromRow, PartialEq)]
 pub(crate) struct DistributedSetData {
 	pub utxo_id_tx_hash: [u8; 32],
@@ -198,6 +204,12 @@ pub(crate) struct MintAction {
 #[derive(Debug, Clone, sqlx::FromRow, PartialEq)]
 pub(crate) struct PolicyId(pub Vec<u8>);
 
+impl From<sidechain_domain::PolicyId> for PolicyId {
+	fn from(id: sidechain_domain::PolicyId) -> Self {
+		Self(id.0.to_vec())
+	}
+}
+
 #[derive(Debug, Clone, sqlx::FromRow, PartialEq)]
 pub(crate) struct StakePoolEntry {
 	pub pool_hash: [u8; 28],
@@ -256,6 +268,28 @@ impl<'r> Decode<'r, Postgres> for StakeDelegation {
 	fn decode(value: <Postgres as HasValueRef<'r>>::ValueRef) -> Result<Self, BoxDynError> {
 		let decoded = <sqlx::types::BigDecimal as Decode<Postgres>>::decode(value)?;
 		let i = decoded.to_u64().ok_or("StakeDelegation is always a u64".to_string())?;
+		Ok(Self(i))
+	}
+}
+
+#[derive(Debug, Clone, PartialEq)]
+pub(crate) struct NativeTokenAmount(pub u64);
+impl From<NativeTokenAmount> for sidechain_domain::NativeTokenAmount {
+	fn from(value: NativeTokenAmount) -> Self {
+		Self(value.0)
+	}
+}
+
+impl sqlx::Type<Postgres> for NativeTokenAmount {
+	fn type_info() -> <Postgres as sqlx::Database>::TypeInfo {
+		PgTypeInfo::with_name("NUMERIC")
+	}
+}
+
+impl<'r> Decode<'r, Postgres> for NativeTokenAmount {
+	fn decode(value: <Postgres as HasValueRef<'r>>::ValueRef) -> Result<Self, BoxDynError> {
+		let decoded = <sqlx::types::BigDecimal as Decode<Postgres>>::decode(value)?;
+		let i = decoded.to_u64().ok_or("NativeTokenQuantity is always a u64".to_string())?;
 		Ok(Self(i))
 	}
 }
@@ -352,7 +386,7 @@ LIMIT 1",
 }
 
 /// Query to get the block by its hash
-#[cfg(feature = "block-source")]
+#[cfg(any(feature = "block-source", feature = "native-token"))]
 pub(crate) async fn get_block_by_hash(
 	pool: &Pool<Postgres>,
 	hash: McBlockHash,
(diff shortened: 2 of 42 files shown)


mainchain-follower/db-sync-follower/src/db_model.rs
```diff
@@ -527,3 +527,36 @@ pub(crate) async fn index_exists(pool: &Pool<Postgres>, index_name: &str) -> boo
...
+		"
+SELECT
+    SUM(ma_tx_out.quantity)
+FROM tx_out
+LEFT JOIN ma_tx_out    ON ma_tx_out.tx_out_id = tx_out.id
+LEFT JOIN multi_asset  ON multi_asset.id = ma_tx_out.ident
+INNER JOIN tx          ON tx_out.tx_id = tx.id
+INNER JOIN block       ON tx.block_id = block.id
+WHERE address = $1
+AND multi_asset.policy = $2
```
AmbientTea, line 552: I think this should also use the token symbol in the query for correctness, but I'd like someone to confirm?

mainchain-follower/db-sync-follower/src/db_model.rs
```diff
@@ -527,3 +527,36 @@ pub(crate) async fn index_exists(pool: &Pool<Postgres>, index_name: &str) -> boo
 		.map(|rows| rows.len() == 1)
 		.unwrap()
 }
+
+#[derive(Debug, Clone, sqlx::FromRow, PartialEq)]
+pub(crate) struct ReleasedTokens(pub Option<StakeDelegation>);
```
AmbientTea, line 532: I used `StakeDelegation` here because it represents a number of tokens, but obviously we'll need either some new type or rename `StakeDelegation` to something universal like `Lovelace`

mainchain-follower/main-chain-follower-api/src/native_token.rs
```diff
@@ -0,0 +1,19 @@
...
+use async_trait::async_trait;
+use sidechain_domain::*;
+
+pub struct NativeTokenManagementReleaseEvent {
+	pub token_amount: u64,
+}
+
+#[async_trait]
+pub trait NativeTokenManagementDataSource {
+	/// Retrieves all native token releases into the illiquid supply in the range (after_block, to_block]
```
AmbientTea, line 11: There's no metadata attached to the illiquid supply utxos, so we can just query for the sum of their tokens

runtime/src/lib.rs
```diff
@@ -317,6 +321,21 @@ impl frame_system::Config for Runtime {
 	type PostTransactions = ();
 }
 
+pub struct TokenReleaseHandler;
+
+impl pallet_native_token_management::TokenReleaseHandler<u128> for TokenReleaseHandler {
```
AmbientTea, line 326: Just a mock handler that logs the data. It will be MN's responsibility to handle these events in their own ledger

mainchain-follower/db-sync-follower/src/db_model.rs
AmbientTea, line 552: I know this query will also need to use the token name here for full correctness

Cargo.toml
```diff
@@ -188,3 +189,5 @@ authority-selection-inherents = { path = "primitives/authority-selection-inheren
 session-manager = { path = "primitives/session-manager", default-features = false }
 sp-sidechain = { path = "primitives/sidechain", default-features = false }
 chain-params = { path = "primitives/chain-params", default-features = false }
+pallet-native-token-management = { path = "pallets/native-token-management", default-features = false }
+sp-native-token-management = { path = "primitives/native-token-management", default-features = false }
```
LGLO, line 193: "native token" is very broad term but this pallet and associate primitives are very specific. I don't have a better candidate for the name currently, but maybe you have.

Cargo.toml
```diff
@@ -188,3 +189,5 @@ authority-selection-inherents = { path = "primitives/authority-selection-inheren
 session-manager = { path = "primitives/session-manager", default-features = false }
 sp-sidechain = { path = "primitives/sidechain", default-features = false }
 chain-params = { path = "primitives/chain-params", default-features = false }
+pallet-native-token-management = { path = "pallets/native-token-management", default-features = false }
```
LGLO, line 192: Design mentioned in the PR description has two pallets: "native token transfer pallet" and "native token management pallet". To be honest I think they are meant to be so related that I think having one pallet is better but still this is discrepant to design = mess.

mainchain-follower/db-sync-follower/src/db_model.rs
LGLO: Please add at least an empty name, so db index is used

mainchain-follower/db-sync-follower/src/db_model.rs
LGLO: Why would you use `Lovelace` or same type as `StakeDelegation` for native tokens? They are different types even by allowed range as far as I know.

mainchain-follower/db-sync-follower/src/native_token/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+use main_chain_follower_api::Result;
+use sidechain_domain::*;
+use sqlx::PgPool;
+
+pub struct NativeTokenManagementDataSourceImpl {
+	pub pool: PgPool,
+	pub metrics_opt: Option<McFollowerMetrics>,
+}
+
+#[async_trait]
```
LGLO, line 17: Was there any particular problem with `observed_async_trait` ? If so, it's a good indicator to reimplement `observed_async_trait` without `async_trait` if possible.

mainchain-follower/db-sync-follower/src/db_model.rs
```diff
@@ -527,3 +527,36 @@ pub(crate) async fn index_exists(pool: &Pool<Postgres>, index_name: &str) -> boo
...
+
+#[cfg(feature = "native-token")]
+pub(crate) async fn get_native_token_releases(
```
LGLO, line 535: It doesn't return "releases" but "sum of release" tokens. Anyway I think that db query is far from business logic and the name should reflect what it does. At Cardano level there are no "releases".

This query sums value of specific tokens transfered to specific address.

mainchain-follower/db-sync-follower/src/native_token/mod.rs
```diff
@@ -0,0 +1,58 @@
...
+			&self.pool,
+			after_slot,
+			to_slot,
+			crate::db_model::PolicyId(native_token_policy.0.to_vec()),
+			Address(illiquid_supply_address.to_string()),
+		)
+		.await?;
+
+		Ok(maybe_release.0.map(|release| NativeTokenManagementReleaseEvent {
+			token_amount: release.0.try_into().expect("i64 not a valid u64"),
```
LGLO, line 55: Aren't both u64 in this code?

Cargo.toml
AmbientTea: Yes, they are extremely similar and I need to make sure we really need to distinguish between those cases (release from reserve vs. someone just sending tokens to the illiquid supply - they all end up in the same place)

mainchain-follower/db-sync-follower/src/db_model.rs
AmbientTea: because I'm a dumbass and primed myself to think it makes sense

Cargo.toml
AmbientTea: Maybe "{sp,pallet}-cross-chain-liquidity"? I don't have a solid one either.

mainchain-follower/db-sync-follower/src/native_token/mod.rs
AmbientTea: I dropped it, because it messed with my auto-completion, and didn't bring it back. I'll re-add it to verify it works

AmbientTea: Added it and it works

mainchain-follower/db-sync-follower/src/db_model.rs
AmbientTea: fixed that silliness

mainchain-follower/db-sync-follower/src/db_model.rs
AmbientTea: Fixed it to use both properly

mainchain-follower/db-sync-follower/src/db_model.rs
AmbientTea: renamed

Cargo.toml
AmbientTea: I've gone through the requirements document and concluded there's nothing distinguishing the two types of event from the observability perspective (they only differ on the MC). I've modified the design in Miro to use a single pallet like this PR does.

PR #39, opened 2024-08-21
fix: artifacts workflow

skylar-simoncelli:
I have made a few opinionated changes to my artifact workflow:

- No longer creates a single zip at all, preventing the double-zip issue (Github zips all downloads regardless) and hopefully influencing future releases to not upload a single zip either
- Properly names artifacts appending architecture and optionally inserting a release tag (have tested with and without)
- Deletes the original PCSC artifact before the rename (so nobody accidentally attaches to release)

Succesful run with tags:
https://github.com/input-output-hk/partner-chains/actions/runs/10489847169

And without tags:
https://github.com/input-output-hk/partner-chains/actions/runs/10489849955


.github/workflows/artifacts.yml
```diff
@@ -0,0 +1,247 @@
+name: Build and Upload Artifacts
+
+on:
+  workflow_dispatch:
+    inputs:
+      partner_chains_smart_contracts_sha:
+        description: "partner-chains-smart-contracts commit SHA"
+        default: "38f4c17e0f4052d885c9d438f7fb8ad0b2e7b25e"
```
LGLO, line 8: Optional.
Would "v6.2.0" work here? Or maybe some dummy value to not make this revision anything special?
With "v6.2.0" one could easily scan repo for updates when changing flake.nix.

skylar-simoncelli: This default value is nothing important, it is just the value that is pre-filled in the textbox

I can change to "master", or I can remove entirely

skylar-simoncelli: v6.2.0 would work if there is a branch named v6.2.0

skylar-simoncelli: I have removed the default values entirely. User should supply a SHA or a branch when building artifacts

.github/workflows/artifacts.yml
```diff
@@ -0,0 +1,245 @@
...
+
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.inputs.partner_chains_sha }}
+
+      - name: Install Rust toolchain
+        uses: actions-rs/toolchain@v1
+        with:
+          toolchain: stable
```
LGLO, line 40: I'm afraid it could backfire at some point, because we have toolchain defined implicitly in Earthfile and explicitly in rust-toolchain.yaml. There is a potential for a hard to understand bug when locally and earthly-ci would builds the code but this action fails due to stricter compiler behavior.

It is not an unlikely scenario. Just yesterday Michael reported that polkadot-sdk fails to build with rust 1.80 but is okay with 1.77.

.github/workflows/artifacts.yml
```diff
@@ -0,0 +1,245 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          repository: input-output-hk/partner-chains-smart-contracts
+          token: ${{ secrets.ACTIONS_PAT }}
+          ref: ${{ github.event.inputs.partner_chains_smart_contracts_sha }}
+          path: partner-chains-smart-contracts
+
+      - name: Build
+        run: nix build ./partner-chains-smart-contracts#sidechain-release-bundle
```
LGLO, line 201: Oh, please no. We should just download artifact from the other repo at most. Why do we build it here?

.github/workflows/artifacts.yml
skylar-simoncelli: Ah my bad! (not a Rust developer)

It seems cargo builds automatically use the `rust-toolchain.toml` if it exists, so I have removed those steps entirely

.github/workflows/artifacts.yml
skylar-simoncelli: I agree, removed

PR #44, opened 2024-08-22
change: Add a set_main_chain_scripts extrinsic to validator-management pallet

AmbientTea:
# Description

Just an extrinsic to change the MC addresses for registrations/d-param, so we can easily migrate whenever these change due to PCSC upgrade.

# Checklist

- [x] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5800,6 +5800,7 @@ dependencies = [
  "log",
  "parity-scale-codec",
  "scale-info",
+ "sidechain-domain",
  "sp-core",
  "sp-io",
  "sp-runtime",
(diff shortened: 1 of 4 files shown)


LGLO:
Have you tested it via polkadot-ui?

mpskowron:
Have you tested if syncing works after someone changes the scripts (and there was at least one rotation from new scripts)?

LGLO (approved):

AmbientTea:
> Have you tested if syncing works after root user changes the scripts (and there was at least one rotation from new scripts)?

I'm having troubles running two-node setup on my machine, but I think someone did a change like that with raw storage write. I validated that it writes the data correctly, so it should work as well.

PR #53, opened 2024-08-26
feat: remove benchmarking from earthly.yml

skylar-simoncelli:


AmbientTea (commented):
Why are we removing it?

PR #57, opened 2024-08-27
fix: install cargo-chef unconditionally - ETCM-7736

LGLO:
This is my attempt to fix, what I've reported in June: https://input-output.atlassian.net/browse/ETCM-7736

Before this change on a clean setup, after `earthly prune` and removing all Earthly docker volumes, when running Earthly:
```
              +setup | --> RUN command -v cargo-chef &> /dev/null || cargo install --locked cargo-chef
```
does not install cargo-chef and build fails with:
```
================================== ❌ FAILURE ===================================

         +build-deps *failed* | Repeating the failure error...
         +build-deps *failed* | --> RUN cargo --locked chef prepare
         +build-deps *failed* | error: no such command: `chef`

         +build-deps *failed* |         Did you mean `check`?

         +build-deps *failed* |         View all installed commands with `cargo --list`
         +build-deps *failed* |         Find a package to install `chef` with `cargo search cargo-chef`
         +build-deps *failed* | ERROR Earthfile line 39:2
         +build-deps *failed* |       The command
         +build-deps *failed* |           RUN cargo --locked chef prepare
         +build-deps *failed* |       did not complete successfully. Exit code 101
```

So I remove the condition and then it works, it installs cargo-chef, and continues the build.


skylar-simoncelli (approved):

nrdxp:
@LGLO, looks good. One thing that will help you avoid this situation in the future as well.

It's possible for the persistent caches to get cleared, while earthly still thinks that the specific layer is "cached", which means it will skip it. Since this step installs the cargo-chef binary into a persistent cache, if it gets cleared while this step is still considered cached, it will not get reinstalled (probably what happened). I'm considering reporting this behavior upstream, as it seems non-sensical to consider a step "cached" if the cache it popullates has been wiped.

In any case, if you do hit that situation again for any reason, you can simply modify this step trivially (and an extra space). Earthly calculates the cache key based on a hash of the command itself, so any modification will invalidate the cache and force the step to re-execute. so:
```
cargo install cargo-chef
# the extra space changes the cache key
cargo install  cargo-chef
```

You don't actually have to commit a change like this, after it is forcibly ran, the underlying persistent cache should be repopullated.

nrdxp (approved):

PR #59, opened 2024-08-27
feat: add tests to not-earthly and fix ECR upload

skylar-simoncelli:


LGLO:
I cannot see `cargo fmt` check in the workflow. It is present in Earthly.

PR #63, opened 2024-08-28
change: ETCM-8262 move reading of config env vars out of the node crate

AmbientTea:
# Description

Moves logic that reads from the environment:
- sidechain params
- selection MC scripts
- native token MC scripts

to their respective crates

# Checklist

- [x] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/primitives/chain-params/src/lib.rs
+++ b/primitives/chain-params/src/lib.rs
@@ -49,6 +49,19 @@ pub fn default_denominator() -> u64 {
 #[cfg(feature = "std")]
 impl SidechainParams {
 	pub fn read_from_env_with_defaults() -> Result<Self, envy::Error> {
+		/// This structure is needed to read sidechain params from the environment variables because the main
+		/// type uses `rename_all = "camelCase"` serde option
+		#[cfg_attr(feature = "serde", derive(Serialize, Deserialize))]
+		struct SidechainParamsEnvConfiguration {
+			#[serde(default = "default_chain_id")]
+			pub chain_id: u16,
+			pub genesis_committee_utxo: UtxoId,
+			#[serde(default = "default_numerator")]
+			pub threshold_numerator: u64,
+			#[serde(default = "default_denominator")]
+			pub threshold_denominator: u64,
+			pub governance_authority: MainchainAddressHash,
+		}
 		let raw = envy::from_env::<SidechainParamsEnvConfiguration>()?;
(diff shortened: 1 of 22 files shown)


utils/configuration/Cargo.toml
```diff
@@ -0,0 +1,8 @@
+[package]
+name = "node-config-utils"
+version = "0.1.0"
+edition = "2021"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
LGLO, line 6: Please remove it

LGLO:
I suggest not adding this util crate at all. Use `envy`. Keep variables names, so some `xxx_policy` would need to be renamed to `xxx_policy_id`. Some `serde(default="default_chain_id"))` would need to be added.
The thing I have not checked in my local sources is difference between error types (I suggest replacing our custom error with envy error).

LGLO (approved):
I am happy to see code movement from `node` to libraries!

primitives/chain-params/src/lib.rs
```diff
@@ -35,3 +35,40 @@ pub struct SidechainParams {
...
 	pub governance_authority: MainchainAddressHash,
 }
+
+pub fn default_chain_id() -> u16 {
+	1
+}
+pub fn default_numerator() -> u64 {
+	2
+}
+pub fn default_denominator() -> u64 {
```
mpskowron, line 45: what do you think of using consts instead of functions?

primitives/chain-params/src/lib.rs
```diff
@@ -35,3 +35,40 @@ pub struct SidechainParams {
...
+			#[serde(default = "default_chain_id")]
+			pub chain_id: u16,
+			pub genesis_committee_utxo: UtxoId,
+			#[serde(default = "default_numerator")]
+			pub threshold_numerator: u64,
+			#[serde(default = "default_denominator")]
+			pub threshold_denominator: u64,
+			pub governance_authority: MainchainAddressHash,
+		}
+		let raw = envy::from_env::<SidechainParamsEnvConfiguration>()?;
```
mpskowron, line 65: Does it log the fact that the default value was used? It might be helpful when debugging configuration related issues

AmbientTea: It doesn't, unfortunately

PR #64, opened 2024-08-28
feat: adds `sidechain_main_cli` into config file instead

falcucci:
As we have the [sidechain-main-cli](https://github.com/input-output-hk/partner-chains-smart-contracts) binary as an external dependency, would be easier for developers/users to load its binary path from the `partner-chains-cli-resources-config.json` file so the user can have it installed in their PATH and update them only if needed but keeping the default one as `./sidechain-main-cli` just in case.


LGLO:
@falcucci Thank you for your contribution!

Having the path to smart-contracts CLI (sorry for the current cryptic name) hardcoded in partner-chains-cli tool was a decision, it was not just because it is easier to implement. The biggest factor to make this decision was to make it hard to use versions of `sidechain-main-cli` that are incompatible with `partner-chains-cli`. Please give us some time to revisit this decision.

falcucci:
@LGLO hey, thank you for the explanation.

that's fine this is a take it or leave it change, once I am using for personal purposes during development part of the walk-through setup of the partner-chain.

Anyway I'll keep this up to date with the upstream changes.

AmbientTea:
We're in the process of completely decoupling the `partner-chains-cli` from the `sidechain-main-cli` completely by rewriting the transaction building code in Rust, so I am closing this PR. Nevertheless, thank you @falcucci for this contribution!

PR #67, opened 2024-08-30
[Minotaur] Refactoring of Ariadne towards handing ETH stake 

aslesarenko:
# Description

In this PR: 
- refactoring necessary to add ETH stake registrations into virtual stake computation and computing combined list of valid trustless candidates from both Cardano and Ethereum. 
- No tests (and test vectors) has been changed (other than adapted to new data structures, and added default values).
- All tests pass
- Main goal is to extend DParameter, CandidateRegistrations, RegisteredData so that [select_authorities](https://github.com/aslesarenko/partner-chains/blob/8530b780bb76709cebf62dd77630eb3fd7ad4f1e/primitives/authority-selection-inherents/src/select_authorities.rs#L33) have all the necessary information. All the other changes are motivated by this.
- added `// TODO ETH: ...` in places where further work is needed to hangle stake from Ethereum
- no changes in db, follower, inherents, storage etc.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/mainchain-follower/db-sync-follower/src/candidates/mod.rs
+++ b/mainchain-follower/db-sync-follower/src/candidates/mod.rs
@@ -219,8 +219,8 @@ impl CandidatesDataSourceImpl {
 		candidates.into_iter().into_group_map_by(|c| c.mainchain_pub_key.clone())
 	}
 
-	fn make_registration_data(c: RegisteredCandidate) -> AdaRegistrationData {
-		AdaRegistrationData {
+	fn make_registration_data(c: RegisteredCandidate) -> CardanoRegistrationData {
+		CardanoRegistrationData {
 			consumed_input: c.consumed_input,
 			sidechain_signature: c.sidechain_signature,
 			mainchain_signature: c.mainchain_signature,
--- a/mainchain-follower/db-sync-follower/src/candidates/mod.rs
+++ b/mainchain-follower/db-sync-follower/src/candidates/mod.rs
@@ -164,7 +164,7 @@ impl CandidateDataSource for CandidatesDataSourceImpl {
 						eth_registrations: vec![]  // TODO ETH: add ETH registrations
 					},
 				stake_delegation: Self::get_stake_delegation(&stake_map, &mainchain_pub_key)
-					.map(|stake| StakeDelegation { ada: stake, eth: StakeAmount::zero() }),
+					.map(|stake| StakeDelegation { ada: stake, eth: EthStakeAmount::zero() }),
 			}
 		}).collect())
 	}
--- a/mainchain-follower/db-sync-follower/src/candidates/mod.rs
+++ b/mainchain-follower/db-sync-follower/src/candidates/mod.rs
@@ -144,7 +144,10 @@ impl CandidateDataSource for CandidatesDataSourceImpl {
 		Ok(Self::group_candidates_by_mc_pub_key(candidates).into_iter().map(|(mainchain_pub_key, candidate_registrations)| {
 			CandidateRegistrations {
 				mainchain_pub_key: mainchain_pub_key.clone(),
-				registrations: candidate_registrations.into_iter().map(Self::make_registration_data).collect(),
+				registrations: Registrations::Ada(candidate_registrations
+					.into_iter()
+					.map(Self::make_registration_data)
+					.collect()),
 				stake_delegation: Self::get_stake_delegation(&stake_map, &mainchain_pub_key),
 			}
 		}).collect())
@@ -200,8 +203,8 @@ impl CandidatesDataSourceImpl {
 		candidates.into_iter().into_group_map_by(|c| c.mainchain_pub_key.clone())
 	}
 
-	fn make_registration_data(c: RegisteredCandidate) -> RegistrationData {
-		RegistrationData {
+	fn make_registration_data(c: RegisteredCandidate) -> AdaRegistrationData {
+		AdaRegistrationData {
 			consumed_input: c.consumed_input,
 			sidechain_signature: c.sidechain_signature,
 			mainchain_signature: c.mainchain_signature,
--- a/mainchain-follower/db-sync-follower/src/candidates/mod.rs
+++ b/mainchain-follower/db-sync-follower/src/candidates/mod.rs
@@ -147,18 +147,19 @@ impl CandidateDataSource for CandidatesDataSourceImpl {
 			.get_registered_candidates(epoch, committee_candidate_address)
 			.await?;
 
-		// TODO: ETH: get registered candidates for ETH
+		// TODO ETH: get registered candidates for ETH
 
 		let stake_map = Self::make_stake_map(db_model::get_stake_distribution(&self.pool, epoch).await?);
 		Ok(Self::group_candidates_by_mc_pub_key(ada_candidates).into_iter().map(|(mainchain_pub_key, candidate_registrations)| {
 			CandidateRegistrations {
 				mainchain_pub_key: mainchain_pub_key.clone(),
+				eth_pub_key: None,  // TODO ETH: add ETH pub key
 				registrations: Registrations {
 						ada_registrations: candidate_registrations
 							.into_iter()
 							.map(Self::make_registration_data)
 							.collect(),
-						eth_registrations: vec![]  // TODO: ETH: add ETH registrations
+						eth_registrations: vec![]  // TODO ETH: add ETH registrations
 					},
 				stake_delegation: Self::get_stake_delegation(&stake_map, &mainchain_pub_key)
 					.map(|stake| StakeDelegation { ada: stake, eth: StakeAmount::zero() }),
--- a/mainchain-follower/db-sync-follower/src/candidates/mod.rs
+++ b/mainchain-follower/db-sync-follower/src/candidates/mod.rs
@@ -312,19 +312,19 @@ impl CandidatesDataSourceImpl {
 	// Datum decoders
 	fn decode_d_parameter_datum(datum: &Datum) -> Result<DParameter> {
 		let d_parameter = match datum {
-			ListDatum(items) => match items.first().zip(items.get(1)).zip(items.get(2)) {
-				Some(((IntegerDatum(p), IntegerDatum(a)), IntegerDatum(e))) => {
-					p.to_u16().zip(a.to_u16()).zip(e.to_u16()).map(|((p, a), e)| DParameter {
+			ListDatum(items) => match items.first().zip(items.get(1)) {
+				Some((IntegerDatum(p), IntegerDatum(t))) => {
+					p.to_u16().zip(t.to_u16()).map(|(p, t)| DParameter {
 						num_permissioned_candidates: p,
-						num_ada_candidates: a,
-						num_eth_candidates: e,
+						num_ada_candidates: t,
+						num_eth_candidates: 0,
 					})
-				}
+				},
 				_ => None,
 			},
 			_ => None,
 		}
-		.ok_or(DatumDecodeError { datum: datum.clone(), to: "DParameter".to_string() });
+			.ok_or(DatumDecodeError { datum: datum.clone(), to: "DParameter".to_string() });
 		if d_parameter.is_err() {
 			error!("Could not decode {:?} to DParameter. Expected [u16, u16].", datum.clone());
 		}
--- a/mainchain-follower/db-sync-follower/src/candidates/mod.rs
+++ b/mainchain-follower/db-sync-follower/src/candidates/mod.rs
@@ -134,7 +134,9 @@ impl CandidateDataSource for CandidatesDataSourceImpl {
 	}
 
 	// Get all registered candidate registrations for the given epoch.
-	// Note, this method requests only ADA candidates from Cardano and
+	// Note, this method requests only ADA candidates from Cardano.
+	// TODO ETH: From Minotaur perspective, there is a choice to make, whether CandidateDataSource
+	// is responsible to request only ADA candidates or ETH candidates as well.
 	async fn get_candidates(
 			&self,
 			epoch: McEpochNumber,
@@ -147,7 +149,7 @@ impl CandidateDataSource for CandidatesDataSourceImpl {
 			.get_registered_candidates(epoch, committee_candidate_address)
 			.await?;
 
-		// TODO ETH: get registered candidates for ETH
+		// TODO ETH: get registered candidates for ETH if CandidateDataSource is acting as unified facade
 
 		let stake_map = Self::make_stake_map(db_model::get_stake_distribution(&self.pool, epoch).await?);
 		Ok(Self::group_candidates_by_mc_pub_key(ada_candidates).into_iter().map(|(mainchain_pub_key, candidate_registrations)| {
--- a/mainchain-follower/db-sync-follower/src/candidates/mod.rs
+++ b/mainchain-follower/db-sync-follower/src/candidates/mod.rs
@@ -133,22 +133,35 @@ impl CandidateDataSource for CandidatesDataSourceImpl {
 		Ok(AriadneParameters { d_parameter: d_param, permissioned_candidates: candidates })
 	}
 
+	// Get all registered candidate registrations for the given epoch.
+	// Note, this method requests only ADA candidates from Cardano and
 	async fn get_candidates(
 			&self,
 			epoch: McEpochNumber,
 			committee_candidate_address: MainchainAddress
 		)-> Result<Vec<CandidateRegistrations>> {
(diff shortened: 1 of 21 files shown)


primitives/domain/src/lib.rs
```diff
@@ -187,11 +217,19 @@ pub const MAX_ASSET_NAME_LEN: u32 = 32;
 pub struct AssetName(pub BoundedVec<u8, ConstU32<MAX_ASSET_NAME_LEN>>);
 
 const MAINCHAIN_PUBLIC_KEY_LEN: usize = 32;
+const ETH_PUBLIC_KEY_LEN: usize = 32; // TODO ETH: is this correct length?
```
LGLO, line 220: Yes and no.
Story: historically this project started as Rust version of a Ethereum client that was able to observe Cardano. From this you can see SidechainPubKey. It is 33 bytes, because in this project we have always used the parity bit (hence all keys in hex start with 02 or 03). It is another complication for Minotaur, how to deal with existence of these keys in the project (which are user facing) and not make it very complicated for the user. Should already existing keys be used to get ETH stake address?

mainchain-follower/db-sync-follower/src/candidates/mod.rs
```diff
@@ -133,19 +133,36 @@ impl CandidateDataSource for CandidatesDataSourceImpl {
...
 		)-> Result<Vec<CandidateRegistrations>> {
 		let epoch = EpochNumber::from(self.get_epoch_of_data_storage(epoch)?);
-		let candidates = self.get_registered_candidates(epoch, committee_candidate_address).await?;
+
+		// get registered candidates for ADA
+		let ada_candidates = self
+			.get_registered_candidates(epoch, committee_candidate_address)
+			.await?;
+
+		// TODO ETH: get registered candidates for ETH
```
LGLO, line 150: D-param lives on Cardano, Ada candidates also live on Cardano. Cardano = Mainchain. We can rename Mainchain to Cardano (I still doubt we should), but we cannot have two Mainchains for sure. My point is that we perhaps should not get ETH candidates from Cardano (Mainnet)

primitives/domain/src/lib.rs
```diff
@@ -460,7 +518,7 @@ pub struct CommitteeHash(pub Vec<u8>);
 /// Note: A Registration Transaction is called by a user on Cardano to register themselves as a Sidechain Authority Candidate
 #[derive(Debug, Clone, Encode, Decode, PartialEq, Eq, TypeInfo)]
 #[cfg_attr(feature = "serde", derive(Serialize))]
-pub struct RegistrationData {
+pub struct AdaRegistrationData {
```
LGLO, line 521: If its on Cardano, then CardanoRegistrationData. Ada is only a token.

mainchain-follower/db-sync-follower/src/candidates/mod.rs
```diff
@@ -313,14 +330,15 @@ impl CandidatesDataSourceImpl {
 				Some((IntegerDatum(p), IntegerDatum(t))) => {
 					p.to_u16().zip(t.to_u16()).map(|(p, t)| DParameter {
 						num_permissioned_candidates: p,
-						num_registered_candidates: t,
+						num_ada_candidates: t,
+						num_eth_candidates: 0,
```
LGLO, line 334: Could this decoder handle situations when DParam on Cardano has 3 values and 2 values? Only in the latter case it would use 0 as default?

mpskowron (commented):
Is there any document with requirements/design/etc. or some jira task on which those changes are based?

mainchain-follower/db-sync-follower/src/candidates/mod.rs
aslesarenko: Yeah, somewhat misleading description, fixed. It make sense to continue calling Cardano as Mainchain since its role is special for governance. On the other hand, from virtual stake calculation point of view Cardano's role is symmetric to Ethereum.

mainchain-follower/db-sync-follower/src/candidates/mod.rs
aslesarenko: Yes, if we generalize a bit and declare it like this:
```rust
pub struct DParameter {
	pub num_permissioned_candidates: u16,
	pub num_ada_candidates: u16,
	pub num_other_candidates: Vec<u16>,  // one item for each chain
}

```
This is more universal encoding, so both empty vector and a vector with single element can be supported. It will also support for example adding PC native token to the virtual stake.

primitives/domain/src/lib.rs
aslesarenko: From Minotaur point of view any key is an opaque type with some trait(s) as constraints (e.g. `Key: MainchainKey`).
From user point of view, I would prefer to have clear distinction between Cardano, Ethereum and PC keys/addresses so that it is clear where each type of key should be used. 
For example, if I want to run PC node and restake both my ADA and ETH, then I already have and understand the difference between Cardano and Ethereum keys/addresses. Then it is very naturally for me to learn about PC keys/addresses, how they look like and how to create and use them. 
If SidechainPubKey has Ethereum legacy, this is probably has to be fixed sooner or later as technical debt.

primitives/domain/src/lib.rs
aslesarenko: I assumed this is only related to ADA staking. In the case we want to add Cardano's native tokens into Minotaur, will they have the same registration data?

aslesarenko:
@mpskowron, yes, replied in Slack.

primitives/domain/src/lib.rs
LGLO: I agree with fixing technical debt. I also do not want to add new one. I only try to raise awareness that:
* ECDSA public key can have different lengths
* For the "same lenght" one can one have version with and without a parity bit
* The parity bit is written as extra byte
* There are many ways of encoding a bit in a byte

This project already has one ECDSA public key in use. This public key is 33 bytes long. 32 bytes for a key and 1 byte that encodes parity bit as either 02 or 03.

This whole message is not trying me to forbid adding 32 bytes ECDSA public key w/o parity bit, I'm only saying that it will probably be confusing (added to already confusing Substrate layers over cryptography).

mainchain-follower/db-sync-follower/src/candidates/mod.rs
LGLO: :+1: My understanding is: for Minotaur there will be many sources of stake that are equal. But Cardano is the source of DParam. Is it correct?

primitives/domain/src/lib.rs
aslesarenko: Ah, that is important context which I didn't consider at all here. 
Then we can make it 33 bytes or even define it as
```rust 
pub struct EthPublicKey(pub Vec<u8>);
```
From Minotaur's perspective it is not important at all, so please let me know how it should be defined and I will change it.

mainchain-follower/db-sync-follower/src/candidates/mod.rs
aslesarenko: Yes, besides ETH, there are talks to add PC's native token. But governance will be done on Cardano, so it is to some degree the "main" chain.

mainchain-follower/db-sync-follower/src/candidates/mod.rs
vvviiimmm: just curious, will there be a new candidate selection algorithm?

vvviiimmm: (also if this new structure remains we should consider renaming it)

primitives/domain/src/lib.rs
vvviiimmm: imo `CardanoRegistrationData` should work for both ada and native tokens, while `AdaRegistrationData` is only for ada

mainchain-follower/db-sync-follower/src/candidates/mod.rs
aslesarenko: Yes, Minotaur "Mini", to replace Ariadne. Currently I assume that if num_other_candidates is empty, then it should behave as Ariadne. It may change in future, though.

aslesarenko: > (also if this new structure remains we should consider renaming it)

We need a good name, please suggest. 
People got used to DParameter and what it means, and the meaning will stay the same, maybe somewhat generalized.

primitives/authority-selection-inherents/src/authority_selection_inputs.rs
```diff
@@ -11,14 +11,33 @@ use {
...
 	pub permissioned_candidates: Vec<PermissionedCandidateData>,
+
+	// TODO ETH: The items in the Vec should be uniquely identified by sidechain_public_key
+	// (instead of  mainchain_public_key). This is necessary to allow one PC operator to register
+	// stake on different chains. The registration schema is symmetric with all chains and hence
+	// the same sidechain_public_key can be related to many public keys on different chains.
+	// Thus:
+	// - only one CandidateRegistrations per each candidate
+	// - each registered_candidates[i]:
+	//   - should contain valid mainchain_pub_key
```
vvviiimmm, line 31: can you please clarify which `mainchain_pub_key`, ethereum? If so, do we need optional `eth_pub_key` as suggested bellow?

primitives/domain/src/lib.rs
```diff
@@ -46,17 +46,47 @@ impl Display for McEpochNumber {
...
 #[derive(Default, Clone, Copy, Debug, Encode, Decode, TypeInfo, ToDatum, PartialEq, Eq)]
 #[cfg_attr(feature = "serde", derive(Serialize))]
-pub struct StakeDelegation(pub u64);
+pub struct StakeDelegation {
+	/// Amount of Lovelace (which is a fraction of 1 ADA) staked/locked on Cardano
+	pub ada: StakeAmount,
+
+	// TODO ETH: u64 if not enough to count stake on Ethereum in Wei
+	/// Amount of Wei (which is a fraction of 1 ETH) staked/locked on Ethereum
+	pub eth: StakeAmount,
```
vvviiimmm, line 58: for later: better to keep these as separate types

primitives/authority-selection-inherents/src/authority_selection_inputs.rs
```diff
@@ -11,14 +11,33 @@ use {
...
+    //   - may have optional eth_pub_key, in which case it may also have valid EthRegistrationData records
+
+	/// List of registered candidates from the main chain, one for each sidechain_public_key.
 	pub registered_candidates: Vec<CandidateRegistrations>,
+
+	/// The nonce for the epoch coming from the main chain.
 	pub epoch_nonce: EpochNonce,
+
+	// TODO ETH: add nonce from Ethereum chain
```
LGLO, line 40: I think that either we target 3 sources and hardcode it in this way, so we have permissioned_candidates, registered_cardano_candidates, registered_ethereum_candidates or this code should be generic for any number of sources, where each source has its "score" and each candidate inside this candidates source has it's score relative to other candidates of this type.

Said that I don't see Minotaur algorithm and I can only guess who it is supposed to be generalization of Ariadne.

mpskowron: I'm strongly for generalization here. I don't think that this part of the code has to know anything about the name of the source of candidates.

mainchain-follower/db-sync-follower/src/candidates/mod.rs
```diff
@@ -133,19 +133,38 @@ impl CandidateDataSource for CandidatesDataSourceImpl {
...
 				mainchain_pub_key: mainchain_pub_key.clone(),
-				registrations: candidate_registrations.into_iter().map(Self::make_registration_data).collect(),
-				stake_delegation: Self::get_stake_delegation(&stake_map, &mainchain_pub_key),
+				eth_pub_key: None,  // TODO ETH: add ETH pub key
+				registrations: Registrations {
+						ada_registrations: candidate_registrations
+							.into_iter()
+							.map(Self::make_registration_data)
+							.collect(),
+						eth_registrations: vec![]  // TODO ETH: add ETH registrations
```
mpskowron, line 164: Eth registrations will be observed on Ethereum, won't they? If so, I'd rather have a separate module for following Ethereum. This way, the code will be more modular and it will be easier for the users to pick up only the parts they are interested in.

primitives/authority-selection-inherents/src/authority_selection_inputs.rs
aslesarenko: clarified in the code comments.

primitives/domain/src/lib.rs
aslesarenko: Introduced EthStakeAmount type

aslesarenko:
@mpskowron, to your question [here](https://github.com/input-output-hk/partner-chains/pull/67#discussion_r1741027079), further generalisation is planned for the next PR (see my [message above](https://github.com/input-output-hk/partner-chains/pull/67#discussion_r1741094064))

mainchain-follower/db-sync-follower/src/candidates/mod.rs
aslesarenko: Yes, observed on Ethereum, but I would suggest to have 2 levels of data sources:
- level 1: data source which is passed everywhere (which is currently CandidateDataSource)
- level 2: one resource-specific data source for each resource type (CardanoDS, EthereumDS, etc.)

Thus defined CandidateDataSource will aggregate information from different sources.
From this POV, the next step would be to define a ResourceDS trait and move the Cardano-specific code there from CandidateDataSource.

primitives/authority-selection-inherents/src/authority_selection_inputs.rs
vvviiimmm: so in case of a valid eth candidate `mainchain_pub_key` == `Some(eth_pub_key)`?

aslesarenko: No, `mainchain_pub_key` is on Cardano, and `eth_pub_key` is on Ethereum, they can never be the same. The `eth_pub_key` field is optional, because `registered_candidates` vector should have only one item for each PC operator identified by `sidechain_public_key`, in other words, `registered_candidates` should have been `Map<SidechainPublicKey, CandidateRegistrations>`.

The reason I didn't introduced this map, is because I didn't want to deal with serialisation, but we can assume there is such mapping next to `registered_candidates`.

mainchain-follower/db-sync-follower/src/candidates/mod.rs
paul-weiss: we will need to consider cases where other chains are the "main chain" as well as where the main chain is the partner chain itself.  not sure it needs to be concerned with right now.

primitives/authority-selection-inherents/src/authority_selection_inputs.rs
paul-weiss: @aslesarenko lets chat more about this one.  given we now have 2 different tokens, makes sense to make generic here.

mainchain-follower/db-sync-follower/src/candidates/mod.rs
LGLO: IMO (u16, u16, vec<u16>) doesn't help.
After all it could be a Vec only or fields only and there has to be connection between types of candidates and parts of D-Parameter. Please bare in mind that this part of code is reading data from Datum and this can be affected by how the data is written.

For the decision Vec vs Map vs named fields should be driven by shape of data and what are we building: a generalization to N types of candidates or generalization to 3 candidates types

mainchain-follower/db-sync-follower/src/candidates/mod.rs
mpskowron: What you wrote looks fine. However,  I would suggest a different approach on getting there. Since this code already is a CardanoDS, instead of modifying it, create a separate module (or even a simple function in authority-selection-inherents for now) which will glue the results of CardanoDS and EthereumDS. This way the diff will be easier to read (especially of the next PR) and the change will be done in less invasive way.

mpskowron (commented):
It's ok that we don't have it done in a generic way at the moment (maybe it's even better, cause I'm not sure if we have enough data to create a good abstraction). However, I think that it would be a good idea to have Eth related code separated more clearly from Ada part (for example, it is tightly coupled together in filter_invalid_candidates now). Then, we can have thin layers of "glue code" on top. I think that this will end up in cleaner and more readable code. This will also make future refactorings, like adding new sources or making generalizations much easier.

AmbientTea:
There's not been any activity on this PR for over a month. Is it still relevant?

PR #68, opened 2024-08-30
fix: ETCM-8267 Configure and use the native token addresses in partner-chains-cli

AmbientTea:
# Description

Generating the chain spec needs three new environmental variables to be configured, after the addition of the native token pallet. This PR modifies `prepare-configuration` step to also prompt for the addresses and save them, and then uses them in `create-chain-spec`.

# Checklist

- [ ] Self-reviewed the diff

# Note on CI
for you.


partner-chains-cli/src/prepare_configuration/prepare_main_chain_config.rs
```diff
@@ -129,6 +132,21 @@ fn run_sidechain_main_cli_addresses<C: IOContext>(
 	Ok(())
 }
 
+fn prepare_native_token<C: IOContext>(context: &C) -> anyhow::Result<()> {
+	if context.prompt_yes_no("Do you want to configure a native token for you Partner Chain?", true)
```
mpskowron, line 136: I'm missing a better explanation here - what is it for and how to prepare the configuration variables.

mpskowron (approved):

partner-chains-cli/src/prepare_configuration/prepare_main_chain_config.rs
```diff
@@ -129,6 +132,21 @@ fn run_sidechain_main_cli_addresses<C: IOContext>(
...
+	{
+		NATIVE_TOKEN_POLICY.prompt_with_default_from_file_and_save(context);
+		NATIVE_TOKEN_ASSET_NAME.prompt_with_default_from_file_and_save(context);
+		ILLIQUID_SUPPLY_ADDRESS.prompt_with_default_from_file_and_save(context);
```
LGLO, line 140: `IlliquidCirculationSupplyValidator` is present in addresses output and should be placed in the part of JSON as the other addresses and policy ids.

LGLO (changes requested):
In this PR missing env variables are set for chain spec creation, but neither user is lectured about the values nor they are used in setup main chain state phase, so it will make this CLI incomplete and chain not set up properly.

partner-chains-cli/src/prepare_configuration/prepare_main_chain_config.rs
AmbientTea: I added a slightly more verbose explanation in the wizard.

partner-chains-cli/src/config.rs
```diff
@@ -376,11 +377,18 @@ pub struct SidechainParams {
 	pub governance_authority: MainchainAddressHash,
 }
 
+#[derive(Deserialize, Serialize, Parser, Clone, Debug)]
+pub struct NativeTokenConfig {
+	native_token_policy: String,
+	native_token_asset_name: String,
```
LGLO, line 383: cosmetic: it could easily be `AssetConfig` with field `policy_id` and `asset_name`

> An asset is uniquely identified by an asset ID, which is a pair of both the policy ID and asset name.

partner-chains-cli/src/config.rs
```diff
@@ -376,11 +377,18 @@ pub struct SidechainParams {
...
+}
+
 #[derive(Deserialize)]
 pub struct ChainConfig {
 	pub cardano: CardanoParameters,
 	pub chain_parameters: SidechainParams,
 	pub cardano_addresses: MainChainAddresses,
+	pub native_token: NativeTokenConfig,
```
LGLO, line 391: A: Is it already approved that we support using exactly one ?

B: this is logically bound to `illiquid_supply_address` from `cardano_addresses`. I think it is better to keep this asset id where we have other policy ids. And what is more, if not structurally bound like:
```
struct IlliquidSupplyConfig {
  asset: AssetConfig, // or assets: Vec[AssetConfig] - depends on decision how many are supported.
  address: String,
}
```
then at least name here should help to bind these together

partner-chains-cli/src/prepare_configuration/prepare_main_chain_config.rs
```diff
@@ -126,6 +129,31 @@ fn run_pc_contracts_cli_addresses<C: IOContext>(
...
+		context,
+	);
+	Ok(())
+}
+
+fn prepare_native_token<C: IOContext>(context: &C) -> anyhow::Result<()> {
+	context.print(
+		"Partner Chains can store their initial token supply on Cardano as Cardano native tokens.",
+	);
+	context.print("Creation of the native token is not supported by this wizard and must be performed manually before this step.");
```
LGLO, line 147: Is there a plan, like in JIRA, to have this supported?
What should be the flow? At what point is policy_id/asset_name chosen? Does it involve using smart-contracts CLI?

AmbientTea: I talked with Piotr about this, and he told me we don't want to support this at all, at least for the time being. The issue I see with doing that, is that the token policy can have custom logic in it, so whatever we provide would be relevant only to those PC devs that want a token with exactly the logic that we provide (presumably none).

AmbientTea: > At what point is policy_id/asset_name chosen? Does it involve using smart-contracts CLI?

No, it can be any native token on Cardano. One can even be created using an existing creator, but someone can roll their own token code with custom logic and deploy it themselves

partner-chains-cli/src/config.rs
AmbientTea: A: Yes, we're supporting only one until MN requires us to support multiple. That's decided by @gilligan 

B: Okay, I moved it into `cardano_addresses` under `native_token`

partner-chains-cli/src/prepare_configuration/prepare_main_chain_config.rs
```diff
@@ -126,6 +129,31 @@ fn run_pc_contracts_cli_addresses<C: IOContext>(
...
+	if context.prompt_yes_no("Do you want to configure a native token for you Partner Chain?", true)
+	{
+		NATIVE_TOKEN_POLICY.prompt_with_default_from_file_and_save(context);
+		NATIVE_TOKEN_ASSET_NAME.prompt_with_default_from_file_and_save(context);
+	} else {
+		NATIVE_TOKEN_POLICY.save_to_file(&PolicyId::default().to_hex_string(), context);
+		NATIVE_TOKEN_ASSET_NAME.save_to_file(&"0x0".into(), context);
```
LGLO, line 154: Isn't it a bug? `0x0` doesn't seem to be a syntactically correct hex string. `0x` should be empty one, right?

LGLO (changes requested):
I cannot even pass `generate-keys` using this branch.
```
➜  test-native-token-cli ./partner-chains-cli generate-keys                      
This 🧙 wizard will generate the following keys and save them to your node's keystore:
→  an ECDSA Cross-chain key
→  an ED25519 Grandpa key
→  an SR25519 Aura key
It will also generate a network key for your node if needed.

> node base path ./data

⚙️ Generating Cross-chain (ecdsa) key
running external command: ./partner-chains-node key generate --scheme ecdsa --output-type json
💾 Inserting Cross-chain (ecdsa) key
running external command: ./partner-chains-node key insert --base-path ./data --scheme ecdsa --key-type crch --suri 'grace fence account between allow autumn toilet river river toilet climb repair'
command output: Error: Input("Reading configuration from environment failed: missing value for field native_token_policy_id")
Running executable failed with status exit status: 1
```

partner-chains-cli/src/prepare_configuration/prepare_main_chain_config.rs
AmbientTea: `0x0` is a valid hex string, I tested it.

`0x` is also valid, but to me it looks invalid so I went with `0x0`.

AmbientTea:
> I cannot even pass generate-keys using this branch.

My bad, I was running it with direnv enabled and didn't catch that. Fixed and re-verified it manually from gen-keys to start-node.

partner-chains-cli/src/prepare_configuration/prepare_main_chain_config.rs
LGLO: So, `0x` is empty hex - reasonable default.
`0x0` is parsed to one byte with value `0`. I still think that `0x` is better choice.

LGLO (commented):
3 remarks with decreasing severity:
* Native token configuration is not displayed in `create-chain-spec`, unlike the rest of data from config that goes into chain-spec
* Generated chain spec has some vague field `"marker": null` under `nativeTokenManagement` key
* 0x00 is one byte asset name, quite strange default

AmbientTea:
> * Native token configuration is not displayed in `create-chain-spec`, unlike the rest of data from config that goes into chain-spec

I've added the prints

> * Generated chain spec has some vague field `"marker": null` under `nativeTokenManagement` key

Can't do anything about it, we need the `GenesisConfig` struct to be parametrized with `T` to implement `BuildGenesisConfig` but we don't need to use it.

> * 0x00 is one byte asset name, quite strange default

Changed to `0x`.

LGLO (approved):

PR #71, opened 2024-09-02
feat: allow manual workflow_dispatch build from external source repos

skylar-simoncelli:


.github/workflows/earthly.yml
```diff
@@ -28,7 +35,10 @@ jobs:
       contents: write
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@v2
```
tgunnoe, line 38: why did you switch from v4 to v2?

PR #91, opened 2024-09-20
change: better result and IO assertions in CLI tests

AmbientTea:
# Description

Changed the helpers that assert empty IO and results to macro rules, so the test errors show the line within the test case, not the helper definition.

I would love to find a way to show the error at the line of the `MockIO` that caused the error, but it's more tricky here in Rust than would be in Scala with the scope implicit. 

# Checklist

- [x] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/partner-chains-cli/src/tests/mod.rs
+++ b/partner-chains-cli/src/tests/mod.rs
@@ -147,7 +147,7 @@ pub(crate) use should_have_no_io_left;
 
 macro_rules! should_be_success {
 	($result:expr, $context:expr) => {{
-		let result = $result.unwrap();
+		let result = $result.expect("Expected the result to be a success");
 		assert!(
 			$context.expected_io.borrow().is_empty(),
 			"Expected IO operations left unperformed: {:?}",
@@ -160,7 +160,7 @@ pub(crate) use should_be_success;
 
 macro_rules! should_be_failure {
 	($result:expr, $context:expr) => {{
-		let err = $result.unwrap_err();
+		let err = $result.expect_err("Expected the result to be an error");
 		assert!(
 			$context.expected_io.borrow().is_empty(),
 			"Expected IO operations left unperformed: {:?}",
--- a/partner-chains-cli/src/tests/mod.rs
+++ b/partner-chains-cli/src/tests/mod.rs
@@ -145,32 +145,6 @@ macro_rules! should_have_no_io_left {
 }
 pub(crate) use should_have_no_io_left;
 
-macro_rules! should_be_success {
-	($result:expr, $context:expr) => {{
-		let result = $result.expect("Expected the result to be a success");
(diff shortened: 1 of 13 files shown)


partner-chains-cli/src/tests/mod.rs
```diff
@@ -132,14 +132,44 @@ impl MockIOContext {
 			None => None,
 		}
 	}
-	pub fn no_more_io_expected(&self) {
+}
+
+macro_rules! should_have_no_io_left {
+	($context:expr) => {{
 		assert!(
```
AmbientTea, line 139: Yes, assertion is repeated 3 times, but reusing `should_have_no_io_left` requires it to be imported (macros don't close on their scope like functions)

AmbientTea: I'm not sure what the best option would behere tbh. Open for suggestions

LGLO: I don't like the two following macros, because the don't assert on the `Ok` of `Err`. I see in diff, that we failed to make better assertions. The two following macros help in hiding our omission.
Even in cases where it's enough to assert on `in_ok`, `is_err`, I don't think lines save are worth another layer of indirection.

AmbientTea: Okay, removed them and replaced them with `expect` and `expect_err`.

Btw. I don't think `is_err` and `is_ok` is ever a good thing to assert, because the first thing I do when assertion like that fails is to replace it with an `expect*` to see what error or what result is returned anyway.

mpskowron (approved):

vvviiimmm (approved):

gilligan:
I understand the idea is to make sure that operations that have been defined to be carried out via MockIOContext should actually be carried out and the test should *fail* if that is not the case. That makes sense.

Searching around I found a suggestion that would boil down to implementing [Drop](https://doc.rust-lang.org/std/ops/trait.Drop.html) for MockIOContext. So the `drop`would check `if !self.expected_io.borrow().is_empty()` and panic if not.

Entirely possible that this doesn't work or doesn't make sense for some reason - You let me know ;)

AmbientTea:
> I understand the idea is to make sure that operations that have been defined to be carried out via MockIOContext should actually be carried out and the test should _fail_ if that is not the case. That makes sense.
> 
> Searching around I found a suggestion that would boil down to implementing [Drop](https://doc.rust-lang.org/std/ops/trait.Drop.html) for MockIOContext. So the `drop`would check `if !self.expected_io.borrow().is_empty()` and panic if not.
> 
> Entirely possible that this doesn't work or doesn't make sense for some reason - You let me know ;)

This is a much better approach and I will switch this PR to it once I'm done with more important tasks!

PR #96, opened 2024-09-24
feat: add deregistration command to pc-cli

LGLO:
# Description

Adds deregister command to partner-chains-cli

# Checklist

- [x] Self-reviewed the diff

# Note on CI
for you.


partner-chains-cli/src/config.rs
```diff
@@ -457,11 +464,20 @@ pub mod config_fields {
...
 			default: Some("payment.skey"),
 			_marker: PhantomData,
 		};
 
+	pub const CARDANO_COLD_VERIFICATION_KEY_FILE: ConfigFieldDefinition<'static, String> =
+		ConfigFieldDefinition {
+			config_file: RESOURCES_CONFIG_FILE_PATH,
+			path: &["cardano_cold_verification_key_file"],
+			name: "path to the cold verification key file",
+			default: Some("cold.vkey"),
```
gilligan, line 477: This is of course only a default value so users can pick whatever makes sense to them but `cold.vkey` seems very generic?

partner-chains-cli/src/deregister/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+use anyhow::anyhow;
+
+#[derive(Debug, clap::Parser)]
+pub struct DeregisterCmd;
+
+impl CmdRun for DeregisterCmd {
+	fn run<C: IOContext>(&self, context: &C) -> anyhow::Result<()> {
+		let (chain_config, cardano_network) = read_chain_config_file(context)?;
+		context.print(
+		  &format!("This wizard will remove the candidate identified by given from the committee candidates of a chain identified by parameters:\n{}.\nCommittee Candidate Validator Address is {}\n",
```
gilligan, line 20:
```suggestion
		  &format!(This wizard will remove the specified candidate from the committee candidates based on the following parameters:\n{}.\nValidator Address of the candidate: {}\n",
```

partner-chains-cli/src/deregister/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+
+fn read_chain_config_file<C: IOContext>(
+	context: &C,
+) -> Result<(crate::config::ChainConfig, crate::config::CardanoNetwork), anyhow::Error> {
+	let chain_config = crate::config::load_chain_config(context);
+	let cardano_network = get_cardano_network_from_file(context);
+	chain_config.and_then(|chain_config| cardano_network.map(|cardano_network| (chain_config, cardano_network)))
+		    .map_err(|_|anyhow!(
+		        "Couldn't parse chain configuration file {}. The chain configuration file that was used for registration is required in the working directory.",
+		        CHAIN_CONFIG_FILE_PATH))
```
gilligan, line 53: So I reckon `partner-chains-cli-config.json` is the main config file for partner-chains-cli and is thus required to always be in CWD when executing the cli?

partner-chains-cli/src/deregister/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+			"{PC_CONTRACTS_CLI_PATH} deregister --network {} --ada-based-staking --spo-public-key {} {} {}",
+			cardano_network.to_network_param(),
+			cold_vkey,
+			smart_contracts::sidechain_params_arguments(&chain_config.chain_parameters),
+			smart_contracts::runtime_config_arguments(
+				&pc_contracts_cli_resources,
+				&payment_signing_key_path
+			)
+		))?;
+		// Smart Contracts CLI return code is 0 even if the deregistration fails, so we need to check the output text.
```
gilligan, line 36: This sounds like something we should fix in the smart contracts cli? Fine to work around it but we should also make sure this is actually fixed

partner-chains-cli/src/deregister/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+use crate::config::{get_cardano_network_from_file, CHAIN_CONFIG_FILE_PATH, PC_CONTRACTS_CLI_PATH};
+use crate::io::IOContext;
+use crate::pc_contracts_cli_resources::establish_pc_contracts_cli_configuration;
+use crate::{smart_contracts, CmdRun};
+use anyhow::anyhow;
+
+#[derive(Debug, clap::Parser)]
+pub struct DeregisterCmd;
+
+impl CmdRun for DeregisterCmd {
```
gilligan, line 16: You could perhaps consider to extract some of the functionality in the function into functions and make the `run` function a bit smaller & neater so it's easy to tell that it just:
1. prints info
2. obtains keys
3. obtains cli resources
4. runs cli
5. validates output

```Rust
impl CmdRun for DeregisterCmd {
    fn run<C: IOContext>(&self, context: &C) -> anyhow::Result<()> {
        let (chain_config, cardano_network) = read_chain_config_file(context)?;
        context.print(&format!(
            "This wizard will remove the candidate identified by the given details from the committee candidates of a chain with the following parameters:\n{}.\nCommittee Candidate Validator Address: {}\n",
            &chain_config.chain_parameters,
            &chain_config.cardano_addresses.committee_candidates_address
        ));
        
        let (payment_signing_key_path, cold_vkey) = get_cardano_keys(context)?;
        let pc_contracts_cli_resources = establish_pc_contracts_cli_configuration(context)?;

        let command = build_deregister_command(
            &cardano_network,
            &cold_vkey,
            &chain_config,
            &pc_contracts_cli_resources,
            &payment_signing_key_path,
        );
        
        let output = context.run_command(&command)?;
        
        // Smart Contracts CLI return code is 0 even if the deregistration fails, so we need to check the output text.
        if output.contains("transactionId") {
            context.print(&format!("Deregistration successful: {}", output));
            Ok(())
        } else {
            Err(anyhow!("Deregistration failed: {}", output))
        }
    }
}

```

partner-chains-cli/src/deregister/mod.rs
```diff
@@ -0,0 +1,82 @@
...
+			cardano_network.to_network_param(),
+			cold_vkey,
+			smart_contracts::sidechain_params_arguments(&chain_config.chain_parameters),
+			smart_contracts::runtime_config_arguments(
+				&pc_contracts_cli_resources,
+				&payment_signing_key_path
+			)
+		))?;
+		// Smart Contracts CLI return code is 0 even if the deregistration fails, so we need to check the output text.
+		if output.contains("transactionId") {
```
gilligan, line 37: I kind of like combinators over explicit `if else` but I don't have strong opinions:

```Rust
output
    .contains("transactionId")
    .then_some(context.print(&format!("Deregistration successful: {}", output)))
    .ok_or_else(|| anyhow!("Deregistration failed: {}", output))
```

partner-chains-cli/src/deregister/tests.rs
```diff
@@ -0,0 +1,246 @@
...
+use crate::tests::{MockIO, MockIOContext};
+use crate::CmdRun;
+use serde_json::json;
+
+const MY_PAYMEMENT_SKEY: &str = "my_payment.skey";
+const MY_COLD_VKEY: &str = "my_cold.vkey";
+
+#[test]
+fn happy_path() {
+	let mock_context = MockIOContext::new()
```
gilligan, line 14: *Maybe* the mock setup code could be moved out of the individual test functions? It makes it a bit harder to read the test and see what the behavior and expectation is.

So the happy_path() would be like
```Rust
fn happy_path() {
  let mock_context = build_mock_context_for_happy_path();
  let res = DeregisterCmd.run(&mock_context);
  assert!(res.is_ok());
}
```
Obviously the mock code doesn't disappear but it makes the relevant test code more visible I think

gilligan (approved):
I left a few comments - none of which are super important but more in the "do as you please" territory ;)

partner-chains-cli/src/config.rs
LGLO: Name of this key is arbitrary. I matches what Christos wonderful toolbox is using https://github.com/chrispalaskas/cardano-skepsis-toolbox/blob/1d3b00388b2307f8432b82b2f24fa915ee819e11/registerSPO.py#L117

partner-chains-cli/src/deregister/mod.rs
LGLO: Almost. This file itself is created by `prepare-main-chain-config`. Governance authority would create it, other users will get it as configuration of the chain.

partner-chains-cli/src/deregister/mod.rs
LGLO: I think it wouldn't hurt. :D

partner-chains-cli/src/deregister/mod.rs
LGLO: I would use combinator if Rust had a proper one.
In Scala I would write
```
Either.cond(
  output.contains("transactionId"),
  context.print(&format!("Deregistration successful: {}", output)),
  anyhow!("Deregistration failed: {}", output)
)
```

partner-chains-cli/src/deregister/mod.rs
LGLO: I used almost this text - not committing suggestion, because it won't compile.

partner-chains-cli/src/deregister/tests.rs
LGLO: In this case I prefer to have explicit "given" without a more layers of indirection. I think irrelevant details, like 5 IO operations to read Cardano key file are abstracted and it is enough.

m2ux (approved):
I see no obvious issues

PR #107, opened 2024-09-25
feat: Add release script

kpinter-iohk:
# Description

This PR adds a release cutting script that updates all crate versions and creates a release branch + commit with a single command.

# Checklist

- [ ] CI passes.
- [ ] Self-reviewed the diff


jankun4 (approved):

release.sh
```diff
@@ -0,0 +1,76 @@
...
+if git show-ref --tags "v$next_version" --quiet; then
+    echo "ERROR: tag already exists for version!"
+    exit 1
+fi
+
+echo "Making release changes for new release $next_version (last version: $current_version)"
+
+cargo set-version $next_version
+
+sed -i "s/# Unreleased/# Unreleased\n\n# v$next_version/" changelog.md
```
LGLO, line 72: This should in fact add whole new:
```
# Unreleased

## Added

## Fixed 
...and so on
```
at the top after substitution

kpinter-iohk: Done. I also added code to remove empty sections.

AmbientTea (commented):
if synchronizing the version requires so many steps, how about we keep a `version` file and treat it as a source of truth on version, so we just bump it and overwrite whatever is in the Cargo files?

gilligan (approved):

PR #139, opened 2024-10-07
ci: implement security auditing checks

nrdxp:
# Description

uses `cargo-audit` and the RustSec Advisory DB

see: https://rustsec.org/

The [config](https://github.com/input-output-hk/partner-chains/blob/audit/.cargo/audit.toml) file is pretty minimal to start.

It is possible to add irrelevant (to us) issues in the ignore list though. Issues can be reviewed in a nice formatted form after
the Action has run in a [follow up job](https://github.com/input-output-hk/partner-chains/actions/runs/11223235779/job/31197443260).

It is still debatable how we would like to run this. On every push for PRs (as is configured now), or perhaps simply just on a
regular interval (currently every day at midnight). We probably don't want to make this a hard requirement to merge PRs though, as it may take dedicated time and effort to:
* actually determine if the issue is relevant to our use of the code
* bump the dependencies, which may require unrelated code changes.

However, perhaps we have a regular review, in the form of a meeting or just async, to ensure we never get too far behind. Currently, we have 3 vulnerabilities that are reported and several warnings.

# Checklist

- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -1,0 +1,11 @@
+[advisories]
+ignore = []
+severity_threshold = "low"
+
+[output]
+format = "json"
+quiet = false
+
+[yanked]
+enabled = true
+
--- a/.github/workflows/audit.yml
+++ b/.github/workflows/audit.yml
@@ -1,0 +1,25 @@
+name: "Audit Dependencies"
+on:
+  push:
+    paths:
+      # Run if workflow changes
+      - '.github/workflows/audit.yml'
+      # Run on changed dependencies
+      - '**/Cargo.toml'
+      - '**/Cargo.lock'
+      # Run if the configuration file changes
+      - '.cargo/audit.toml'
+  # Rerun periodicly to pick up new advisories
+  schedule:
+    - cron: '0 0 * * *'
+  # Run manually
+  workflow_dispatch:
+
+jobs:
+  security_audit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: rustsec/audit-check@v2.0.0
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3539,7 +3539,7 @@ dependencies = [
  "httpdate",
  "itoa",
  "pin-project-lite",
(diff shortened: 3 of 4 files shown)


skylar-simoncelli:
Can this be migrated into `ci.yml` instead of another new workflow? I already have plans to remove all other superflous workflows and narrow down to only `ci.yml` and `cd.yml`, and even then I would like to eventually merge these two together

The below PR would move this step into `ci.yml` and execute in every pre-merge, post-merge and workflow_dispatch run:

https://github.com/input-output-hk/partner-chains/pull/149

This way it would be running several times per day, which should remove the need for the daily cron trigger

nrdxp:
Its possible, however, if we want to maintain the chron job, I'm not sure it is. We may want to run this exclusively as a chron job to avoid the noise of running for every PR push, perhaps.

skylar-simoncelli (approved):

skylar-simoncelli:
Sorry forgot to approve. PR looks good however you decide to implement.

I would have a preference for just including the job in ci.yml to run on PR as it only takes 30 seconds and could potentially highlight new vulnerabilities introduced in a branch before they are merged

AmbientTea (commented):
It looks like this can stop PRs that change `Cargo.toml` from being merged even if the audit failure doesn't come from their changes, just some committed version having a new vulnerability. Ideally we'd be able to merge changes that don't introduce issues, but I don't think that's possible to do with this tool. Perhaps we should just run this on cron?

nrdxp:
> It looks like this can stop PRs that change `Cargo.toml` from being merged even if the audit failure doesn't come from their changes, just some committed version having a new vulnerability. Ideally we'd be able to merge changes that don't introduce issues, but I don't think that's possible to do with this tool. Perhaps we should just run this on cron?

As is, I did not make this a require check so it wouldn't block merge in that way. However, I've also thought about just setting a cron job. The downside seems like less visibility.

gilligan:
Feel free to re-open when work in this is restarted

gilescope:
We have an ignore list that we review when we upgrade substrate:

https://github.com/input-output-hk/midnight-substrate-prototype/blob/d24df37ddda666d67e4f13d6756b2276901182a2/Earthfile#L703

PR #149, opened 2024-10-09
feat: merge audit step into ci.yml

skylar-simoncelli:


AmbientTea (commented):
Will this block PRs from merging if some existing dependency doesn't pass this check?

skylar-simoncelli:
> Will this block PRs from merging if some existing dependency doesn't pass this check?

No, informational only

PR #150, opened 2024-10-11
Ogmios client lib - minimal functionality

LGLO:
# Description

* Currently minimal functionality that will allow getting "network parameters" in pc-cli.
* Allows change of the backend, currently it is jsonrpsee, but it has some drawbacks
* Traits for nicer syntax

Anticipated development of this is adding query for UTXOs for given address - to not use cardano-cli for it and submitTx/evaluateTx for offchain-code.

# Checklist

- [x] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/utils/ogmios-client/src/jsonrpsee.rs
+++ b/utils/ogmios-client/src/jsonrpsee.rs
@@ -1,0 +1,52 @@
+//! OgmiosClient implementation with jsonrpsee.
+//! Major drawback is that it swallows the error response from the server in case of 400 Bad Request.
+
+use crate::{OgmiosClient, OgmiosClientError, OgmiosParams};
+use jsonrpsee::core::params::ArrayParams;
+use jsonrpsee::{
+	core::{client::ClientT, params::ObjectParams},
+	http_client::HttpClient,
+};
+use serde::de::DeserializeOwned;
+
+impl OgmiosClient for HttpClient {
+	async fn request<T: DeserializeOwned>(
+		&self,
+		method: &str,
+		params: OgmiosParams,
+	) -> Result<T, OgmiosClientError> {
+		match params {
+			OgmiosParams::ByName(map) => {
+				let mut object_params = ObjectParams::new();
+				map.into_iter()
+					.try_for_each(|(k, v)| object_params.insert(k, v))
+					.map_err(serde_error_to_ogmios_error)?;
+				ClientT::request(self, method, object_params)
+					.await
+					.map_err(jsonrpsee_to_ogmios_error)
+			},
+			OgmiosParams::Positional(v) => {
+				let mut array_params = ArrayParams::new();
+				v.into_iter()
+					.try_for_each(|v| array_params.insert(v))
+					.map_err(serde_error_to_ogmios_error)?;
+				ClientT::request(self, method, array_params)
+					.await
+					.map_err(jsonrpsee_to_ogmios_error)
+			},
+		}
+	}
+}
+
+fn jsonrpsee_to_ogmios_error(e: jsonrpsee::core::ClientError) -> OgmiosClientError {
+	match e {
+		jsonrpsee::core::ClientError::ParseError(e) => {
+			OgmiosClientError::ResponseError(e.to_string())
+		},
+		e => OgmiosClientError::RequestError(e.to_string()),
+	}
+}
+
+fn serde_error_to_ogmios_error(e: serde_json::Error) -> OgmiosClientError {
+	OgmiosClientError::ParametersError(e.to_string())
+}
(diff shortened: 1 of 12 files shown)


utils/ogmios-client/tests/query_ledger_state.rs
```diff
@@ -0,0 +1,123 @@
...
+use jsonrpsee::{http_client::HttpClient, server::Server, RpcModule};
+use ogmios_client::{
+	query_ledger_state::{EpochBoundary, EpochParameters, EraSummary, QueryLedgerState},
+	types::{SlotLength, TimeSeconds},
+};
+use serde_json::json;
+use std::net::SocketAddr;
+use tokio::sync::OnceCell;
+
+const SERVER_ADDRESS: OnceCell<SocketAddr> = OnceCell::const_new();
```
LGLO, line 12: This `OnceCell` here is more than required for a single test. IDK if it should stay as something usable when more tests are implemented?

mpskowron (approved):

utils/ogmios-client/src/query_network.rs
```diff
@@ -0,0 +1,33 @@
+//! Queries that start with `queryNetwork/`.
+
+use crate::{types::SlotLength, OgmiosClient, OgmiosClientError, OgmiosParams};
+use serde::Deserialize;
+use serde_json::Value;
+use std::collections::HashMap;
+
+pub trait QueryNetwork: OgmiosClient {
+	#[allow(async_fn_in_trait)]
+	async fn shelley_genesis_configuration(
```
gilligan, line 10: Would it make any sense to already make this generic instead of hardcoding shelley? It would be sufficient to just introduce an `era` argument wouldn't it?

utils/ogmios-client/src/query_network.rs
```diff
@@ -0,0 +1,33 @@
...
+
+impl<T: OgmiosClient> QueryNetwork for T {}
+
+#[derive(Clone, Debug, Deserialize, PartialEq, Eq)]
+#[serde(rename_all = "camelCase")]
+pub struct ShelleyGenesisConfigurationResponse {
+	pub network_magic: u32,
+	pub security_parameter: u32,
+	// Ogmios returns active_slots_coefficient as string representing rational number, like "1/20"
+	pub active_slots_coefficient: String,
```
gilligan, line 28: Do we have to process/work with this coefficient? If not it's probably fine to leave it as String. Otherwise we might want to parse it into some adequate rational type?

gilligan (approved):

utils/ogmios-client/src/jsonrpsee.rs
```diff
@@ -0,0 +1,52 @@
...
+					.map_err(serde_error_to_ogmios_error)?;
+				ClientT::request(self, method, array_params)
+					.await
+					.map_err(jsonrpsee_to_ogmios_error)
+			},
+		}
+	}
+}
+
+fn jsonrpsee_to_ogmios_error(e: jsonrpsee::core::ClientError) -> OgmiosClientError {
```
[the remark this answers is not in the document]
LGLO: By "swallow" I mean that there is no meaningful data in error, but let me check it again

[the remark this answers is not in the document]
LGLO: I've checked. It is `Transport` variant with boxed `Rejected { status_code: 400 }` inside, nothing more. `.source() == None`

utils/ogmios-client/src/query_network.rs
LGLO: No, because responses are different. If `era` becomes a parameter, then response cannot be that specific. I would say json Value.

utils/ogmios-client/src/query_network.rs
LGLO: I used String and was going to leave it to user to chose his library, but I think I'll add `fraction` dependency and parse in the client.

utils/ogmios-client/src/jsonrpsee.rs
```diff
@@ -0,0 +1,52 @@
...
+		jsonrpsee::core::ClientError::ParseError(e) => {
+			OgmiosClientError::ResponseError(e.to_string())
+		},
+		e => OgmiosClientError::RequestError(e.to_string()),
+	}
+}
+
+fn serde_error_to_ogmios_error(e: serde_json::Error) -> OgmiosClientError {
+	OgmiosClientError::ParametersError(e.to_string())
+}
```
AmbientTea, line 52: Those could be `From` instances

utils/ogmios-client/tests/query_ledger_state.rs
AmbientTea: How expensive is creating the server for each test? Imo it's better to have the mocked values defined together with each unit test, than to have one central definition that handles all cases.

AmbientTea (approved):

utils/ogmios-client/tests/query_ledger_state.rs
LGLO: It is cheap. I've refactored to have a handler definition inside test.

utils/ogmios-client/src/jsonrpsee.rs
LGLO: I've changed the first one, but left the second one, because it valid only in this context. In other context serde_json::Error could be ResponseError.

PR #151, opened 2024-10-11
change: Decouple mc hash DataSource from main-chain-follower-api crate

AmbientTea:
# Description

to the issue.

# Checklist

- [ ] Self-reviewed the diff

# Note on CI
for you.


mainchain-follower/mock/src/block.rs
```diff
@@ -10,17 +8,25 @@ pub struct BlockDataSourceMock {
 	mc_epoch_duration_millis: u32,
 }
 
-#[async_trait]
-impl BlockDataSource for BlockDataSourceMock {
-	async fn get_latest_block_info(&self) -> Result<MainchainBlock> {
+#[derive(Debug, Clone, PartialEq, serde::Serialize)]
+pub struct MainchainBlock {
```
LGLO, line 12: This struct is present 3 times: in db-sync-follower, mc-hash and mock. At the same time it is composed of `domain` types. What do you think about defining this struct in `domain` crate?

mainchain-follower/db-sync-follower/src/mc_hash/mod.rs
```diff
@@ -0,0 +1,59 @@
...
+
+	async fn get_latest_stable_block_for(
+		&self,
+		reference_timestamp: sp_timestamp::Timestamp,
+	) -> std::result::Result<Option<sidechain_mc_hash::MainchainBlock>, Self::Error> {
+		Ok(self.inner.get_latest_stable_block_for(
+			Timestamp(reference_timestamp.as_millis())
+		)
+		.await?
+		.map(|block| sidechain_mc_hash::MainchainBlock {
```
LGLO, line 31: I would go with `.map(From::from)`. Especially if return type is changed to `domain::MainchainBlock`

primitives/sidechain-mc-hash/src/lib.rs
```diff
@@ -59,32 +57,80 @@ impl Deref for McHashInherentDataProvider {
...
+	pub hash: McBlockHash,
+	pub epoch: McEpochNumber,
+	pub slot: McSlotNumber,
+	pub timestamp: u64, // seconds since UNIX_EPOCH
+}
+
+/// Queries about Cardano Blocks
+#[async_trait]
+pub trait McHashDataSource {
+	type Error;
```
LGLO, line 73: I see quite complex `&(dyn McHashDataSource<Error = E> + Send + Sync)` because of this.
What would we lose, in terms of error handling, if concrete error with message field was used here instead?

mainchain-follower/mock/src/block.rs
AmbientTea: Yeah, that's a good idea, at least for now. Different IDPs use different fields from this type so there's some space for reducing the APIs, but block data is still served by the same cache db-sync query, so we can re-use a common type in the APIs until their needs drift apart, if they ever do.

mainchain-follower/db-sync-follower/src/mc_hash/mod.rs
AmbientTea: all those casts went away completely after switching to `domain::MainchainBlock`

primitives/sidechain-mc-hash/src/lib.rs
AmbientTea: That might be an over-engineering on my part tbh. I didn't like that either the trait here must anticipate what kinds of error the data source might return, or use an opaque `Box<dyn Error>`, which is used anyway further down the execution flow. I'll try and see if switching this to a concrete type has any real drawbacks here.

AmbientTea: I changed it to use boxed error, we don't really loose anything here, since we're wrapping the error later anyway. I will do the same change for the native token IDP in a separate PR

AmbientTea: https://github.com/input-output-hk/partner-chains/pull/153

node/src/tests/inherent_data_tests.rs
```diff
@@ -90,27 +94,29 @@ async fn block_proposal_cidp_should_be_created_correctly() {
...
-		)]
-		.into(),
-	);
+	}]);
 	let data_sources = DataSources {
-		block: Arc::new(block_data_source),
 		candidate: Arc::new(MockCandidateDataSource::default()),
 		native_token: Arc::new(native_token_data_source),
+		mc_hash: Arc::new(mc_hash_data_source),
+		sidechain_rpc: Arc::new(SidechainRpcDataSourceMock::new(parent_stable_block)),
```
LGLO, line 119: Does inherent data provider need `sidechain_rpc` data source? I think it should not.

AmbientTea: You're right. I changed the CIDPs to only take the data sources they need.

LGLO (approved):

PR #157, opened 2024-10-15
ETCM-8596 query ogmios for network params instead of asking user for them

LGLO:
# Description

This removes prompt for Cardano Network and optional query for custom parameters. It uses Ogmios and formulas to calculate values.
Cardano Network is now part of CaradanoParameters. Before it was only stored in the same place in JSON.

# Checklist

- [x] Self-reviewed the diff

# Note on CI
for you.


AmbientTea (approved):
huge

LGLO:
> huge

Huge mistake. Mine.

What should have been done instead is reading network magic from Ogmios and removing other variables, because they are configuration of db-sync-follower. So, db-sync-follower should look into postgres to self-configure and then we have the proper solution.

LGLO:
Anyway, I'll merge it, because it still is UX improvement. Depending on investigation of node being able to figure our these parameters it will be properly simplified.

PR #171, opened 2024-10-21
Implement pallet_session_validator_managment and pallet_session integration

LGLO:
# Description

This simple integration allows to run a partner chain without using our custom `pallet-partner-chains-session`, using standard polkadot-sdk `pallet-session`.

It is known that this implementation has additional lag of one partner chains epoch when applying committees to sessions.

When `pallet-session` is used instead of `pallet-partner-chains-session` users have to call `setKeys` in `pallet-session`.

# Checklist

- [x] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/primitives/domain/src/lib.rs
+++ b/primitives/domain/src/lib.rs
@@ -229,7 +229,7 @@ impl TryFrom<Vec<u8>> for MainchainPublicKey {
 pub const MAINCHAIN_ADDRESS_HASH_LEN: usize = 28;
 
 /// Some hash of MainchainAddress, 28 bytes. Presumably blake2b_224.
-/// Way to get it: cardano-cli address key-hash --payment-verification-key-file <path to vkey>
+/// Way to get it: cardano-cli address key-hash --payment-verification-key-file FILE
 #[derive(
 	Clone, Copy, Decode, Default, Eq, Encode, Hash, MaxEncodedLen, PartialEq, ToDatum, TypeInfo,
 )]
@@ -524,7 +524,7 @@ impl CandidateRegistrations {
 /// -- keyi - 33 bytes compressed ECDSA public key of a committee member
 /// -- @
 /// newtype ATMSPlainAggregatePubKey = ATMSPlainAggregatePubKey ByteString
-/// https://github.com/input-output-hk/partner-chains-smart-contracts/blob/5b19d25a95c3ab49ae0e4c6ce0ec3376f13b3766/docs/Specification.md#L554-L561
+/// <https://github.com/input-output-hk/partner-chains-smart-contracts/blob/5b19d25a95c3ab49ae0e4c6ce0ec3376f13b3766/docs/Specification.md#L554-L561>
 #[derive(Clone, Debug, Decode, Encode, Eq, PartialEq, ToDatum, TypeInfo)]
 pub struct ATMSPlainAggregatePubKey(pub [u8; 32]);
 
--- a/primitives/session-manager/Cargo.toml
+++ b/primitives/session-manager/Cargo.toml
@@ -11,9 +11,11 @@ frame-system = { workspace = true }
 log = { workspace = true }
 pallet-session = { workspace = true }
 pallet-session-validator-management = { workspace = true }
+sp-session-validator-management = { workspace = true, optional = true }
 pallet-sidechain = { workspace = true }
 pallet-partner-chains-session = { workspace = true }
 sidechain-domain = { workspace = true }
+sp-runtime = { workspace = true, optional = true }
 sp-staking = { workspace = true }
 sp-std = { workspace = true }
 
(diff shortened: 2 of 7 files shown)


primitives/domain/src/lib.rs
```diff
@@ -229,7 +229,7 @@ impl TryFrom<Vec<u8>> for MainchainPublicKey {
 pub const MAINCHAIN_ADDRESS_HASH_LEN: usize = 28;
 
 /// Some hash of MainchainAddress, 28 bytes. Presumably blake2b_224.
-/// Way to get it: cardano-cli address key-hash --payment-verification-key-file <path to vkey>
+/// Way to get it: cardano-cli address key-hash --payment-verification-key-file FILE
```
LGLO, line 232: `cargo doc` was thinking it is HTML marker. I've `FILE` instead. `cardano-cli` help message is using `FILE` as well.

primitives/domain/src/lib.rs
```diff
@@ -524,7 +524,7 @@ impl CandidateRegistrations {
 /// -- keyi - 33 bytes compressed ECDSA public key of a committee member
 /// -- @
 /// newtype ATMSPlainAggregatePubKey = ATMSPlainAggregatePubKey ByteString
-/// https://github.com/input-output-hk/partner-chains-smart-contracts/blob/5b19d25a95c3ab49ae0e4c6ce0ec3376f13b3766/docs/Specification.md#L554-L561
+/// <https://github.com/input-output-hk/partner-chains-smart-contracts/blob/5b19d25a95c3ab49ae0e4c6ce0ec3376f13b3766/docs/Specification.md#L554-L561>
```
LGLO, line 527: Made it clickable in the rust docs.

primitives/session-manager/Cargo.toml
```diff
@@ -11,9 +11,11 @@ frame-system = { workspace = true }
 log = { workspace = true }
 pallet-session = { workspace = true }
 pallet-session-validator-management = { workspace = true }
+sp-session-validator-management = { workspace = true, optional = true }
```
LGLO, line 14: This is not a new dependency. It was a transitive dependency before. Without this line it was impossible to build this (session-manager) crate without building the whole project: `cargo build -p session-manager` was failing on std/no-std configuration errors. All changes in this file are related to this issue.

AmbientTea: This happens very often, maybe we should have a step checking if all the crates build solo, but this would probably cost too much time. :\

primitives/session-manager/src/pallet_session_support.rs
```diff
@@ -0,0 +1,108 @@
...
+/// Implements [`pallet_session::SessionManager`] and [`pallet_session::ShouldEndSession`] integrated with [`pallet_session_validator_management`].
+///
+/// To use it, wire it in runtime configuration of [`pallet_session`].
+#[allow(dead_code)]
+#[derive(new)]
+pub struct PalletSessionSupport<T> {
+	_phantom: PhantomData<T>,
+}
+
+impl<T: pallet_session_validator_management::Config + pallet_session::Config>
```
AmbientTea, line 19: I would move those trait bounds directly to the type definition above. `PalletSessionSupport` serves no purpose without those traits defined, and the compiler will then produce much better error messages when they are missing.

primitives/session-manager/src/pallet_session_support.rs
```diff
@@ -0,0 +1,108 @@
...
+	/// Updates the session index of [`pallet_session`].
+	// Instead of Some((*).expect) we could just use (*). However, we rather panic in presence of important programming errors.
+	fn new_session(new_index: SessionIndex) -> Option<Vec<T::AccountId>> {
+		debug!("PalletSessionSupport: New session {new_index}");
+		pallet_session::pallet::CurrentIndex::<T>::put(new_index);
+		Some(
+			pallet_session_validator_management::Pallet::<T>::rotate_committee_to_next_epoch()
+				.expect(
+					"PalletSessionSupport: Session should never end without current epoch validators defined. \
+				Check ShouldEndSession implementation or if it is used before starting new session",
```
AmbientTea, line 43: Hard to understand wording imo. How about this?

```suggestion
				This may be caused by ShouldEndSession containing an error or being used before starting new session",
```

primitives/session-manager/src/pallet_session_support.rs
```diff
@@ -0,0 +1,108 @@
...
+	T: pallet_session_validator_management::Config<ScEpochNumber = EpochNumber>,
+	EpochNumber: Clone + PartialOrd,
+{
+	fn should_end_session(n: BlockNumberFor<T>) -> bool {
+		let current_epoch_number = T::current_epoch_number();
+
+		let should_end = current_epoch_number
+			> pallet_session_validator_management::Pallet::<T>::current_committee_storage().epoch
+			&& pallet_session_validator_management::Pallet::<T>::next_committee().is_some();
+		debug!("PalletSessionValidatorManagementSessionManager: should_end_session({n:?}) = {should_end}");
```
AmbientTea, line 71: Maybe we should log a warning if we can't switch epochs because of a missing committee?

AmbientTea (approved):

primitives/session-manager/src/pallet_session_support.rs
LGLO: This required adding all bounds to the ShouldEndSession, keeping them here and required some additional code in the test, so I will keep it here, like for the other SessionManager in this crate.

PR #185, opened 2024-10-25
address associations pallet - demo

AmbientTea:
# Description

This is a demo of the client-facing API of the address associations pallet.

### DIFF
--- a/pallets/address-associations/src/lib.rs
+++ b/pallets/address-associations/src/lib.rs
@@ -61,7 +61,7 @@ pub mod pallet {
 		/// in the runtime and should be invoked only when necessary.
 		#[pallet::call_index(1)]
 		#[pallet::weight((1, DispatchClass::Normal))]
-		pub fn set_main_chain_scripts(
+		pub fn set_observability_configuration(
 			origin: OriginFor<T>,
 			observability_config: T::ObservabilityConfigurationType,
 		) -> DispatchResult {
--- a/pallets/address-associations/src/lib.rs
+++ b/pallets/address-associations/src/lib.rs
@@ -1,0 +1,101 @@
+//! Pallet storing associations from main chain addresses to parter chain address.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+
+pub use pallet::*;
+
+#[frame_support::pallet]
+pub mod pallet {
+	use super::*;
+	use frame_support::pallet_prelude::*;
+	use frame_system::pallet_prelude::*;
+	use scale_info::prelude::fmt::Debug;
+	use sp_std::vec::Vec;
+
+	pub const PALLET_VERSION: u32 = 1;
+
+	#[pallet::pallet]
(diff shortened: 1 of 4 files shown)


pallets/address-associations/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+
+	pub const PALLET_VERSION: u32 = 1;
+
+	#[pallet::pallet]
+	pub struct Pallet<T>(_);
+
+	#[pallet::config]
+	pub trait Config: frame_system::Config {
+		/// Main chain address type
+		type MainChainAddress: Member + Parameter + MaxEncodedLen;
```
LGLO, line 23: I would make it a `Key`

pallets/address-associations/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+
+		/// Partner chain address type
+		type PartnerChainAddress: Member + Parameter + MaxEncodedLen;
```
LGLO, line 26: I would make it a `Value` and make the pallet cardano-associations. I don't even know if cardano is important part here.

pallets/address-associations/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+	impl<T: Config> Pallet<T> {
+		/// Changes the observability configuration.
+		///
+		/// This extrinsic must be run either using `sudo` or some other chain governance mechanism.
+		///
+		/// Warning: this extrinsic also resets the internal state stored by the observability data source
+		/// in the runtime and should be invoked only when necessary.
+		#[pallet::call_index(1)]
+		#[pallet::weight((1, DispatchClass::Normal))]
+		pub fn set_main_chain_scripts(
```
LGLO, line 64: I propose `set_observability_configuration`.

pallets/address-associations/src/lib.rs
```diff
@@ -0,0 +1,101 @@
...
+	}
+
+	impl<T: Config> Pallet<T> {
+		/// Returns the current pallet version.
+		pub fn get_version() -> u32 {
+			PALLET_VERSION
+		}
+
+		/// Retrieves all main chain - partner chain address associations from the runtime storage.
+		pub fn get_all_address_associations() -> Vec<(T::MainChainAddress, T::PartnerChainAddress)>
```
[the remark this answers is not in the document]
AmbientTea: In principle they are. As Midnight or any other chain grows, there will be more and more associations. Also, there's no de-association, so they stay forever. Perhaps this function won't be needed at all, or we could return `impl Iterator`.

pallets/address-associations/src/lib.rs
AmbientTea: I don't think Cardano is important here, but having a `key-pair-pallet` would be too general. I'd like to stay within the domain "associating some main chain identifier with some partner chain identifier", whatever chain the main chain and partner chain happen to be.

PR #205, opened 2024-11-04
change: offchain commands crate

AmbientTea:
# Description

Boiler plate base for adding the smart contracts offchain commands to the node.

# Checklist

- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/cli/offchain-commands/src/lib.rs
+++ b/cli/offchain-commands/src/lib.rs
@@ -1,0 +1,45 @@
+#[derive(Clone, Debug, clap::Subcommand)]
+#[allow(clippy::large_enum_variant)]
+pub enum OffchainCmd {
+	/// Get the script addresses for a given sidechain
+	Addresses,
+	/// Register a committee candidate
+	Register,
+	/// Deregister a committee candidate
+	Deregister,
+	/// Insert new D parameter
+	InsertDParameter,
+	/// Update a D parameter
+	UpdateDParameter,
+	/// Set or update permissioned candidates list
+	UpdatePermissionedCandidates,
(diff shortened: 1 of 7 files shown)


cli/offchain-commands/src/lib.rs
```diff
@@ -0,0 +1,45 @@
...
+#[allow(clippy::large_enum_variant)]
+pub enum OffchainCmd {
+	/// Get the script addresses for a given sidechain
+	Addresses,
+	/// Register a committee candidate
+	Register,
+	/// Deregister a committee candidate
+	Deregister,
+	/// Insert new D parameter
+	InsertDParameter,
```
LGLO, line 11: I think I will implement it as one, idempotent upsert-d-parameter command.

After it succeeded you can be sure, that the current value is what users wanted to be set. 

1) having two is not convenient for user
2) having two is not convenient from wizards developer perspective

cli/offchain-commands/src/lib.rs
```diff
@@ -0,0 +1,43 @@
...
+	ReserveHandover,
+	/// Release currently available funds from an existing
+	ReserveRelease,
+}
+
+#[derive(Clone, Debug, clap::Parser)]
+#[command(author, version, about, long_about = None)]
+pub struct CommonArguments<SidechainParams: clap::Args> {
+	#[clap(flatten)]
+	sidechain_params: SidechainParams,
```
LGLO, line 28: I've learned that the next smart-contracts release will have smart-contracts changed, so they do not take all params but only parameters relevant to their logic, besides genesis-utxo that would serve identifier role

PR #245, opened 2024-11-16
change: use new_if_pallet_present to create native token IDP

AmbientTea:
# Description

Uses `new_if_pallet_present` in our reference node implementation. Unfortunately, Substrate's mock runtime version API is `uninmplemented!`, so this has to go behind a feature flag so our tests don't fail.

# Checklist

- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/node/node/src/inherent_data.rs
+++ b/node/node/src/inherent_data.rs
@@ -104,6 +104,16 @@ where
 				"SIDECHAIN_BLOCK_BENEFICIARY",
 			)?;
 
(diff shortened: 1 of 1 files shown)


LGLO:
Why is it needed to use this in our refrence node?

AmbientTea:
> Why is it needed to use this in our refrence node?

We recommend using it in the migration doc for native token management. I thought we could have it there as well (because I don't trust people to read the doc and expect them to rather go to the code). But maybe that's not needed. :P

PR #248, opened 2024-11-19
fix: CardanoNetwork in pc-cli should not be feed with 'network-magic' but with 'network'

LGLO:
# Description
network id wasn't  network_magic - it was a bug.
`CardanoNetwork`, since we query ogmios for network parameters, doesn't need to know network id, so the type can now be simplified and it's sole purpose is for invoking pc-smart-contracts. When we write offchain in Rust, it will go away completly. 
BTW#2 - this parameter is redundant in pc-smart-contracts, they should be able to figure out network from Ogmios.

# Checklist

- [x] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/toolkit/partner-chains-cli/src/config.rs
+++ b/toolkit/partner-chains-cli/src/config.rs
@@ -261,16 +261,26 @@ impl SelectOptions for NetworkProtocol {
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Serialize, Deserialize)]
-pub struct CardanoNetwork(pub u32);
+#[serde(rename_all = "lowercase")]
+pub enum CardanoNetwork {
+	Mainnet,
+	Testnet,
+}
 
 impl CardanoNetwork {
-	pub fn to_id(&self) -> u32 {
-		self.0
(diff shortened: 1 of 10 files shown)


toolkit/partner-chains-cli/src/config.rs
```diff
@@ -751,9 +759,7 @@ mod tests {
 
 	#[test]
 	fn test_from_network_id() {
-		assert_eq!(NetworkIdKind::from(CardanoNetwork(0)), NetworkIdKind::Mainnet);
-		assert_eq!(NetworkIdKind::from(CardanoNetwork(1)), NetworkIdKind::Testnet);
-		assert_eq!(NetworkIdKind::from(CardanoNetwork(2)), NetworkIdKind::Testnet);
-		assert_eq!(NetworkIdKind::from(CardanoNetwork(42)), NetworkIdKind::Testnet);
```
LGLO, line 757: Apparently, I was doing the wrong thing, but in a right and tested way.

kpinter-iohk (approved):

PR #258, opened 2024-11-20
remove usage of Balances::deposit from transfer handler

AmbientTea:
# Description

Removes `Balances::deposit` from native token transfer handler, which caused problems when the total transfer was less than the existential token amount. This "feature" was only added for visibility and is not needed by anyone.

# Checklist

- [ ] Self-reviewed the diff

# Note on CI
for you.


LGLO (approved):
Please make sure with SETs that they don't monitor this address/events.

PR #261, opened 2024-11-21
feat: add transformation from RegisterValidatorDatum to PlutusData

LGLO:
# Description

This transformation is required for implementing `register` command. It is also handy for `deregister` tests with trustworthy values

# Checklist

~- [ ] Any changes are noted in the `changelog.md` for affected crate~
- [x] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/toolkit/primitives/plutus-data/src/lib.rs
+++ b/toolkit/primitives/plutus-data/src/lib.rs
@@ -81,14 +81,17 @@ pub(crate) trait VersionedDatum: Sized {
 	}
 }
 
-pub struct VersionedDatumShape {
+/// This struct has the same shape as `VersionedGenericDatum` from smart-contracts.
+/// It is used to help implementing a proper `From` trait for `PlutusData` for
+/// datum types.
+pub(crate) struct VersionedGenericDatumShape {
 	pub datum: PlutusData,
 	pub generic_data: PlutusData,
 	pub version: u64,
 }
 
-impl From<VersionedDatumShape> for PlutusData {
-	fn from(value: VersionedDatumShape) -> Self {
+impl From<VersionedGenericDatumShape> for PlutusData {
+	fn from(value: VersionedGenericDatumShape) -> Self {
 		let mut list = PlutusList::new();
 		list.add(&value.datum);
 		list.add(&value.generic_data);
--- a/toolkit/primitives/plutus-data/src/lib.rs
+++ b/toolkit/primitives/plutus-data/src/lib.rs
@@ -1,4 +1,4 @@
-use cardano_serialization_lib::PlutusData;
+use cardano_serialization_lib::{PlutusData, PlutusList};
 
 pub mod d_param;
 pub mod permissioned_candidates;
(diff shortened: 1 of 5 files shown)


toolkit/primitives/plutus-data/src/lib.rs
```diff
@@ -81,6 +81,22 @@ pub(crate) trait VersionedDatum: Sized {
...
+impl From<VersionedDatumShape> for PlutusData {
+	fn from(value: VersionedDatumShape) -> Self {
+		let mut list = PlutusList::new();
+		list.add(&value.datum);
+		list.add(&value.generic_data);
+		list.add(&PlutusData::new_integer(&value.version.into()));
+		PlutusData::new_list(&list)
+	}
+}
+
```
kpinter-iohk, line 99: I'm wondering if we plan to introduce something more typed in the future like this:
```rust
pub struct VersionedGenericDatum<Datum, GenericData> {
	pub datum: Datum,
	pub generic_data: GenericData,
	pub version: u64,
}

impl<Datum, GenericData> From<VersionedGenericDatum<Datum, GenericData>> for PlutusData
where
	cardano_serialization_lib::PlutusData: From<Datum>,
	cardano_serialization_lib::PlutusData: From<GenericData>,
{
	fn from(value: VersionedGenericDatum<Datum, GenericData>) -> Self {
		let mut list = PlutusList::new();
		list.add(&value.datum.into());
		list.add(&value.generic_data.into());
		list.add(&PlutusData::new_integer(&value.version.into()));
		PlutusData::new_list(&list)
	}
}
```

LGLO: Looking at example from this PR, what types would you use for `Datum` and `GenericData` ?

AmbientTea (approved):

toolkit/primitives/plutus-data/src/lib.rs
kpinter-iohk: If we want to match the type produced by offchain (since 7.0.0) then `VersionedGenericDatum<MainchainAddressHash, RegisterValidatorDatum>` but with the `own_pkh` field dropped from `RegisterValidatorDatum`.
See https://github.com/input-output-hk/partner-chains-smart-contracts/blob/4df4e5e/offchain/src/TrustlessSidechain/CommitteeCandidateValidator.purs#L249-L262

LGLO: The problem with it is that then you need another type that keeps this `own_pkh`. Or we need to somehow name `VersionedGenericDatum<MainchainAddressHash, RegisterValidatorDatum>` but even then we have `datum` field that has lost its meaningful name `own_pkh`. 

I think having domain type that has all required fields: `RegisterValidatorDatum` and a way of transforming it to `PlutusData` is the thing that is needed. I've added `VersionedDatumShape` to help ourselves with this transformation: it has the same shape as smart-contracts type. Without it, I still can transform `RegisterValidatorDatum` to `PlutusData`, but it is easier to get it wrong: user constructor instead of list and/or confuse the order of the fields

kpinter-iohk: Re-reading `impl From<RegisterValidatorDatum> for PlutusData` now I see that I didn't get the original intent and/or was too excited about my own idea. Thank you for explaining it!

kpinter-iohk (approved):

toolkit/primitives/plutus-data/src/lib.rs
LGLO: Before merging I'll make this struct `pub(crate)` and explain reason I've added it. It is missing and thus confusing.

PR #265, opened 2024-11-22
Upsert permissioned-candidates function

LGLO:
# Description

Upsert for permissioned candidates. Made very similar to D-parameter. One difference is that we should enforce comparision of candidates when they are sorted.

# Checklist

~- [ ] Any changes are noted in the `changelog.md` for affected crate~
- [x] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/toolkit/offchain/src/d_param.rs
+++ b/toolkit/offchain/src/d_param.rs
@@ -7,7 +7,7 @@
 //! `datum` field being `[num_permissioned_candidates, num_registered_candidates]`.
 
 use crate::csl::{
-	convert_ex_units, get_builder_config, InputsBuilderExt, TransactionBuilderExt,
+	get_builder_config, get_first_validator_budget, InputsBuilderExt, TransactionBuilderExt,
 	TransactionContext,
 };
 use crate::plutus_script::PlutusScript;
@@ -16,8 +16,8 @@ use cardano_serialization_lib::{
 	ExUnits, JsError, PlutusData, Transaction, TransactionBuilder, TxInputsBuilder,
 };
 use ogmios_client::{
-	query_ledger_state::QueryLedgerState, query_network::QueryNetwork,
-	transactions::OgmiosEvaluateTransactionResponse, transactions::Transactions, types::OgmiosUtxo,
+	query_ledger_state::QueryLedgerState, query_network::QueryNetwork, transactions::Transactions,
+	types::OgmiosUtxo,
 };
 use partner_chains_plutus_data::d_param::{d_parameter_to_plutus_data, DParamDatum};
 use sidechain_domain::{DParameter, McTxHash, UtxoId};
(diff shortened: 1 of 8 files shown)


toolkit/offchain/src/d_param.rs
```diff
@@ -108,12 +108,12 @@ where
 	Ok(McTxHash(res.transaction.id))
 }
 
-fn get_mint_ex_units(response: Vec<OgmiosEvaluateTransactionResponse>) -> anyhow::Result<ExUnits> {
+fn get_mint_ex_units(validators_budgets: Vec<OgmiosEvaluateTransactionResponse>) -> anyhow::Result<ExUnits> {
```
kpinter-iohk, line 111: Could this and `get_spend_ex_units` be the same function instead? eg `get_ex_units`

toolkit/offchain/src/permissioned_candidates.rs
```diff
@@ -1,15 +1,195 @@
...
+			"Submit insert permissioned candidates transaction request failed: {}, bytes: {}",
+			e,
+			hex::encode(tx.to_bytes())
+		)
+	})?;
+	log::info!("Transaction submitted: {}", hex::encode(res.transaction.id));
+	Ok(McTxHash(res.transaction.id))
+}
+
+fn get_mint_ex_units(validators_budgets: Vec<OgmiosEvaluateTransactionResponse>) -> anyhow::Result<ExUnits> {
```
kpinter-iohk, line 128: Same with this.

toolkit/offchain/src/permissioned_candidates.rs
```diff
@@ -1,15 +1,195 @@
...
+			))
+		},
+		None => {
+			log::info!("There are permissioned candidates. Inserting new ones.");
+			Ok(Some(insert_permissioned_candidates(&validator, &policy, &candidates, ctx, ogmios_client).await?))
+		},
+	}
+}
+
+fn get_current_permissioned_candidates(
```
kpinter-iohk, line 65: AFAICT this is nearly identical to `get_current_d_parameter` so maybe they could be shared and renamed eg `get_validator_datum`

kpinter-iohk (approved):
Minor comments

toolkit/offchain/src/permissioned_candidates.rs
LGLO: It is but my DRY threshold is 3.

I could make it generic but this would require:
* type parameter itself
* parameter for constructing log messages

Also it would require to change the result type here to PermissionedCandidatesDatums and move sorting outside, but I don't really think it is a problem.

toolkit/offchain/src/d_param.rs
LGLO: Yes, I will add a function to `csl` so it could be used in 4 places (maybe more).

PR #273, opened 2024-11-27
feat: await transaction outputs after its submission

LGLO:
# Description

After transaction submission to the node, via Ogmios, we only know it is in the mempool.
This PR adds awaiting for transaction outputs being present address of the first transaction output.

This PR is using `log::info!`, not `println!`. I think we should discuss how do we want to output from the offchain code.

I was considering different approaches to making the awaiting fast in tests and being able to have production implementation as well. I'm also happy to discuss this aspect.
Alternatives I was thinking about were using a function that has implementation with: `#[cfg(test)]` and having smarter mock for ogmios, that would record UTXOs (decode transaction and modify itself).

# Checklist

- [x] Self-reviewed the diff

# Note on CI
for you.


toolkit/offchain/src/await_tx.rs
```diff
@@ -0,0 +1,192 @@
...
+
+impl AwaitTx for FixedDelayRetries {
+	async fn await_tx_output<Q: QueryLedgerState>(
+		&self,
+		query: &Q,
+		tx_id: McTxHash,
+		tx: &Transaction,
+	) -> anyhow::Result<()> {
+		let strategy = FixedInterval::new(self.delay).take(self.retries);
+		let address = tx.body().outputs().get(0).address().to_bech32(None)?;
```
AmbientTea, line 41: We can also query for the first utxo of the transaction directly:
```
curl -H 'content-type: application/json' -X POST http://localhost:1337 -d \
'{"jsonrpc": "2.0",
"method":"queryLedgerState/utxo",
"params":{
    "outputReferences":[
        {"transaction": {"id": "f9a60946cf9ad9578c71260637bdd4cbdbde4ea6b2c0ae916809e1e0ce6a0b74"}, "index": 0}
     ]
}}'  | jq
```
and just test for non-emptiness.

LGLO: Oh well, this mean a lot for this PR, because without knowing tx outputs we can in fact move the check out of current methods and use at higher level:
```
fn high_level_op(){
  let tx_id = make_and_submit_tx()?;
  await_tx(tx_id)
}
```
Thank you very much!

LGLO: I've updated the PR with new ogmios trait for this query

AmbientTea (approved):

toolkit/offchain/src/d_param.rs
```diff
@@ -103,9 +103,23 @@ where
...
 	})?;
-	log::info!("Transaction submitted: {}", hex::encode(res.transaction.id));
-	// TODO: await for the transaction output to be visible in validator wallet
-	Ok(McTxHash(res.transaction.id))
+	let tx_id = McTxHash(res.transaction.id);
+	log::info!("Transaction submitted: {}", hex::encode(tx_id.0));
+	Ok(tx_id)
+}
+
+/// Returns
```
kpinter-iohk, line 111: leftover comment?

kpinter-iohk: Actually this function seems unused.

LGLO: You are right. Removed.

kpinter-iohk (approved):

PR #234, opened 2024-11-14
ETCM-8250 test update d param

ladamesny:
# Description

Add your description here, if it fixes a particular issue please provide a
[link](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword=)
to the issue.

# Checklist

- [x] Commit sequence broadly makes sense and commits have useful messages.
- [x] New tests are added if needed and existing tests are updated.
- [ ] Relevant logging and metrics added
- [x] CI passes. See note on CI.
- [ ] Any changes are noted in the `changelog.md` for affected crate
- [x] Self-reviewed the diff

# Note on CI
If your PR is from a fork, the necessary CI jobs won't trigger automatically for security reasons.
You will need to get someone with write privileges.  Please contact IOG Partner Chains developers to do this
for you.

PR #274, opened 2024-11-27
Revert "ETCM-8250 test update d param"

ladamesny:
Reverts input-output-hk/partner-chains#234

We discovered a bug with `extract_block_author` - need to revert to prevent failing builds.


ladamesny:
The extract block author issue was fixed last week by @rsporny

PR #275, opened 2024-11-28
add: ETCM-8948 get-scripts command

AmbientTea:
# Description

Adds `get-scripts` command that prints all validator addresses and policy IDs for a given genesis utxo.

# Checklist

- [x] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -33,7 +33,7 @@ This changelog is based on [Keep A Changelog](https://keepachangelog.com/en/1.1.
 
 ## Added
 
-* Added `smart-contracts` command to the node with first command `init-governance`.
+* Added `smart-contracts` command to the node with sub-commands `init-governance` and `get-scripts`.
 
 # v1.3.0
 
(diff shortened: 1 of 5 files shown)


toolkit/cli/smart-contracts-commands/src/addresses.rs
```diff
@@ -0,0 +1,25 @@
+use partner_chains_cardano_offchain::{csl::NetworkTypeExt, scripts_data::get_scripts_data};
+use sidechain_domain::{NetworkType, UtxoId};
+
+use crate::NetworkArg;
+
+#[derive(Clone, Debug, clap::Parser)]
+pub struct AddressesCmd {
+	#[arg(long, short = 'n')]
+	network: NetworkArg,
```
LGLO, line 9: I think use ogmios connection data and read the network type from Ogmios.

AmbientTea: I thought about it, but I'm not sure. As it stands, the command can be used offline which is a bonus. Maybe I can change it so it requires either ogmios connection or network type so the user can choose

LGLO (approved):

PR #279, opened 2024-11-29
feat: script validate license compatibility for all workspace crates

mpskowron:
# Description

This pull request introduces a Python script to verify license compliance for dependencies in this dual-licensed repository. The script will reduce the risk of inadverent license violations.
I've tried to enforce it using cargo-deny. However, it doesn't seem to provide useful functionality. If you think it might be beneficial, I can add a deny.toml configuration which specifies all possible licenses and crate exceptions, but it won't ensure the same checks as the script provided.

# Checklist

~- [ ] New tests are added if needed and existing tests are updated.~ Scripts don't require tests, do they?
~- [ ] Any changes are noted in the `changelog.md` for affected crate~ Doesn't seem like this requires a note in changelog file, does it?
- [x] Self-reviewed the diff

### DIFF
--- a/dev/nix/shell.nix
+++ b/dev/nix/shell.nix
@@ -31,6 +31,7 @@
         gnumake
         gawk
         cargo-edit
+        cargo-license
         perl
       ]
       ++ (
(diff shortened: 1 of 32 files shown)


LGLO (approved):
I think it is a good starting point. I haven't tested it, because of missing packages in my python3.

PR #280, opened 2024-11-29
ETCM-8754: feat: implement register transaction

kpinter-iohk:
# Description

to the issue.

# Checklist

- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/toolkit/offchain/src/lib.rs
+++ b/toolkit/offchain/src/lib.rs
@@ -14,6 +14,7 @@ pub mod ogmios_mock;
 pub mod permissioned_candidates;
 /// Utilities for handling Plutus script data
 mod plutus_script;
+mod register;
 /// Provides synthetized scripts data
 pub mod scripts_data;
 #[cfg(test)]
--- a/toolkit/offchain/src/lib.rs
+++ b/toolkit/offchain/src/lib.rs
@@ -14,6 +14,7 @@ pub mod ogmios_mock;
 pub mod permissioned_candidates;
 /// Utilities for handling Plutus script data
 mod plutus_script;
+/// Supports candidate registration
 mod register;
 /// Provides synthetized scripts data
 pub mod scripts_data;
--- a/toolkit/offchain/src/register.rs
+++ b/toolkit/offchain/src/register.rs
@@ -13,7 +13,7 @@ use sidechain_domain::BlockProducerRegistration;
 fn register_tx(
 	validator: &PlutusScript,
 	block_producer_registration: &BlockProducerRegistration,
-	input_utxo: &OgmiosUtxo,
+	registration_utxo: &OgmiosUtxo,
 	own_registration_utxos: &[OgmiosUtxo],
 	ctx: &TransactionContext,
 	validator_redeemer_ex_units: ExUnits,
@@ -30,7 +30,7 @@ fn register_tx(
 				validator_redeemer_ex_units.clone(),
 			)?;
 		}
-		inputs.add_key_inputs(&[input_utxo.clone()], &ctx.payment_key_hash())?;
+		inputs.add_key_inputs(&[registration_utxo.clone()], &ctx.payment_key_hash())?;
 		tx_builder.set_inputs(&inputs);
 	}
 
@@ -85,7 +85,7 @@ mod tests {
 	const MIN_UTXO_LOVELACE: u64 = 1000000;
 	const FIVE_ADA: u64 = 5000000;
 
-	fn block_producer_registration(consumed_input: UtxoId) -> BlockProducerRegistration {
+	fn block_producer_registration(registration_utxo: UtxoId) -> BlockProducerRegistration {
 		BlockProducerRegistration {
 			stake_ownership: AdaBasedStaking {
 				pub_key: test_values::mainchain_pub_key(),
@@ -93,7 +93,7 @@ mod tests {
 			},
 			sidechain_pub_key: SidechainPublicKey(Vec::new()),
 			sidechain_signature: SidechainSignature(Vec::new()),
-			consumed_input,
+			registration_utxo,
 			own_pkh: MainchainAddressHash([0; 28]),
 			aura_pub_key: AuraPublicKey(Vec::new()),
 			grandpa_pub_key: GrandpaPublicKey(Vec::new()),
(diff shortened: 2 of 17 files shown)


toolkit/offchain/src/register.rs
```diff
@@ -0,0 +1,293 @@
...
+use crate::plutus_script::PlutusScript;
+use cardano_serialization_lib::{
+	BigNum, DataCost, ExUnits, JsError, MinOutputAdaCalculator, Transaction, TransactionBuilder,
+	TransactionOutputBuilder, TxInputsBuilder,
+};
+use ogmios_client::types::OgmiosUtxo;
+use partner_chains_plutus_data::registered_candidates::block_producer_registration_to_plutus_data;
+use sidechain_domain::BlockProducerRegistration;
+
+pub fn register_tx(
```
LGLO, line 13: I don't think it should be public. Like with init_goveranance or d-param, there will be public higher level function, that calls this one after querying Cardano (via Ogmios) for the state.

toolkit/offchain/src/register.rs
```diff
@@ -0,0 +1,293 @@
...
+		let min_ada = MinOutputAdaCalculator::new(
+			&output,
+			&DataCost::new_coins_per_byte(
+				&ctx.protocol_parameters.min_utxo_deposit_coefficient.into(),
+			),
+		)
+		.calculate_ada()?;
+		let output = amount_builder.with_coin(&min_ada).build()?;
+		tx_builder.add_output(&output)?;
+	}
```
LGLO, line 50: I'm not saying it is an error here, but why do we need to have payment_address output? Doesn't balancing add change if required?
I'm wondering if I should have added it to d-param and permissioned-candidates.

I think that with this output, tx requires over `2 x min_ada + C`, where `C` is the cost of datum at validator address.
OTOH without this address and wallet that has less than that, say `min_ada + C + D` will lose `D`, because it is not enough for keeping change UTXO.

toolkit/primitives/plutus-data/src/registered_candidates.rs
```diff
@@ -167,6 +174,7 @@ fn decode_tx_hash_datum(datum: PlutusData) -> Option<McTxHash> {
 	Some(McTxHash(TryFrom::try_from(bytes).ok()?))
 }
 
+//
```
LGLO, line 177: Please clean-up

LGLO (commented):
How far have you tested transactions created with this code?

toolkit/offchain/src/lib.rs
```diff
@@ -14,6 +14,7 @@ pub mod ogmios_mock;
 pub mod permissioned_candidates;
 /// Utilities for handling Plutus script data
 mod plutus_script;
+mod register;
```
LGLO, line 17: Please add a comment that `register` is about block producer (candiate) registration messages

toolkit/offchain/src/register.rs
```diff
@@ -0,0 +1,362 @@
...
+	fn register_tx_regression_test() {
+		let ex_units = ExUnits::new(&0u32.into(), &0u32.into());
+		let payment_key_utxos = vec![lesser_payment_utxo(), greater_payment_utxo(), input_utxo()];
+		let ctx = TransactionContext {
+			payment_key: payment_key(),
+			payment_key_utxos: payment_key_utxos.clone(),
+			network: NetworkIdKind::Testnet,
+			protocol_parameters: protocol_parameters(),
+		};
+		let own_registration_utxos = vec![payment_key_utxos.get(1).unwrap().clone()];
```
LGLO, line 193: This is confusing, because you have chosen `greater_payment_utxo` as `input_utxo`. Use `get(2)` to use the value produced by `input_utxo()`.
Perhaps will cascade change to assertions further down.

LGLO (approved):
`register_tx_regression_test` has this confusing choice of `input_utxo` parameter, please fix. Other than that LGTM!

toolkit/offchain/src/register.rs
kpinter-iohk: `own_registration_utxos` are the old registrations that have to be invalidated, while `input_utxo` is just whatever utxo that you use to prove that you can create this transaction. I wanted to make sure there is no overlap.
Could you please confirm if this makes sense?

LGLO: `input_utxo` - lets rename it to `registration_utxo` to be consistent with pc-contracts-cli and make it clearer, at least for us.
So, this `input_utxo` = `registration_utxo` = a chosen UTXO, it it in the signed message, so cannot be "whatever".

Yes, `own_registration_utxos` are the old registrations. Invalidated = spent in the tx we just create here.

LGLO: Side note: I really hope that on-chain code prevent malicious actors from spending UTXOs not sent by them.

toolkit/offchain/src/register.rs
```diff
@@ -0,0 +1,362 @@
...
+) -> anyhow::Result<McTxHash> {
+	let ctx = TransactionContext::for_payment_key(payment_signing_key, ogmios_client).await?;
+	let validator = crate::scripts_data::registered_candidates_scripts(genesis_utxo)?;
+	let validator_address = validator.address_bech32(ctx.network)?;
+	let input_utxo = ctx
+		.payment_key_utxos
+		.iter()
+		.find(|u| u.to_domain() == input_utxo)
+		.ok_or(anyhow!("input utxo not found at payment address"))?;
+	let own_registration_utxos = ogmios_client.query_utxos(&[validator_address]).await?;
```
AmbientTea, line 35: I don't fully understand this line here. This queries for all registrations under the validator address, not only the transaction submitter's own ones, right? What will happen with utxos of other candidates?

kpinter-iohk: Good point! I checked and it turns out that there was another filtering step in the original code here that I missed. Will implement now.

LGLO: I've missed that as well in review, even though I have my deregister draft that uses this filtering by `own_pkh`

toolkit/offchain/src/register.rs
LGLO: But it is `registration_utxo` in the current UI and documentation. So, if `BlockProducerRegistration` is a new thing in this code base, I would really like to see some consistency. 

It is indeed arbitrarily chosen, but not before calling `register` but before getting signatures.

toolkit/offchain/src/register.rs
kpinter-iohk: I'm moving this logic to the next PR.

toolkit/offchain/src/register.rs
kpinter-iohk: I agree that it should be consistent, I just wanted to provide the other alternative that I've seen. I made a commit that makes the name consistent across the code base, it might deserve its own PR.

PR #283, opened 2024-12-02
change: ETCM-8907 add logging of ogmios requests

AmbientTea:
# Description

Logs Ogmios requests and responses into `ogmios_client.log` file in HttpClient. There's some hoops to jump through to log json values because we're using jsonrpsee which hides them. I thought about introducing a wrapper type for HttpClient which would make it possible to configure the file to which we log but I'm not convinced it's really needed. 

I went with the simples 

# Checklist

- [x] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/toolkit/cli/smart-contracts-commands/Cargo.toml
+++ b/toolkit/cli/smart-contracts-commands/Cargo.toml
@@ -13,3 +13,6 @@ jsonrpsee = { workspace = true, features = ["client-core", "http-client", "macro
 tokio = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
+log = { workspace = true }
+env_logger = { workspace = true }
+log4rs = { workspace = true }
--- a/toolkit/cli/smart-contracts-commands/Cargo.toml
+++ b/toolkit/cli/smart-contracts-commands/Cargo.toml
@@ -14,5 +14,4 @@ tokio = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 log = { workspace = true }
-env_logger = { workspace = true }
 log4rs = { workspace = true }
--- a/toolkit/cli/smart-contracts-commands/src/lib.rs
+++ b/toolkit/cli/smart-contracts-commands/src/lib.rs
@@ -1,3 +1,7 @@
+use log4rs::{
+	append::{console::ConsoleAppender, file::FileAppender},
+	config::Appender,
+};
 use sidechain_domain::MainchainPrivateKey;
 
 pub mod get_scripts;
@@ -30,6 +34,8 @@ impl SmartContractsCmd {
 	}
 
 	pub fn execute_blocking(self) -> CmdResult<()> {
+		setup_logging()?;
+
 		tokio::runtime::Runtime::new()?.block_on(self.execute())
 	}
 }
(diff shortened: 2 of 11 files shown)


toolkit/utils/ogmios-client/src/jsonrpsee.rs
```diff
@@ -1,36 +1,72 @@
...
+	core::{client::ClientT, traits::ToRpcParams, ClientError},
 	http_client::HttpClient,
 };
 use serde::de::DeserializeOwned;
 
+const CONTRACTLOG_FILE: &str = "contractlog.json";
+
+macro_rules! contractlog {
+	($($arg:tt)*) => {
+		add_to_contractlog(&format!("{}\n", format!($($arg)*)))
```
LGLO, line 17: I expected to see `log::info!` and the user - smart contract commands, maybe pc-cli wizards setting logger to file.

LGLO: Also, `contractlog.json` name was leaking technical detail - Contract monad from CTL. We can do something less traumatic.

AmbientTea: Yeah, good idea. `env_logger` doesn't support that, but `log4rs` turns out to be very nice to use.

toolkit/cli/smart-contracts-commands/src/lib.rs
```diff
@@ -45,3 +51,23 @@ pub(crate) fn read_private_key_from_file(path: &str) -> CmdResult<MainchainPriva
 		.map_err(|_| format!("{} is not the valid lengh of 32", key_hex))?;
 	Ok(MainchainPrivateKey(key_bytes))
 }
+
+pub fn setup_logging() -> Result<(), Box<dyn std::error::Error + Send + Sync>> {
+	let stdout = ConsoleAppender::builder().build();
+	let ogmios_log = FileAppender::builder().build("ogmios_log.json")?;
```
LGLO, line 57: I suggest "ogmios_client.log", there is no point for having ".json" extension

toolkit/utils/ogmios-client/src/jsonrpsee.rs
```diff
@@ -2,35 +2,41 @@
...
-use jsonrpsee::core::params::ArrayParams;
 use jsonrpsee::{
-	core::{client::ClientT, params::ObjectParams},
+	core::{client::ClientT, traits::ToRpcParams, ClientError},
 	http_client::HttpClient,
 };
 use serde::de::DeserializeOwned;
 
+fn request_to_json(req: impl ToRpcParams) -> Result<String, OgmiosClientError> {
+	let json_str = match req.to_rpc_params().expect("Parameters are correct") {
```
LGLO, line 12: Please map error with `serde_error_to_parameters_error` or change return type, because with .expect id doesn't return Err.

Also, there is no need to match, `to_rpc_params().map_err(...).unwrap_or_default()` will yield `Box<RawValue>` that has display.

toolkit/utils/ogmios-client/src/jsonrpsee.rs
```diff
@@ -2,35 +2,41 @@
...
-			},
-			OgmiosParams::Positional(v) => {
-				let mut array_params = ArrayParams::new();
-				v.into_iter()
-					.try_for_each(|v| array_params.insert(v))
-					.map_err(serde_error_to_parameters_error)?;
-				Ok(ClientT::request(self, method, array_params).await?)
-			},
-		}
+		log::debug!("request: {}", request_to_json(params.clone())?);
```
LGLO, line 33: >2024-12-03T10:06:41.748986130+01:00 DEBUG ogmios_client::jsonrpsee - request: {"era":"shelley"}

Please add "method" at least

toolkit/cli/smart-contracts-commands/Cargo.toml
```diff
@@ -13,3 +13,6 @@ jsonrpsee = { workspace = true, features = ["client-core", "http-client", "macro
 tokio = { workspace = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
+log = { workspace = true }
+env_logger = { workspace = true }
```
LGLO, line 17: Is it used?

LGLO (commented):
I've run prepare-configuration wizard and there are no logs in file.

Please update PR description

toolkit/cli/smart-contracts-commands/Cargo.toml
AmbientTea: no, removed

PR #284, opened 2024-12-02
feat: remove numtide devshell

tgunnoe:
[github.com/numtide/devshell](https://github.com/numtide/devshell) eliminates a lot of environmental variables that pkgs.mkShell creates in favor of constructing them from scratch, in addition providing a few helpers for creating commands and envs. 

However, with removing this devshell helper we don't need to reconstruct more difficult environments such as those for macOS, as nixpkgs has greatly improved its darwin environments upstream and they work well with pkgs.mkShell (even using `nix develop --ignore-environment`). In addition, we don't need to use the abstracted way of doing the commands/envs that this extra dependency provides.

# Checklist

- [ ] Self-reviewed the diff

# Note on CI
for you.


dev/nix/shell.nix
```diff
@@ -15,140 +15,63 @@
...
+          self'.packages.cardano-cli
+          self'.packages.pc-contracts-cli
+        ]
+        ++ (
+          if isDarwin
+          then
+            [ pkgs.darwin.apple_sdk.frameworks.SystemConfiguration ]
+          else [ pkgs.clang ]
+        );
+        shellHook = ''
```
nrdxp, line 52: stylistic choice, but instead of using self hook for setting vars, the args passed to `mkShell` (its really `mkDerivation` underneath) are set in the environment automatically.

Up to you, but I typically prefer that unless there is some reason I actually need the hook, like some shell logic to generate a value, etc.

nrdxp (approved):
LGTM, nothing blocking and seems to be passing all the tests, and also seems to simplify the code. :+1:

dev/nix/shell.nix
tgunnoe: oh yeah, I prefer that way actually. getting back from PTO i just blanked my mind completely. thanks. shellHooks can be for other things

skylar-simoncelli:
LGTM but seems to be a conflict

LGLO (commented):
Seems it removes "the menu display" when we enter a devshell:
```
🔨 Welcome to Partner Chains Substrate Node Devshell

[CI/CD]

  awscli2         - Unified tool to manage your AWS services
  earthly         - Build automation for the container era
  kubectl         - Kubernetes CLI
  kubernetes-helm - Package manager for kubernetes

[Partner Chains]

  cardano-cli     - CLI v10.1.2 that is used in partner-chains dependency stack

[Rust]

  check           - Check rustc and clippy warnings
  fix             - Automatically fix rustc and clippy warnings

[[general commands]]

  menu            - prints this menu
  ```
  Doesn't it?
  
  It is not something I'll miss much. Only noting the difference.

tgunnoe:
> Doesn't it?

Correct. I will revisit more useful discoverability but I don't think anyone relies on this right now.

PR #299, opened 2024-12-05
ETCM-8754: implement register tx submit

kpinter-iohk:
# Description

This PR implements candidate registration transaction submission and provides a CLI for it as well.

# Checklist

- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/toolkit/cli/smart-contracts-commands/src/register.rs
+++ b/toolkit/cli/smart-contracts-commands/src/register.rs
@@ -1,5 +1,5 @@
 use jsonrpsee::http_client::HttpClient;
-use partner_chains_cardano_offchain::register::run_register;
+use partner_chains_cardano_offchain::{await_tx::FixedDelayRetries, register::run_register};
 use sidechain_domain::{
 	AdaBasedStaking, AuraPublicKey, BlockProducerRegistration, GrandpaPublicKey,
 	MainchainPublicKey, MainchainSignature, SidechainPublicKey, SidechainSignature, UtxoId,
@@ -51,6 +51,7 @@ impl RegisterCmd {
 			self.registration_utxo,
 			payment_key,
 			&client,
+			FixedDelayRetries::two_minutes(),
 		)
 		.await?;
 
--- a/toolkit/cli/smart-contracts-commands/src/register.rs
+++ b/toolkit/cli/smart-contracts-commands/src/register.rs
@@ -1,0 +1,59 @@
+use jsonrpsee::http_client::HttpClient;
+use partner_chains_cardano_offchain::register::run_register;
+use sidechain_domain::{
+	AdaBasedStaking, AuraPublicKey, BlockProducerRegistration, GrandpaPublicKey,
+	MainchainPublicKey, MainchainSignature, SidechainPublicKey, SidechainSignature, UtxoId,
+};
+
+use crate::{parse_sidechain_public_keys, read_private_key_from_file};
+
+#[derive(Clone, Debug, clap::Parser)]
+pub struct RegisterCmd {
+	#[clap(flatten)]
+	common_arguments: crate::CommonArguments,
+	#[arg(long)]
+	genesis_utxo: UtxoId,
+	#[arg(long)]
+	registration_utxo: UtxoId,
+	#[arg(long)]
+	payment_key_file: String,
+	#[arg(long)]
+	stake_signing_key_file: String,
+	#[arg(long, value_name = "SIDECHAIN_KEY:AURA_KEY:GRANDPA_KEY", value_parser = parse_sidechain_public_keys)]
+	sidechain_public_keys: (SidechainPublicKey, AuraPublicKey, GrandpaPublicKey),
+	#[arg(long)]
+	sidechain_signature: SidechainSignature,
+	#[arg(long)]
+	spo_public_key: MainchainPublicKey,
+	#[arg(long)]
+	spo_signature: MainchainSignature,
+}
+
+impl RegisterCmd {
+	pub async fn execute(self) -> crate::CmdResult<()> {
+		let payment_key = read_private_key_from_file(&self.payment_key_file)?;
+		let client = HttpClient::builder().build(self.common_arguments.ogmios_host)?;
+		let block_producer_registration = BlockProducerRegistration {
+			stake_ownership: AdaBasedStaking {
+				pub_key: self.spo_public_key,
+				signature: self.spo_signature,
+			},
+			sidechain_pub_key: self.sidechain_public_keys.0,
+			sidechain_signature: self.sidechain_signature,
+			registration_utxo: self.registration_utxo,
+			aura_pub_key: self.sidechain_public_keys.1,
+			grandpa_pub_key: self.sidechain_public_keys.2,
+		};
+
+		run_register(
+			self.genesis_utxo,
+			&block_producer_registration,
+			self.registration_utxo,
+			payment_key,
+			&client,
+		)
+		.await?;
+
+		Ok(())
+	}
+}
--- a/toolkit/cli/smart-contracts-commands/src/register.rs
+++ b/toolkit/cli/smart-contracts-commands/src/register.rs
@@ -38,6 +38,7 @@ impl RegisterCmd {
 			},
 			sidechain_pub_key: self.sidechain_public_keys.0,
 			sidechain_signature: self.sidechain_signature,
+			own_pkh: crate::payment_signing_key_to_mainchain_address_hash(payment_key.clone())?,
 			registration_utxo: self.registration_utxo,
 			aura_pub_key: self.sidechain_public_keys.1,
 			grandpa_pub_key: self.sidechain_public_keys.2,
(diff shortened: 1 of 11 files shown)


toolkit/cli/smart-contracts-commands/src/register.rs
```diff
@@ -0,0 +1,60 @@
...
+	#[clap(flatten)]
+	common_arguments: crate::CommonArguments,
+	#[arg(long)]
+	genesis_utxo: UtxoId,
+	#[arg(long)]
+	registration_utxo: UtxoId,
+	#[arg(long)]
+	payment_key_file: String,
+	#[arg(long)]
+	stake_signing_key_file: String,
```
LGLO, line 21: I think this should not be present here?

toolkit/primitives/domain/src/lib.rs
```diff
@@ -636,11 +656,20 @@ pub struct BlockProducerRegistration {
 	pub sidechain_pub_key: SidechainPublicKey,
 	pub sidechain_signature: SidechainSignature,
 	pub registration_utxo: UtxoId,
-	pub own_pkh: MainchainAddressHash,
+	// pub own_pkh: MainchainAddressHash,
```
LGLO, line 659: Why cannot it be here?

toolkit/offchain/src/register.rs
```diff
@@ -1,17 +1,194 @@
...
+		own_pkh,
+		block_producer_registration.stake_ownership.pub_key.clone(),
+		&all_registration_utxos,
+	)?;
+	let own_registration_utxos = own_registrations.iter().map(|r| r.0.clone()).collect::<Vec<_>>();
+
+	if own_registrations.iter().any(|(_, existing_registration)| {
+		block_producer_registration.matches_keys(existing_registration)
+	}) {
+		return Err(anyhow!("BlockProducer with given set of keys is already registered"));
```
LGLO, line 98: This code prevents users fixing their existing registration. Imagine they have given wrong signatures. Instead of matching keys, everything should be checked.

In case of perfect match, return None instead of error. This way you have nice idempotent API.
User wants particular registration to be on chain. So if he runs same command again, it will be Ok(None).

kpinter-iohk: I copied the logic from the existing offchain code. You do make a good point though, I will fix this.

toolkit/primitives/domain/src/lib.rs
kpinter-iohk: If you remember when pair coding we had some trouble with a function and had to pass in own_pkh along with BlockProducerRegistration. When trying to construct a BlockProducerRegistration value for command submission, own_pkh is only available later too. In offchain we don't have this field in BlockProducerRegistration either. So far it seems that own_pkh doesn't belong in BlockProducerRegistration.

toolkit/cli/smart-contracts-commands/src/register.rs
kpinter-iohk: Indeed, thank you!

toolkit/primitives/domain/src/lib.rs
```diff
@@ -267,6 +267,16 @@ impl MainchainAddressHash {
 #[byte_string(debug, hex_serialize)]
 pub struct MainchainSignature(pub Vec<u8>);
 
+impl FromStr for MainchainSignature {
```
LGLO, line 270: Just add `decode_hex` to the bytes_string arguments above

toolkit/offchain/src/register.rs
```diff
@@ -1,14 +1,168 @@
...
+
+fn get_own_registrations(
+	own_pkh: MainchainAddressHash,
+	spo_pub_key: MainchainPublicKey,
+	validator_utxos: &[OgmiosUtxo],
+) -> Result<Vec<(OgmiosUtxo, BlockProducerRegistration)>, anyhow::Error> {
+	let mut own_registrations = Vec::new();
+	for validator_utxo in validator_utxos {
+		let datum = validator_utxo.datum.clone().ok_or_else(|| {
+			anyhow!("Invalid state: an UTXO at the validator script address does not have a datum")
```
LGLO, line 147: Is the presence of datum checked by smart contract? Is its shape checked as well? I think you should instead ignore such UTXOs but for sure don't crash this method because some malicious actor can transfer UTXO that will prevent anyone from registering with this code.

toolkit/offchain/src/register.rs
```diff
@@ -1,14 +1,168 @@
...
+		block_producer_registration.stake_ownership.pub_key.clone(),
+		&all_registration_utxos,
+	)?;
+	let own_registration_utxos = own_registrations.iter().map(|r| r.0.clone()).collect::<Vec<_>>();
+
+	if own_registrations
+		.iter()
+		.any(|(_, existing_registration)| block_producer_registration == existing_registration)
+	{
+		log::info!("✅ Block producer already registered with same keys.");
```
LGLO, line 91: Cosmetic: in the command and in few logs below we use "candidate" not "block producer", in user facing messages lets be consistent (in the code we could be as well)

toolkit/cli/smart-contracts-commands/src/register.rs
```diff
@@ -0,0 +1,58 @@
...
+	#[arg(long)]
+	genesis_utxo: UtxoId,
+	#[arg(long)]
+	registration_utxo: UtxoId,
+	#[arg(long)]
+	payment_key_file: String,
+	#[arg(long, value_name = "SIDECHAIN_KEY:AURA_KEY:GRANDPA_KEY", value_parser = parse_sidechain_public_keys)]
+	sidechain_public_keys: (SidechainPublicKey, AuraPublicKey, GrandpaPublicKey),
+	#[arg(long)]
+	sidechain_signature: SidechainSignature,
```
LGLO, line 23: I'm not sure what to do about using `sidechain`. We are "partner chains" for quite a long time but we have not renamed it in many places.

toolkit/offchain/src/register.rs
```diff
@@ -1,18 +1,174 @@
...
-	TransactionOutputBuilder, TxInputsBuilder,
+	BigNum, DataCost, ExUnits, JsError, MinOutputAdaCalculator, PlutusData, Transaction,
+	TransactionBuilder, TransactionOutputBuilder, TxInputsBuilder,
+};
+use ogmios_client::{
+	query_ledger_state::{QueryLedgerState, QueryUtxoByUtxoId},
+	query_network::QueryNetwork,
+	transactions::Transactions,
+	types::OgmiosTx,
+	types::OgmiosUtxo,
```
LGLO, line 22: cosmetic: please group these 2

toolkit/offchain/src/register.rs
```diff
@@ -1,18 +1,174 @@
...
+		.iter()
+		.find(|u| u.to_domain() == candidate_registration.registration_utxo)
+		.ok_or(anyhow!("registration utxo not found at payment address"))?;
+	let all_registration_utxos = ogmios_client.query_utxos(&[validator_address]).await?;
+	let own_registrations = get_own_registrations(
+		candidate_registration.own_pkh,
+		candidate_registration.stake_ownership.pub_key.clone(),
+		&all_registration_utxos,
+	);
+	let own_registration_utxos = own_registrations.iter().map(|r| r.0.clone()).collect::<Vec<_>>();
```
LGLO, line 85: cosmetic: this could go lower, because now it breaks into checking if there exists registration with given keys

toolkit/offchain/src/register.rs
```diff
@@ -1,18 +1,174 @@
...
+		anyhow!(
+			"Submit candidate registration transaction request failed: {}, bytes: {}",
+			e,
+			hex::encode(tx.to_bytes())
+		)
+	})?;
+	let tx_id = result.transaction.id;
+	log::info!("✅ Transaction submited. ID: {}", hex::encode(result.transaction.id));
+	await_tx
+		.await_tx_output(ogmios_client, UtxoId { tx_hash: McTxHash(tx_id), index: UtxoIndex(0) })
```
LGLO, line 133: Cosmetic: `UtxoId::new(tx_id.0, 0)` is a bit more succint.

toolkit/offchain/src/register.rs
```diff
@@ -1,18 +1,174 @@
...
+	for validator_utxo in validator_utxos {
+		match get_candidate_registration(validator_utxo.clone()) {
+			Ok(candidate_registration) => {
+				if candidate_registration.stake_ownership.pub_key == spo_pub_key
+					&& candidate_registration.own_pkh == own_pkh
+				{
+					own_registrations.push((validator_utxo.clone(), candidate_registration.clone()))
+				}
+			},
+			Err(e) => log::warn!("Found invalid UTXO at validator address: {}", e),
```
LGLO, line 154: IMO it is debug, because for the reasons previously outlined. We'll already pay for malicious inputs in DB access time, we can at least make it better for us when we mostly ignore them.

LGLO (approved):

PR #302, opened 2024-12-05
fix: Change upgrade_and_set_addresses dispatch class to Operational

AmbientTea:
# Description

Normal dispatch class is affected by block weight limits. `upgrade_and_set_addresses` by necessity consumes all remaining weight of the block (this behaviour comes form `set_code`), which on my local chain caused it to never be included in a block. Changing the dispatch class to operational removes this issue. 

# Checklist

- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/pallets/sidechain/src/lib.rs
+++ b/pallets/sidechain/src/lib.rs
@@ -113,7 +113,7 @@ pub mod pallet {
 	#[pallet::call]
 	impl<T: Config> Pallet<T> {
 		#[pallet::call_index(1)]
-		#[pallet::weight((0, DispatchClass::Normal))]
+		#[pallet::weight((0, DispatchClass::Operational))]
 		pub fn upgrade_and_set_addresses(
 			origin: OriginFor<T>,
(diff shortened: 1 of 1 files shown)


LGLO (approved):
Approved, however I don't fully understand the issue, because this call has weight = 0.

AmbientTea:
> Approved, however I don't fully understand the issue, because this call has weight = 0.

It has weight = 0, but nevertheless always consumes the whole weight of the block (because it invokes `set_code`). Normal dispatch extrinsics are only allowed to consume a part of the block per [the documentation](https://docs.substrate.io/build/tx-weights-fees/). Operational ones are not constrained in this way. This of course raises the question of why we didn't get the same problem running the extrinsic on other chains, and the answer is - I don't know. Perhaps we were lucky and it co-incided with some inherents being run, so it never was technically "the whole block", or there was some other reason. 🤷  But just to be safe, it's better to switch it to Operational.

PR #318, opened 2024-12-10
ETCM-8753: implement deregister command

kpinter-iohk:
# Description

to the issue.

# Checklist

- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/toolkit/cli/smart-contracts-commands/src/register.rs
+++ b/toolkit/cli/smart-contracts-commands/src/register.rs
@@ -1,5 +1,8 @@
 use jsonrpsee::http_client::HttpClient;
-use partner_chains_cardano_offchain::{await_tx::FixedDelayRetries, register::run_register};
+use partner_chains_cardano_offchain::{
+	await_tx::FixedDelayRetries,
+	register::{run_deregister, run_register},
+};
 use sidechain_domain::{
 	AdaBasedStaking, CandidateRegistration, MainchainPublicKey, MainchainSignature,
 	PermissionedCandidateData, SidechainSignature, UtxoId,
@@ -21,7 +24,7 @@ pub struct RegisterCmd {
 		long,
 		value_name = "PARTNERCHAIN_KEY:AURA_KEY:GRANDPA_KEY",
 		alias = "sidechain-public-keys",
-		value_parser=parse_partnerchain_public_keys
+		value_parser = parse_partnerchain_public_keys
 	)]
 	partnerchain_public_keys: PermissionedCandidateData,
 	#[arg(long, alias = "sidechain-signature")]
@@ -53,6 +56,39 @@ impl RegisterCmd {
 			self.genesis_utxo,
 			&candidate_registration,
 			payment_key,
+			&client,
+			FixedDelayRetries::two_minutes(),
+		)
+		.await?;
+
+		Ok(())
+	}
+}
+
+#[derive(Clone, Debug, clap::Parser)]
+pub struct DeregisterCmd {
+	#[clap(flatten)]
+	common_arguments: crate::CommonArguments,
+	#[arg(long)]
+	genesis_utxo: UtxoId,
+	#[arg(long)]
+	payment_key_file: String,
+	#[arg(long)]
+	spo_public_key: MainchainPublicKey,
+}
+
+impl DeregisterCmd {
+	pub async fn execute(self) -> crate::CmdResult<()> {
+		let payment_signing_key = read_private_key_from_file(&self.payment_key_file)?;
+		let client = HttpClient::builder().build(self.common_arguments.ogmios_url)?;
+		let own_pkh =
+			crate::payment_signing_key_to_mainchain_address_hash(payment_signing_key.clone())?;
+
+		run_deregister(
+			self.genesis_utxo,
+			payment_signing_key,
+			own_pkh,
+			self.spo_public_key,
 			&client,
 			FixedDelayRetries::two_minutes(),
 		)
--- a/toolkit/cli/smart-contracts-commands/src/register.rs
+++ b/toolkit/cli/smart-contracts-commands/src/register.rs
@@ -46,7 +46,7 @@ impl RegisterCmd {
 			},
 			partnerchain_pub_key: self.partnerchain_public_keys.sidechain_public_key,
 			partnerchain_signature: self.partnerchain_signature,
-			own_pkh: crate::payment_signing_key_to_mainchain_address_hash(payment_key.clone())?,
+			own_pkh: payment_key.to_pub_key_hash(),
 			registration_utxo: self.registration_utxo,
 			aura_pub_key: self.partnerchain_public_keys.aura_public_key,
 			grandpa_pub_key: self.partnerchain_public_keys.grandpa_public_key,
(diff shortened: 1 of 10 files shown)


toolkit/offchain/src/register.rs
```diff
@@ -133,6 +133,95 @@ pub async fn run_register<
...
+		anyhow!(
+			"Submit candidate unregistration transaction request failed: {}, bytes: {}",
+			e,
+			hex::encode(tx.to_bytes())
+		)
+	})?;
+	let tx_id = result.transaction.id;
+	log::info!("✅ Transaction submitted. ID: {}", hex::encode(result.transaction.id));
+	await_tx
+		.await_tx_output(ogmios_client, UtxoId { tx_hash: McTxHash(tx_id), index: UtxoIndex(0) })
```
LGLO, line 219: Cosmetic UtxoId::new

toolkit/offchain/src/register.rs
```diff
@@ -133,6 +133,95 @@ pub async fn run_register<
...
+	C: QueryLedgerState + QueryNetwork + QueryUtxoByUtxoId + Transactions,
+	A: AwaitTx,
+>(
+	genesis_utxo: UtxoId,
+	payment_signing_key: MainchainPrivateKey,
+	own_pkh: MainchainAddressHash,
+	stake_ownership_pub_key: MainchainPublicKey,
+	ogmios_client: &C,
+	await_tx: A,
+) -> anyhow::Result<Option<OgmiosTx>> {
```
LGLO, line 181: I think we usually return equivalent domain type `McTxHash`

toolkit/offchain/src/register.rs
```diff
@@ -208,6 +297,30 @@ fn register_tx(
...
+				own_registration_utxo,
+				validator,
+				validator_redeemer_ex_units.clone(),
+			)?;
+		}
+		tx_builder.set_inputs(&inputs);
+	}
+
+	tx_builder.balance_update_and_build(ctx)
+}
```
LGLO, line 322: Wow, quite simple.

toolkit/offchain/src/register.rs
```diff
@@ -130,7 +130,94 @@ pub async fn run_register<
...
+	let all_registration_utxos = ogmios_client.query_utxos(&[validator_address]).await?;
+	let own_registrations =
+		get_own_registrations(own_pkh, stake_ownership_pub_key.clone(), &all_registration_utxos);
+
+	if own_registrations.is_empty() {
+		log::info!("✅ Candidate is not registered.");
+		return Ok(None);
+	}
+
+	let own_registration_utxos = own_registrations.iter().map(|r| r.0.clone()).collect::<Vec<_>>();
```
AmbientTea, line 194: This could be moved to `get_own_registrations`

toolkit/offchain/src/register.rs
```diff
@@ -208,6 +295,30 @@ fn register_tx(
...
+	own_registration_utxos: &[OgmiosUtxo],
+	ctx: &TransactionContext,
+	validator_redeemer_ex_units: ExUnits,
+) -> Result<Transaction, JsError> {
+	let config = crate::csl::get_builder_config(ctx)?;
+	let mut tx_builder = TransactionBuilder::new(&config);
+
+	{
+		let mut inputs = TxInputsBuilder::new();
+		for own_registration_utxo in own_registration_utxos {
```
AmbientTea, line 309: In what situation would there be more than one registration utxo? A bug? I think re-registration burns the previous one, right? Anyway, I think burning all in deregister like this is the way to go.

toolkit/offchain/src/register.rs
```diff
@@ -208,6 +295,30 @@ fn register_tx(
...
+			inputs.add_script_utxo_input(
+				own_registration_utxo,
+				validator,
+				validator_redeemer_ex_units.clone(),
```
AmbientTea, line 313: Is this right? If there were multiple utxos, we'd probably get multiple ex_units, no?

AmbientTea: It looks like the `validator_redeemer_ex_units` should be a vec, and it should get zipped with the utxos vector

toolkit/offchain/src/register.rs
LGLO: If it is a bug depends on the specification, which I do not know.
Mainchain follower component sorts registrations and regards only that latest one.

toolkit/offchain/src/register.rs
LGLO: This is a good remark, but why zipped?  I'm not sure we have guarantees about order of adding inputs here and the order of costs returned by Ogmios.
If anything I would give each redeemer a different index and then match with the response from Ogmios.
I think we have exceeded the limit of expecting that we control order well enough..

toolkit/offchain/src/register.rs
```diff
@@ -130,7 +130,94 @@ pub async fn run_register<
...
+
+impl<T> Deregister for T
+where
+	T: QueryLedgerState + Transactions + QueryNetwork + QueryUtxoByUtxoId,
+{
+	async fn deregister(
+		&self,
+		genesis_utxo: UtxoId,
+		payment_signing_key: MainchainPrivateKey,
+		own_pkh: MainchainAddressHash,
```
LGLO, line 155: Isn't it derivative of payment_signing_key?

toolkit/cli/smart-contracts-commands/src/register.rs
```diff
@@ -61,3 +64,36 @@ impl RegisterCmd {
...
+	#[arg(long)]
+	spo_public_key: MainchainPublicKey,
+}
+
+impl DeregisterCmd {
+	pub async fn execute(self) -> crate::CmdResult<()> {
+		let payment_signing_key = read_private_key_from_file(&self.payment_key_file)?;
+		let client = HttpClient::builder().build(self.common_arguments.ogmios_url)?;
+		let own_pkh =
+			crate::payment_signing_key_to_mainchain_address_hash(payment_signing_key.clone())?;
```
LGLO, line 85: 1. I think payment signing key is enough to pass to run_register.
2. Now I see that the name `payment_signing_key_to_mainchain_address_hash` amplifies confusion of wrong type name. McAddressHash is in fact McKeyHash or McPublicKeyHash. It has nothing to do with address.
This is one of things that we should clean up.

toolkit/offchain/src/register.rs
```diff
@@ -130,7 +130,94 @@ pub async fn run_register<
...
+				e,
+				hex::encode(tx.to_bytes())
+			)
+		})?;
+	let validator_redeemer_ex_units = get_first_validator_budget(evaluate_response)?;
+	let tx = deregister_tx(&validator, &own_registration_utxos, &ctx, validator_redeemer_ex_units)?;
+	let signed_tx = ctx.sign(&tx).to_bytes();
+	let result = ogmios_client.submit_transaction(&signed_tx).await.map_err(|e| {
+		anyhow!(
+			"Submit candidate unregistration transaction request failed: {}, bytes: {}",
```
LGLO, line 211: Cosmetic: "deregistration"

toolkit/offchain/src/register.rs
LGLO: For me clear is that if there were 2 matching utxos then we should spend both, because keeping registration(s) after deregister command is a bad behavior for me.

LGLO (approved):

toolkit/offchain/src/register.rs
AmbientTea: Yes, it is

toolkit/offchain/src/register.rs
kpinter-iohk: `get_own_registrations` return `(OgmiosUtxo, CandidateRegistration)` pairs the second component of which is used in `register`

PR #319, opened 2024-12-10
Add integration tests for smart-contracts

LGLO:
# Description

Adds integration tests for:
* `init-goveranance`
* `upsert-d-parameter` - ignored, because it fails
* `registration` - ignored, because it fails

# Checklist

- [x] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/toolkit/offchain/tests/docker/entrypoint.sh
+++ b/toolkit/offchain/tests/docker/entrypoint.sh
@@ -1,5 +1,8 @@
 #! /bin/bash
 
+chmod 400 /config/keys/kes.skey
+chmod 400 /config/keys/vrf.skey
+
 byron_startTime=$(( $(date +%s) + 1 ))
 shelley_systemStart=$(date --utc +"%Y-%m-%dT%H:%M:%SZ" --date="@$byron_startTime")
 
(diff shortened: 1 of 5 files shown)


toolkit/offchain/tests/docker/entrypoint.sh
```diff
@@ -1,5 +1,8 @@
 #! /bin/bash
 
+chmod 400 /config/keys/kes.skey
+chmod 400 /config/keys/vrf.skey
+
```
LGLO, line 5: New image is already pushed. There was a problem with old image, because it was not enough to set 400 in my local repo.

toolkit/offchain/tests/integration_tests.rs
```diff
@@ -6,4 +6,204 @@
...
+		.submit_transaction(&signed_tx_bytes)
+		.await
+		.map_err(|_| ())
+		.map(|response| McTxHash(response.transaction.id))?;
+	FixedDelayRetries::new(Duration::from_millis(500), 100)
+		.await_tx_output(client, UtxoId::new(tx_hash.0, 0))
+		.await
+		.map_err(|_| ())?;
+	Ok(tx_hash)
+}
```
LGLO, line 135: I think that in the future I'll remove this and I'll build a transaction with CSL instead. It will be a lot of code, but later it will be easier to add more UTXOs when needed.

PR #361, opened 2025-01-09
Etcm 8983 fix get block author test

rsporny:
# Description

All in commit msg.

# Checklist

- [x] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/E2E-tests/config/substrate/staging_nodes.json
+++ b/E2E-tests/config/substrate/staging_nodes.json
@@ -1,7 +1,7 @@
 {
-    "deployment_mc_epoch": 779,
+    "deployment_mc_epoch": 808,
     "genesis_utxo": "de7b8efb08243b770be30cfa22a9ed64317c50052bb4394392960ddcc9810048#1",
-    "deployment_version": "v1.4.0",
+    "deployment_version": "798bc5dbff4f6910061afd1627d736ee99d8a081",
     "test_environment": "staging",
     "nodes_config": {
         "nodes": {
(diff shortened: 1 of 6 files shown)


alexromanov (approved):

chrispalaskas (commented):
Looks much nicer!! Please make sure it covers all corner cases though, epoch boundaries, first validator of an epoch missing, last validator of an epoch missing etc

mpskowron (commented):
Unfortunately, without signature verification, you can never be sure if the block was authored by the claimed author. I cannot find that part in the code anymore.

rsporny:
@chrispalaskas I've run it for multiple mc epochs, we should be good.

@mpskowron the test verifies if block author is in the committee. Extracting block signature was just a way to verify it, but your point is good, we lost that additional check now. @chrispalaskas, since we now have simpler method to get block author, we may utilize it in a new test that would validate if the claimed author really signed a block. Although, that might be slightly excessive... we would be testing aura's round robin essentially.

mpskowron:
> @mpskowron the test verifies if block author is in the committee. Extracting block signature was just a way to verify it, but your point is good, we lost that additional check now. @chrispalaskas, since we now have simpler method to get block author, we may utilize it in a new test that would validate if the claimed author really signed a block. Although, that might be slightly excessive... we would be testing aura's round robin essentially.

Not only aura's round robin but also the signing process and if slots are claimed correctly, etc. Anyway, the solution looks correct.

mpskowron (approved):

PR #367, opened 2025-01-13
feat: add WebSockets Ogmios client to overcome problem with swallowed error message

LGLO:
# Description

`HttpClient` of jsonrpsee (rightfully) returns `TransportError` when Ogmios returns 400 in HTTP layer (that occurs in case of bad request in JsonRPC layer). Unfortunately `TransportError` only carries error code, without the body.

WebSockets implementation doesn't return client error if but the error is contained in JsonRPC layer, so it can easily presented to the user.

Changes in this PR introduce a client that can be build with both HTTP and WebSockets URL and uses one or another underneath.

`partner-chains-cli` hasn't be updated to default to WebSockets yet.

# Checklist

- [x] Self-reviewed the diff

# Note on CI
for you.


kpinter-iohk (approved):

AmbientTea (approved):
awesome

PR #401, opened 2025-01-20
remove: initial_incentive parameter

AmbientTea:
# Description

The current implementation of the reserve release scripts uses a flat reward for the `reserve-release` transaction submitter, which means that it can be called in small increments to effectively drain the reserve of tokens. This PR removes this parameter from the user-facing parts of the code and hard-codes initial_incentive to 0 as a temporary fix before we either rework the incentivisation or remove it altogether. 

# Checklist

- [ ] Self-reviewed the diff

# Note on CI
for you.


toolkit/offchain/src/reserve/create.rs
```diff
@@ -110,7 +109,7 @@ impl From<&ReserveParameters> for ReserveDatum {
 				total_accrued_function_script_hash: value
 					.total_accrued_function_script_hash
 					.clone(),
-				initial_incentive: value.initial_incentive,
+				initial_incentive: 0,
```
LGLO, line 112: I think it deserves a comment explaining this hardcoded value.

Alternatively could be removed as well, but then datum encoding/decoding would deserve a comment.

AmbientTea: I added a comment here. I want the schema itself to stay true to the reality

LGLO: Understand. So it is not, because I have earlier chosen the other way.

LGLO (approved):

kpinter-iohk (approved):

PR #406, opened 2025-01-21
feat: cleanup too specific helpers and make sense of redeemer data 

LGLO:
# Description

* Merged `add_script_utxo_input` and `add_script_utxo_input_with_data` into one, because  the former assumed too much about redeemer data and is was re-usable only by chance
* `add_mint_one_script_token` - now requires redeemer data to make it really re-usable
* removed redundant `add_required_signer`, because our `balance_update_and_build` does it internally
* removed some redundant `add_script_reference_input`

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/toolkit/offchain/src/csl.rs
+++ b/toolkit/offchain/src/csl.rs
@@ -472,12 +472,7 @@ impl TransactionBuilderExt for TransactionBuilder {
 		);
 		let mint_witness = MintWitness::new_plutus_script(
 			&validator_source,
-			&Redeemer::new(
-				&RedeemerTag::new_mint(),
-				&0u32.into(),
-				&PlutusData::new_empty_constr_plutus_data(&0u32.into()),
-				ex_units,
-			),
+			&Redeemer::new(&RedeemerTag::new_mint(), &0u32.into(), &unit_plutus_data(), ex_units),
 		);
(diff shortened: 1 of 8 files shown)


toolkit/offchain/src/csl.rs
```diff
@@ -922,7 +899,13 @@ mod prop_tests {
 			protocol_parameters: protocol_parameters(),
 		};
 		let mut tx_builder = TransactionBuilder::new(&get_builder_config(&ctx).unwrap());
-		tx_builder.add_mint_one_script_token(&test_policy(), &zero_ex_units()).unwrap();
+		tx_builder
+			.add_mint_one_script_token(
+				&test_policy(),
+				&PlutusData::new_empty_constr_plutus_data(&0u32.into()),
```
kpinter-iohk, line 905: Could we define a `PlutusData::unit()` somewhere?

kpinter-iohk (approved):
LGTM with a cosmetic suggestion

toolkit/offchain/src/csl.rs
LGLO: I think that `[]` as `Constr 0 []`. This would be copying arbitrary choice from smart-contracts, but OTOH this crate is made to be compatible with our Plutus. I don't like adding code that requires comments, but I'll do it for you.

PR #409, opened 2025-01-21
feat: add static build config for MUSL

nrdxp:
# Description

Currently the node does not build totally statically due to the rocksdb crate. Newer versions of the crate do contain flags to statically compile, though, if we can find a way to update the transitive dependency cleanly.

Everything else is compiled in statically, including the MUSL C standard library

# Checklist

- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,12 @@
 [net]
 git-fetch-with-cli = true
+
+[target.'cfg(target_os = "linux")']
+linker    = "clang"
+rustflags = ["-C", "link-arg=-fuse-ld=mold"]
+
+[target.x86_64-unknown-linux-musl]
+rustflags = [
+    "-C", "target-feature=+crt-static",
+    "-C", "relocation-model=static",
+]
--- a/dev/nix/shell.nix
+++ b/dev/nix/shell.nix
@@ -19,7 +19,7 @@
       };
   in {
     devShells = {
-      default = pkgs.mkShell {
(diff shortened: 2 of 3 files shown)


skylar-simoncelli:
Nice! 

I guess we need to sit on the draft PR until we upgrade the rocksdb crate to support the static link compilation

If we get this working can then integrate the changes in the CI and CD workflows

nrdxp:
Yeah the rocksdb dep is transitive, which makes a bit more difficult to upgrade. However, if we really need this we could potentially still clean this up a touch and merge. Having only one dependency to keep track of in dynamic environments still reduces complexity and attack surface considerably

LGLO:
> if we really need this

* Nick sometimes needs partner-chains-node image to give to people so they can play with it 
* we use the same (or similar) image for our tests
 
Providing a way to build partner-chains-node with statically linked dependencies has some value, but it really is not our business as long as the "demo/test" image is providing compatible dependencies.

If it turns out increase local or CI build times, I would rather think twice about it.

nrdxp:
MUSL is lighter weight so it shouldn't really affect build times (at least it hasn't seemed to on my build server in testing).

I think the main benefit of static deps besides the improved simplicity of distribution is security, there is no way to inject a malicious dependency at runtime, which is a nice property. Still, since the current version of rocksdb leaves us hanging a bit, perhaps we can just leave off until the next update of the SDK

gilligan:
I am closing this for now. This really isn't aligned with any immediate needs or product plans for the moment. But we can refer back to this PR if that need/priority changes

PR #419, opened 2025-01-23
feat: make 'wizards' subcommand of node

LGLO:
# Description

`partner-chains-cli` separate binary is transformed to a library crated and integreted in `partner-chains-node-commands` library crate.
Every invocation of `partner-chains-cli` should be replaced with `<node> wizards` subcommand of the node built with Partner Chains SDK.
The only other change is that "node executable path" configuration is not present in `partner-chains-cli-resources.json` anymore, because it is not needed anymore.
Code will always invoke "self" executable instead.
Since this change, all functionality of Partner Chains is available in the one executable of the node.

# Checklist

- [x] Commit sequence broadly makes sense and commits have useful messages. -> Will be squashed on merge
- [x] Self-reviewed the diff

### DIFF
--- a/toolkit/partner-chains-cli/src/generate_keys/mod.rs
+++ b/toolkit/partner-chains-cli/src/generate_keys/mod.rs
@@ -11,25 +11,20 @@ use sp_core::{ed25519, Pair};
 #[cfg(test)]
 mod tests;
 
-#[derive(Debug, clap::Parser)]
+#[derive(Clone, Debug, clap::Parser)]
 pub struct GenerateKeysCmd {}
 
 #[derive(Debug)]
 pub struct GenerateKeysConfig {
 	pub chain_name: String,
 	pub substrate_node_base_path: String,
-	pub node_executable: String,
 }
 impl GenerateKeysConfig {
 	pub fn load<C: IOContext>(context: &C) -> Self {
-		// ETCM-7825: hardcoded node executable
-		let node_executable = config_fields::NODE_EXECUTABLE
-			.save_if_empty(config_fields::NODE_EXECUTABLE_DEFAULT.to_string(), context);
 		Self {
 			chain_name: DEFAULT_CHAIN_NAME.into(),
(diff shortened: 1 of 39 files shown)


toolkit/partner-chains-cli/src/generate_keys/mod.rs
```diff
@@ -242,7 +229,7 @@ pub fn generate_or_load_key<C: IOContext>(
 			&format!("A {} key already exists in store: {key} - overwrite it?", key_def.name),
 			false,
 		) {
-			let new_key = generate_keys(context, node_executable, key_def)?;
+			let new_key = generate_keys(context, &node_executable, key_def)?;
```
AmbientTea, line 232: `generate_keys` already has access to `context`, it can get the executable internally

toolkit/partner-chains-cli/src/lib.rs
```diff
@@ -59,7 +58,7 @@ pub trait CmdRun {
 	fn run<C: IOContext>(&self, context: &C) -> anyhow::Result<()>;
 }
 
-fn main() -> anyhow::Result<()> {
+pub fn run(command: Command) -> anyhow::Result<()> {
```
AmbientTea, line 61: How about `Command::run`? It doesn't have to be from the trait

AmbientTea (approved):

toolkit/partner-chains-cli/src/generate_keys/mod.rs
LGLO: Will update, thank you.

toolkit/partner-chains-cli/src/lib.rs
LGLO: Yes. And obviously setting up logging is now redundant because there is identical code in `partner-chains-node-commands` present already.

kpinter-iohk (approved):
:tada:

PR #425, opened 2025-01-27
feat: CD workflow improvements

skylar-simoncelli:
In this PR:

- Public release and Public GHCR image jobs de-coupled, so can omit either one of these independently
- S3 integration so any artifacts built will always be uploaded to a bucket, so running the workflow in the future against the same SHA can re-use these artifacts
- Removal of partner-chains-cli artifact from all composites
- Removal of workflows release.yml and ghcr.yml, as cd.yml is now flexible enough to handle these seperately and simply
- Improved wording of inputs for clarity

Scenarios tested:
- Build, deploy and create release
- Build again from same SHA to test overwriting S3 artifacts
- Deploy and create release from S3 artifacts

image


rsporny (approved):
I like it!

PR #428, opened 2025-01-28
test-refactor: do not overspecify read operations order or number

LGLO:
# Description

Step 1 of files operations overspecification clean-up. It is simple, because this is about reads only. We should not care about when something is read or how many times. Mock files are setup in tests and updated by write operations. Nothing else matters.

# Checklist

- [x] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/toolkit/partner-chains-cli/src/tests/mod.rs
+++ b/toolkit/partner-chains-cli/src/tests/mod.rs
@@ -30,7 +30,6 @@ pub enum MockIO {
 	Prompt { prompt: String, default: Option<String>, input: String },
 	PromptYN { prompt: String, default: bool, choice: bool },
 	PromptMultiOption { prompt: String, options: Vec<String>, choice: String },
-	FileRead { path: String },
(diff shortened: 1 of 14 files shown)


toolkit/partner-chains-cli/src/tests/mod.rs
```diff
@@ -30,7 +30,7 @@ pub enum MockIO {
 	Prompt { prompt: String, default: Option<String>, input: String },
 	PromptYN { prompt: String, default: bool, choice: bool },
 	PromptMultiOption { prompt: String, options: Vec<String>, choice: String },
-	FileRead { path: String },
+	//FileRead { path: String },
```
kpinter-iohk, line 33: leftover?

kpinter-iohk (approved):

toolkit/partner-chains-cli/src/tests/mod.rs
LGLO: Yes, thank you.

PR #429, opened 2025-01-29
test-refactor: do not over-specify file write interactions

LGLO:
# Description

Replaces verification for each JSON field write operation with assertions on the final shape of given file.

`pretty_assertions` are used to diff serde_json::Value, in so error message pretty.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/toolkit/partner-chains-cli/src/create_chain_spec/tests.rs
+++ b/toolkit/partner-chains-cli/src/create_chain_spec/tests.rs
@@ -1,7 +1,7 @@
 use crate::config::CHAIN_CONFIG_FILE_PATH;
 use crate::create_chain_spec::{CreateChainSpecCmd, INITIAL_PERMISSIONED_CANDIDATES_EXAMPLE};
 use crate::tests::{MockIO, MockIOContext};
-use crate::CmdRun;
+use crate::{verify_json, CmdRun};
 use anyhow::anyhow;
 use colored::Colorize;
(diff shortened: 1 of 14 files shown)


toolkit/partner-chains-cli/src/create_chain_spec/tests.rs
```diff
@@ -16,11 +16,11 @@ fn happy_path() {
 			show_initial_permissioned_candidates(),
 			MockIO::prompt_yes_no("Do you want to continue?", true, true),
 			run_build_spec_io(Ok("ok".to_string())),
-			write_updated_chain_spec_io(),
 			show_outro(),
 		]);
 	let result = CreateChainSpecCmd.run(&mock_context);
 	result.expect("should succeed");
+	verify_json!(mock_context, "chain-spec.json", updated_chain_spec())
```
LGLO, line 23: I introduced macro, because `assert_eq!` inside a function doesn't show where the error occurred. It would point to line inside verify_json function.

AmbientTea (approved):
makes so much more sense

toolkit/partner-chains-cli/src/create_chain_spec/tests.rs
AmbientTea: There's also `#[trace_caller]` but I had mixed results with it

PR #430, opened 2025-01-29
change: ETCM-9277 rename wizard config files

AmbientTea:
# Description

Changes the names of the config files used by wizard commands to be shorter.

# Checklist

- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/toolkit/partner-chains-cli/src/generate_keys/tests.rs
+++ b/toolkit/partner-chains-cli/src/generate_keys/tests.rs
@@ -5,7 +5,7 @@ use scenarios::key_file_content;
 use scenarios::resources_file_content;
 
 const DATA_PATH: &str = "/path/to/data";
-const RESOURCES_CONFIG_PATH: &str = "partner-chains-cli-resources-config.json";
+const RESOURCES_CONFIG_PATH: &str = "pc-resources-config.json";
 const KEYS_FILE: &str = "partner-chains-public-keys.json";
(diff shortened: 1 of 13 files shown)


toolkit/partner-chains-cli/src/generate_keys/tests.rs
```diff
@@ -5,7 +5,7 @@ use scenarios::key_file_content;
 use scenarios::resources_file_content;
 
 const DATA_PATH: &str = "/path/to/data";
-const RESOURCES_CONFIG_PATH: &str = "partner-chains-cli-resources-config.json";
```
LGLO, line 8: Cosmetic: we could get rid of this and reference it from `config` crate in every place.

toolkit/partner-chains-cli/src/register/register1.rs
```diff
@@ -396,8 +396,8 @@ mod tests {
 		assert!(result.is_err());
 	}
 
-	const CHAIN_CONFIG_PATH: &str = "partner-chains-cli-chain-config.json";
-	const RESOURCE_CONFIG_PATH: &str = "partner-chains-cli-resources-config.json";
+	const CHAIN_CONFIG_PATH: &str = "pc-chain-config.json";
```
LGLO, line 399: Defined again. I understand that tests maybe would like to add extra check if production code uses expected name, but it's funny that we still use the literal on line 234.

PR #450, opened 2025-02-04
chore: update polkadot-sdk dependency to polkadot-stable-2412

LGLO:
# Description

## Updates polkadot-sdk to stable-2412

`RuntimeString` and `create_runtime_str!` are deprecated, using `String` and `Cow::Borrowed` from `alloc` instead.

`U256` and `[u8;32]` `From` instances were removed. More explicit `from_big_endian` is used instead.

Benchmarking command of node is removed because `frame-omni-bencher` is supposed to run runtime benchmarks. Supporting crate became GPLv3 without linking exception. Our benchmarking mechanism will have to be updated in the follow up.

# Checklist

- [x] Self-reviewed the diff


node/node/Cargo.toml
```diff
@@ -108,18 +106,3 @@ authority-selection-inherents = { workspace = true, features = ["mock"] }
...
 [features]
 default = ["sidechain-domain/std"]
-rocksdb = [
-	"sc-cli/rocksdb",
-	"sc-service/rocksdb",
-	"sidechain-runtime/with-rocksdb-weights",
-]
-paritydb = [
-	"sidechain-runtime/with-paritydb-weights",
-]
```
LGLO, line 118: I've removed the features to go leaner in our demo node. We use only rocksdb, so removing these features.

changelog.md
```diff
@@ -4,6 +4,8 @@ This changelog is based on [Keep A Changelog](https://keepachangelog.com/en/1.1.
 
 # Unreleased
 
+* Update polkadot-sdk to polkadot-stable2412-1
```
kpinter-iohk, line 7: Is the removal of benchmark commands worth mentioning?

kpinter-iohk (approved):

node/runtime/src/lib.rs
```diff
@@ -288,6 +287,8 @@ impl frame_system::Config for Runtime {
 	type AccountData = pallet_balances::AccountData<Balance>;
 	/// Weight information for the extrinsics of this pallet.
 	type SystemWeightInfo = frame_system::weights::SubstrateWeight<Runtime>;
+	// WeightInfo for extensions is present but not accessible in polkadot-stable2412-1, because of that we are using () in our demo runtime
+	type ExtensionsWeightInfo = ();
```
Klapeyron, line 291:
```suggestion
	type ExtensionsWeightInfo = frame_system_extensions::weights::WeightInfo<Runtime>;
```
it will compile?

Klapeyron: ok, seems not :/

Klapeyron (approved):

PR #451, opened 2025-02-04
ETCM-9199: better error when connecting to ogmios fails

kpinter-iohk:
# Description

Adds more descriptive error for when connecting to Ogmios client fails.

previously:
```
Error: Application("Couldn't create WebSockets client: Error when opening the TCP socket: Connection refused (os error 61)")
```
with the change:
```
Error: Application("Failed to connect to Ogmios at ws://localhost:1337 with: Couldn't create WebSockets client: Error when opening the TCP socket: Connection refused (os error 111)")
```

# Checklist

- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/toolkit/cli/smart-contracts-commands/src/lib.rs
+++ b/toolkit/cli/smart-contracts-commands/src/lib.rs
@@ -1,3 +1,4 @@
+use ogmios_client::jsonrpsee::{client_for_url, OgmiosClients};
 use partner_chains_cardano_offchain::cardano_keys::{
 	CardanoKeyFileContent, CardanoPaymentSigningKey,
 };
@@ -36,6 +37,14 @@ pub enum SmartContractsCmd {
 pub struct CommonArguments {
 	#[arg(default_value = "ws://localhost:1337", long, short = 'O')]
 	ogmios_url: String,
+}
+
+impl CommonArguments {
+	pub async fn get_ogmios_client(&self) -> crate::CmdResult<OgmiosClients> {
+		Ok(client_for_url(&self.ogmios_url).await.map_err(|e| {
+			format!("Failed to connect to Ogmios at {} with: {}", &self.ogmios_url, e)
+		})?)
+	}
 }
 
 type CmdResult<T> = Result<T, Box<dyn std::error::Error + Send + Sync>>;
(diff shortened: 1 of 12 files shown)


toolkit/cli/smart-contracts-commands/src/lib.rs
```diff
@@ -38,6 +39,14 @@ pub struct CommonArguments {
 	ogmios_url: String,
 }
 
+impl CommonArguments {
+	pub async fn get_ogmios_client(&self) -> crate::CmdResult<OgmiosClients> {
```
kpinter-iohk, line 43: I originally made this a standalone function but it's a bit more convenient to use like this.

toolkit/cli/smart-contracts-commands/src/lib.rs
```diff
@@ -38,6 +39,14 @@ pub struct CommonArguments {
...
+		Ok(client_for_url(&self.ogmios_url).await.map_err(|e| {
+			format!("Failed to connect to Ogmios at {} with: {}", &self.ogmios_url, e)
+		})?)
```
LGLO, line 46: Could you provide examples of errors before and after? What is displayed to user when for example `get-scripts` subcommand is used.

LGLO: More precisely:
```
./target/debug/partner-chains-node smart-contracts get-scripts -c f8fbe7316561e57de9ecd1c86ee8f8b512a314ba86499ba9a584bfa8fe2edc8d#0 --ogmios-url http://localhost:1337
```

kpinter-iohk: Added to description (but using websocket url from a different example i've been using)

LGLO: This is the clue of my comment. There are basically only two path to test, please test both of them.

kpinter-iohk: I thought you wanted to see the exact error output. There are a lot more than two paths to test, since each smart-contract command is touched by this, and also the wizards. The error message from `client_for_url` is prepended with additional info regardless if the ogmios address is HTTP or WS, so I see no sense in testing with different URLs, as long as they are unreachable.

kpinter-iohk: To address the issue that HTTP connection is not validated until the first HTTP call, I added a call to `get_tip` to client initialization. (It seems like a low cost call, and using something else that seems more appropriate like `/health` would require adding a lot of boilerplate just for this one use case.)

LGLO (approved):

PR #458, opened 2025-02-05
bugfix: fix reserve release panicking with overflow error

LGLO:
# Description

Fixes the overflow error when user passes `--amount` lower than total of already release amount.

Before:
```
thread 'main' panicked at toolkit/offchain/src/reserve/release.rs:109:30:
attempt to subtract with overflow
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

After:
```
Error: Application(The requested total amount: 2 is not greater than the amount already transferred to illiquid supply: 10!)
```

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/toolkit/offchain/src/reserve/release.rs
+++ b/toolkit/offchain/src/reserve/release.rs
@@ -106,11 +106,12 @@ fn reserve_release_tx(
 	let mut tx_builder = TransactionBuilder::new(&get_builder_config(ctx)?);
 
 	let reserve_balance = previous_reserve.utxo.get_asset_amount(&token);
-	let amount_to_transfer = cumulative_total_transfer - stats.token_total_amount_transferred;
-	if amount_to_transfer as i128 > reserve_balance {
-		return Err(anyhow!("Not enough funds in the reserve to transfer {amount_to_transfer} tokens (reserve balance: {reserve_balance})"));
-	}
-	let left_in_reserve: u64 = (reserve_balance - (amount_to_transfer as i128)) as u64;
+	let token_total_amount_transferred = stats.token_total_amount_transferred;
+	let amount_to_transfer = cumulative_total_transfer.checked_sub(token_total_amount_transferred)
+		.filter(|x| *x > 0)
+		.ok_or_else(||anyhow!("The requested total amount: {cumulative_total_transfer} is not greater than the amount already transferred to illiquid supply: {token_total_amount_transferred}.",))?;
+	let left_in_reserve = reserve_balance.checked_sub(amount_to_transfer)
+		.ok_or_else(||anyhow!("Not enough funds in the reserve to transfer {amount_to_transfer} tokens (reserve balance: {reserve_balance})"))?;
 
 	// Additional reference scripts
 	tx_builder.add_script_reference_input(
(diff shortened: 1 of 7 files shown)


toolkit/offchain/src/reserve/release.rs
```diff
@@ -106,7 +106,13 @@ fn reserve_release_tx(
...
 
 	let reserve_balance = previous_reserve.utxo.get_asset_amount(&token);
-	let amount_to_transfer = cumulative_total_transfer - stats.token_total_amount_transferred;
+	let token_total_amount_transferred = stats.token_total_amount_transferred;
+	if cumulative_total_transfer <= token_total_amount_transferred {
+		return Err(anyhow!(
+			"The requested total amount: {cumulative_total_transfer} is not greater than the amount already transferred to illiquid supply: {token_total_amount_transferred}.",
+		));
+	}
+	let amount_to_transfer = cumulative_total_transfer - token_total_amount_transferred;
```
LGLO, line 115: I'm wondering about moving these calculation out of this function, and passing `amount_to_transfer` and `left_in_reserve`. This would allow logging `amount_to_transfer` to user one time. OTOH this need should go away if we fix UI.

Klapeyron: what about somethnig like this?
```rust
	let amount_to_transfer = cumulative_total_transfer
		.checked_sub(token_total_amount_transferred)
		.ok_or(anyhow!("Overflow"))?;
```

LGLO: Yes, more idiomatic. But the question about place where to check user input remains. I tend to go near the edge.
Pros for having it here is that it is easier to test, because no mocks are involved.

Klapeyron: If there will be an option to check it earlier that would be great, but we should be also careful to not have a check in one place of the codebase and computations that make assumptions on top of it in the other. If we will not eliminate a subtraction, then keeping additional check here is a good idea

LGLO: I meant to pass result of subtraction to this function

AmbientTea (approved):

toolkit/offchain/src/reserve/release.rs
```diff
@@ -106,7 +106,13 @@ fn reserve_release_tx(
...
 	if amount_to_transfer as i128 > reserve_balance {
 		return Err(anyhow!("Not enough funds in the reserve to transfer {amount_to_transfer} tokens (reserve balance: {reserve_balance})"));
 	}
```
Klapeyron, line 118: I can't make suggestion for line 119, consider something like this:
```suggestion
	let reserve_balance: u64 = previous_reserve
		.utxo
		.get_asset_amount(&token)
		.try_into()
		.expect("Expected reserve_balance to be non-negative");

	let token_total_amount_transferred = stats.token_total_amount_transferred;
	let amount_to_transfer = cumulative_total_transfer.checked_sub(token_total_amount_transferred).ok_or_else(|| anyhow!("The requested total amount: {cumulative_total_transfer} is not greater than the amount already transferred to illiquid supply: {token_total_amount_transferred}."))?;

	let left_in_reserve = reserve_balance.checked_sub(amount_to_transfer).ok_or_else(|| anyhow!("Not enough funds in the reserve to transfer {amount_to_transfer} tokens (reserve balance: {reserve_balance})"))?;
```

LGLO: I've updated ogmios client and we don't need to `try_into()`

Klapeyron (approved):

PR #460, opened 2025-02-05
use fixed length array in MainchainSignature

kpinter-iohk:
# Description

Changes MainchainSignature to use a fixed length array instead of a vec.

# Checklist

- [ ] Self-reviewed the diff

# Note on CI
for you.


toolkit/primitives/domain/src/lib.rs
```diff
@@ -300,9 +300,12 @@ impl MainchainKeyHash {
 	}
 }
 
+/// ECDSA signature of MainchainPrivateKey, 64 bytes.
+const MAINCHAIN_SIGNATURE_LEN: usize = MAINCHAIN_PRIVATE_KEY_LEN * 2;
```
Klapeyron, line 304: can we remove `pub` attribute from `MAINCHAIN_KEY_HASH_LEN` in line 277 once we are here?

toolkit/primitives/plutus-data/src/registered_candidates.rs
```diff
@@ -176,7 +176,9 @@ fn decode_ada_based_staking_datum(datum: PlutusData) -> Option<AdaBasedStaking>
 		.filter(|datum| datum.data().len() >= 2)?
 		.data();
 	let pub_key = TryFrom::try_from(fields.get(0).as_bytes()?).ok()?;
-	let signature = MainchainSignature(fields.get(1).as_bytes()?);
+	let signature = MainchainSignature(
+		fields.get(1).as_bytes()?.try_into().expect("Mainchain signature is 64 bytes"),
```
AmbientTea, line 180: We can't 'expect' here because anyone can put any blob of bytes in the datum. Use `.ok()?` instead to return `None`.

AmbientTea (changes requested):
See my note

toolkit/primitives/plutus-data/src/registered_candidates.rs
kpinter-iohk: Good catch, I didn't notice that we are inside an `Option`.

toolkit/primitives/domain/src/lib.rs
LGLO, line 304: Is there any relation between private key length and signature length?
Why not use literal 64 ?
Why not use 2*MAINCHAIN_PUBLIC_KEY_LEN ?

toolkit/primitives/authority-selection-inherents/src/filter_invalid_candidates.rs
```diff
@@ -381,7 +381,9 @@ mod tests {
 			let registration_data = RegistrationData {
 				registration_utxo: signed_message.registration_utxo,
 				sidechain_signature: SidechainSignature(sidechain_signature),
-				mainchain_signature: MainchainSignature(mainchain_signature.0.to_vec()),
+				mainchain_signature: MainchainSignature(
+					mainchain_signature.0.try_into().expect("TODO"),
```
LGLO, line 385: TODO left

jankun4 (approved):

AmbientTea (approved):

PR #462, opened 2025-02-05
feat: make 'reserve release' --amount parameter the amount to release by this execution command

LGLO:
# Description

Changes `--amount` to the amount that should be released from reserve during this execution of command, not the total amount from the creation of the reserve.

Additionally, `--token` parameter is removed as it was not used by the present code. I don't know if it should or rather it should be removed from `deposit` for example. For sure users would be surprised if they given some asset id but something else was released.

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/toolkit/cli/smart-contracts-commands/src/reserve.rs
+++ b/toolkit/cli/smart-contracts-commands/src/reserve.rs
@@ -12,6 +12,7 @@ use partner_chains_cardano_offchain::{
 	},
 };
 use sidechain_domain::{AssetId, ScriptHash, UtxoId};
+use std::num::NonZero;
 
 #[derive(Clone, Debug, clap::Subcommand)]
 #[allow(clippy::large_enum_variant)]
@@ -207,12 +208,9 @@ pub struct ReleaseReserveCmd {
 	/// Reference UTXO containing the V-Function script
 	#[arg(long, short('r'))]
 	reference_utxo: UtxoId,
-	/// Encoded asset id in form <policy_id_hex>.<asset_name_hex>.
+	/// Amount of reserve tokens to be released to the illiquid supply.
 	#[arg(long)]
-	token: AssetId,
-	/// Current value of the V-Function
-	#[arg(long)]
-	amount: u64,
+	amount: NonZero<u64>,
 }
 
 impl ReleaseReserveCmd {
@@ -220,7 +218,7 @@ impl ReleaseReserveCmd {
 		let payment_key = self.payment_key_file.read_key()?;
 		let client = self.common_arguments.get_ogmios_client().await?;
 		let _ = release_reserve_funds(
-			TokenAmount { token: self.token, amount: self.amount },
+			self.amount,
 			self.genesis_utxo,
 			self.reference_utxo,
 			&payment_key,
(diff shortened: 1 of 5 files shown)


toolkit/offchain/src/reserve/release.rs
```diff
@@ -459,12 +460,12 @@ mod tests {
...
 			.expect("Should transfer reserve token to illiquid supply");
 
-		assert_eq!(illiquid_supply_output, 10u64.into())
+		assert_eq!(illiquid_supply_output, 5u64.into())
 	}
 
 	#[test]
 	fn should_leave_unreleased_tokens_at_reserve_validator() {
-		let illiquid_supply_output = (reserve_release_test_tx().body().outputs().into_iter())
+		let validator_output = (reserve_release_test_tx().body().outputs().into_iter())
```
LGLO, line 468: Yeah, imagine I was wondering why after effectively decreasing the transfer amount from 10 to 5, I've seen the illiquid supply output to grow.

toolkit/offchain/src/reserve/release.rs
```diff
@@ -392,7 +393,7 @@ mod tests {
 			&reserve_data(),
 			&previous_reserve_utxo(),
 			&reference_utxo(),
-			20,
+			5,
```
LGLO, line 396: This 20 meant 10 over the other 10 from `previous_reserve_utxo()`, to not be confused by two different 10s, I've changed this to 5.

toolkit/cli/smart-contracts-commands/src/reserve.rs
```diff
@@ -208,20 +209,17 @@ pub struct ReleaseReserveCmd {
...
 	#[arg(long, short('r'))]
 	reference_utxo: UtxoId,
-	/// Encoded asset id in form <policy_id_hex>.<asset_name_hex>.
+	/// Amount of reserve tokens to be released to the illiquid supply.
 	#[arg(long)]
-	token: AssetId,
-	/// Current value of the V-Function
-	#[arg(long)]
-	amount: u64,
+	amount: NonZero<u64>,
```
LGLO, line 214: User who tries to use `--amount 0`, is rewarded with:
```
error: invalid value '0' for '--amount <AMOUNT>': number would be zero for non-zero type
```

toolkit/offchain/src/reserve/release.rs
```diff
@@ -96,20 +95,21 @@ fn reserve_release_tx(
...
 	let mut tx_builder = TransactionBuilder::new(&get_builder_config(ctx)?);
 
 	let reserve_balance = previous_reserve.utxo.get_asset_amount(&token);
 	let token_total_amount_transferred = stats.token_total_amount_transferred;
-	let amount_to_transfer = cumulative_total_transfer.checked_sub(token_total_amount_transferred)
-		.filter(|x| *x > 0)
-		.ok_or_else(||anyhow!("The requested total amount: {cumulative_total_transfer} is not greater than the amount already transferred to illiquid supply: {token_total_amount_transferred}.",))?;
+	let cumulative_total_transfer: u64 = token_total_amount_transferred
+		.checked_add(amount_to_transfer)
+		.expect("cumulative_total_transfer overflow u64");
```
LGLO, line 111: I know this can be controversial. Problem is that with the current CSL it will be really hard to make it u128. In other words our implementation is not ready for going over u64. Therefore panic.

AmbientTea (approved):

PR #475, opened 2025-02-07
feat: add block-production-log primitives and pallet

LGLO:
# Description

block-production-log pallet.

Node will not work, because inherent data provider is not wired it yet
Unit tests  for `take_prefix` are missing

# Checklist

- [ ] Self-reviewed the diff


toolkit/pallets/block-production-log/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+		#[pallet::constant]
+		type MaxLogSize: Get<u32>;
+
+		#[pallet::constant]
+		type MaxProducerIdLength: Get<u32>;
+
+		type BlockProducerId: Member
+			+ Parameter
+			+ MaxEncodedLen
+			+ From<BoundedVec<u8, Self::MaxProducerIdLength>>;
```
LGLO, line 39: I admit I wish to not have this one, added only because without a `From` I had a problem creating instance in benchmarks/tests.

toolkit/pallets/block-production-log/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		#[pallet::call_index(0)]
+		#[pallet::weight((T::WeightInfo::append(), DispatchClass::Mandatory))]
+		pub fn append(origin: OriginFor<T>, id: T::BlockProducerId) -> DispatchResult {
+			ensure_none(origin)?;
+			let slot = T::current_slot();
+			Log::<T>::try_append((slot, id)).map_err(|_| sp_runtime::DispatchError::BadOrigin)
```
LGLO, line 83: TODO: what error when user has not called `take_prefix` on time and the new item doesn't fit in BoundedVec?

toolkit/primitives/authority-selection-inherents/src/ariadne_inherent_data_provider.rs
```diff
@@ -2,11 +2,10 @@
 use crate::authority_selection_inputs::AuthoritySelectionDataSource;
 use crate::authority_selection_inputs::AuthoritySelectionInputs;
 use parity_scale_codec::{Decode, Encode};
-use sidechain_domain::mainchain_epoch::MainchainEpochConfig;
 #[cfg(feature = "std")]
 use {
 	crate::authority_selection_inputs::AuthoritySelectionInputsCreationError,
-	sidechain_domain::mainchain_epoch::MainchainEpochDerivation,
+	sidechain_domain::mainchain_epoch::{MainchainEpochConfig, MainchainEpochDerivation},
```
LGLO, line 8: Unsolicited fix of unused import warning

toolkit/pallets/block-production-log/src/lib.rs
Klapeyron: It is possible to add additional parameterizations in benchmarks, maybe it will work:
https://paritytech.github.io/polkadot-sdk/master/frame_benchmarking/v2/index.html#where-clause

toolkit/pallets/block-production-log/src/test.rs
```diff
@@ -0,0 +1,39 @@
...
+	new_test_ext().execute_with(|| {
+		let existing_item = vec![(Slot::from(100), make_id::<Test>(1))];
+		Log::<Test>::put(
+			BoundedVec::<_, <mock::Test as pallet::Config>::MaxLogSize>::truncate_from(
+				existing_item,
+			),
+		);
+
+		let call = Call::<Test>::append { id: make_id::<Test>(2) };
+		call.dispatch_bypass_filter(RuntimeOrigin::none()).unwrap();
```
Klapeyron, line 28: consider call without bypass of the filter
```suggestion
		assert_ok!(BlockProductionLog::append(RuntimeOrigin::none(), make_id::<Test>(2)));
```

toolkit/pallets/block-production-log/src/test.rs
```diff
@@ -0,0 +1,39 @@
...
+use frame_support::traits::UnfilteredDispatchable;
+use mock::*;
+use sp_consensus_slots::Slot;
+use sp_runtime::BoundedVec;
+
+#[test]
+fn first_append_should_succeed() {
+	new_test_ext().execute_with(|| {
+		let call = Call::<Test>::append { id: make_id::<Test>(1) };
+		call.dispatch_bypass_filter(RuntimeOrigin::none()).unwrap();
```
Klapeyron, line 11: same here

toolkit/pallets/block-production-log/src/test.rs
```diff
@@ -0,0 +1,39 @@
...
+
+		assert_eq!(Log::<Test>::get().to_vec(), vec![(Slot::from(1001000), make_id::<Test>(1))]);
+	})
+}
+
+#[test]
+fn append_to_end_of_log() {
```
Klapeyron, line 18: maybe we can merge both testcases, what about following scenario:
1) we test if vec is empty
2) we append something
3) we check if there is elem
4) we append again something else
5) we check if there are both

toolkit/pallets/block-production-log/src/lib.rs
```diff
@@ -0,0 +1,94 @@
...
+		pub fn append(origin: OriginFor<T>, id: T::BlockProducerId) -> DispatchResult {
+			ensure_none(origin)?;
+			let slot = T::current_slot();
+			Log::<T>::try_append((slot, id)).map_err(|_| sp_runtime::DispatchError::Exhausted)
+		}
+	}
+
+	impl<T: Config> Pallet<T> {
+		pub fn take_prefix(slot: &Slot) -> Vec<(Slot, T::BlockProducerId)> {
+			let (to_return, to_retain) = Log::<T>::take().into_iter().partition(|(s, _)| s <= slot);
```
Klapeyron, line 87: maybe it is worth to consider storing data ordered, so taking of prefix may be faster:
https://paritytech.github.io/polkadot-sdk/master/frame_support/traits/trait.SortedMembers.html

Klapeyron: for instance pallet-membership maintains vector and exposes it via this trait 
https://github.com/paritytech/polkadot-sdk/blob/master/substrate/frame/membership/src/lib.rs#L394

toolkit/primitives/block-production-log/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+		&self,
+		identifier: &InherentIdentifier,
+		mut error: &[u8],
+	) -> Option<Result<(), sp_inherents::Error>> {
+		if *identifier == INHERENT_IDENTIFIER {
+			let error = InherentError::decode(&mut error).ok()?;
+			Some(Err(sp_inherents::Error::Application(Box::from(error))))
+		} else {
+			None
+		}
```
Klapeyron, line 75: boolean has helpfull [method](https://doc.rust-lang.org/std/primitive.bool.html#method.then) for this

toolkit/primitives/block-production-log/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+
+#[cfg(test)]
+mod test;
+
+use parity_scale_codec::Encode;
+use sp_inherents::{InherentIdentifier, IsFatalError};
+#[cfg(feature = "std")]
+use {parity_scale_codec::Decode, sp_inherents::InherentData};
+
+pub const INHERENT_IDENTIFIER: InherentIdentifier = *b"blprdlog";
```
LGLO, line 11: :shushing_face:

toolkit/pallets/block-production-log/src/lib.rs
LGLO: Works like a charm, thank you!

toolkit/pallets/block-production-log/src/lib.rs
```diff
@@ -0,0 +1,92 @@
...
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		#[pallet::call_index(0)]
+		#[pallet::weight((T::WeightInfo::append(), DispatchClass::Mandatory))]
+		pub fn append(origin: OriginFor<T>, id: T::BlockProducerId) -> DispatchResult {
+			ensure_none(origin)?;
+			let slot = T::current_slot();
+			Log::<T>::try_append((slot, id)).map_err(|_| sp_runtime::DispatchError::Exhausted)
+		}
```
Klapeyron, line 82: we should probably prevent somehow multiple executions of this call per block, malicious validator may spam otherwise

Klapeyron: it may be good scenario for a test as well, double execution for the same block and other for pushing of the same id twice maybe

toolkit/pallets/block-production-log/src/test.rs
```diff
@@ -0,0 +1,39 @@
...
+	new_test_ext().execute_with(|| {
+		let existing_item = vec![(Slot::from(100), make_id::<Test>(1))];
+		Log::<Test>::put(
+			BoundedVec::<_, <mock::Test as pallet::Config>::MaxLogSize>::truncate_from(
+				existing_item,
+			),
+		);
+
```
Klapeyron, line 26: just in case, block number can be modified via `System::set_block_number(1);` call

toolkit/pallets/block-production-log/src/test.rs
```diff
@@ -0,0 +1,39 @@
+use super::*;
+use frame_support::traits::UnfilteredDispatchable;
+use mock::*;
+use sp_consensus_slots::Slot;
+use sp_runtime::BoundedVec;
+
+#[test]
```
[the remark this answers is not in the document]
LGLO: Unit tests are TBD, I've opened this draft to have review before critically important reviewers go for vacations.

toolkit/pallets/block-production-log/src/lib.rs
LGLO: They use exactly same structure as I do. 
```
#[pallet::storage]
pub type Members<T: Config<I>, I: 'static = ()> =
     StorageValue<_, BoundedVec<T::AccountId, T::MaxMembers>, ValueQuery>;
```
Having said that I'll ensure that `slot` is greater than the last currently stored.
This will at same time fix the problem with possible multiple calls.

Klapeyron: Right. I am not sure who will be using `take_prefix`, but maybe it will be worth to change it and expose it via SortedMembers, so end user will be able to make some additional operations on it, without adjusting logic of this pallet.
I am currious if something like `pub fn take_prefix(slot: &Slot) -> impl SortedMembers<..> {` will work

toolkit/primitives/block-production-log/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+	) -> Result<(), sp_inherents::Error> {
+		inherent_data.put_data(INHERENT_IDENTIFIER, &self.id)
+	}
+
+	async fn try_handle_error(
+		&self,
+		identifier: &InherentIdentifier,
+		mut error: &[u8],
+	) -> Option<Result<(), sp_inherents::Error>> {
+		if *identifier == INHERENT_IDENTIFIER {
```
Klapeyron, line 70:
```suggestion
		if identifier == &INHERENT_IDENTIFIER {
```

toolkit/pallets/block-production-log/src/test.rs
LGLO: Appending twice is a negative test case. Since we have constant current_slot, it is illegal to run it twice.

toolkit/pallets/block-production-log/src/lib.rs
LGLO: If only `append` is use, this is sorted all the time. Appending with slot lower than the latest slot is illegal.

toolkit/primitives/block-production-log/src/lib.rs
LGLO: I have tried. It doesn't work well with `?` macro. Old school if-else does.

toolkit/pallets/block-production-log/src/lib.rs
LGLO: Implemented!

toolkit/pallets/block-production-log/src/test.rs
LGLO: Thank you!
I do not care about block numbers here. This pallet keeps mapping from Slot numbers to some identifiers.

toolkit/pallets/block-production-log/src/test.rs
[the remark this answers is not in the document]
LGLO: Done

toolkit/primitives/block-production-log/src/lib.rs
LGLO: Done

toolkit/pallets/block-production-log/src/lib.rs
```diff
@@ -0,0 +1,112 @@
...
+	pub type CurrentBlockProducerId<T: Config> = StorageValue<_, T::BlockProducerId, OptionQuery>;
+
+	#[pallet::storage]
+	pub type Log<T: Config> =
+		StorageValue<_, BoundedVec<(Slot, T::BlockProducerId), T::MaxLogSize>, ValueQuery>;
+
+	/// This storage protects against calling `append` multiple times for the same slot.
+	/// It is required because `Log` can be emptied and there is no other way to remember the last stored slot.
+	#[pallet::storage]
+	pub type LatestSlot<T: Config> = StorageValue<_, Slot, OptionQuery>;
```
Klapeyron, line 51: if we will assume that starting point is 0, then switching from `OptionQuery` to `ValueQuery` will allow to skip check if value exists

toolkit/pallets/block-production-log/src/lib.rs
Klapeyron: Right, but as we discussed, once it is sorted then we can try also something like binary_search instead of partition

toolkit/pallets/block-production-log/src/test.rs
Klapeyron: hmm, but what if we will advance blocks in the mean time, so there will be new session?

toolkit/pallets/block-production-log/src/test.rs
```diff
@@ -0,0 +1,138 @@
...
+		LatestSlot::<Test>::put(Slot::from(100));
+		Log::<Test>::put(
+			BoundedVec::<_, <mock::Test as pallet::Config>::MaxLogSize>::truncate_from(
+				existing_item,
+			),
+		);
+
+		let call = Call::<Test>::append { block_producer_id: make_id(2) };
+		call.clone().dispatch_bypass_filter(RuntimeOrigin::none()).unwrap();
+		let err = call.dispatch_bypass_filter(RuntimeOrigin::none()).unwrap_err();
```
Klapeyron, line 57:
```suggestion
		assert_err!(BlockProductionLog::append(RuntimeOrigin::none(), make_id(2)), Error::<Test>::SlotNotIncreased);
```

toolkit/pallets/block-production-log/src/test.rs
```diff
@@ -0,0 +1,138 @@
...
+
+		assert!(format!("{:?}", err.error).contains("SlotNotIncreased"),);
```
Klapeyron, line 59: this can be removed then

toolkit/pallets/block-production-log/src/test.rs
```diff
@@ -0,0 +1,138 @@
...
+		let existing_item = vec![(Slot::from(100), make_id(1))];
+		LatestSlot::<Test>::put(Slot::from(100));
+		Log::<Test>::put(
+			BoundedVec::<_, <mock::Test as pallet::Config>::MaxLogSize>::truncate_from(
+				existing_item,
+			),
+		);
+
+		let call = Call::<Test>::append { block_producer_id: make_id(2) };
+		assert_ok!(call.dispatch_bypass_filter(RuntimeOrigin::none()));
```
Klapeyron, line 35:
```suggestion
		assert_ok!(BlockProductionLog::append(RuntimeOrigin::none(), make_id(2)));
```

toolkit/pallets/block-production-log/src/test.rs
```diff
@@ -0,0 +1,138 @@
...
+#[test]
+fn can_not_append_same_item_twice() {
+	new_test_ext().execute_with(|| {
+		let existing_item = vec![(Slot::from(100), make_id(1))];
+		LatestSlot::<Test>::put(Slot::from(100));
+		Log::<Test>::put(
+			BoundedVec::<_, <mock::Test as pallet::Config>::MaxLogSize>::truncate_from(
+				existing_item,
+			),
+		);
```
Klapeyron, line 53: for BoundedVec setup consider BoundedSlice
```suggestion
		let binding = vec![(Slot::from(100), make_id(1))];
		let existing_item = BoundedSlice::try_from(binding.as_slice()).unwrap();
		LatestSlot::<Test>::put(Slot::from(100));
		Log::<Test>::put(existing_item);
```

toolkit/pallets/block-production-log/src/test.rs
LGLO: Slots are monotonic. They are not going to 0 at new session in any meaning of session

toolkit/pallets/block-production-log/src/lib.rs
LGLO: `partition` is `O(n)` . Even if we could use something like `binary_search` we still have to copy either one or another part or both parts, so it still has same complexity. At best we could optimize for some scenario when assume that part to retain is much smaller, but I think it won't be easy to best this very simple solution with `partition`.

BTW `binary_search` requires exact hit also, as I allowed multiple blocks in same slot, it is hard to get the "right-most" one.

toolkit/pallets/block-production-log/src/test.rs
LGLO: Done, thanks

node/runtime/src/lib.rs
```diff
@@ -456,6 +465,18 @@ impl pallet_block_rewards::Config for Runtime {
 	type GetBlockRewardPoints = sp_block_rewards::SimpleBlockCount;
 }
 
+impl pallet_block_production_log::Config for Runtime {
+	type MaxLogSize = ConstU32<1024>;
+	type MaxProducerIdLength = ConstU32<32>;
```
AmbientTea, line 470: is this limit here necessary? `BlockProducerId` already needs to be sized

toolkit/pallets/block-production-log/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+			origin: OriginFor<T>,
+			block_producer_id: T::BlockProducerId,
+		) -> DispatchResult {
+			ensure_none(origin)?;
+			let current_block = <frame_system::Pallet<T>>::block_number();
+			let slot = T::current_slot();
+			match LatestBlock::<T>::get() {
+				Some(b) if b >= current_block => Err(Error::<T>::BlockNumberNotIncreased),
+				_ => Ok(()),
+			}?;
```
AmbientTea, line 95: I don't think this situation is possible in a real chain

node/runtime/src/lib.rs
AmbientTea: This parameter also looks to be unused anyway

toolkit/pallets/block-production-log/src/test.rs
```diff
@@ -0,0 +1,142 @@
...
+		System::set_block_number(1011);
+
+		let call = Call::<Test>::append { block_producer_id: make_id(2) };
+		call.clone().dispatch_bypass_filter(RuntimeOrigin::none()).unwrap();
+		assert_err!(
+			call.dispatch_bypass_filter(RuntimeOrigin::none()),
+			Error::<Test>::BlockNumberNotIncreased
+		);
+	})
+}
```
AmbientTea, line 53: This can't happen unless someone explicitly calls the inherent a second time using `sudo`, at which point they're screwing with their own chain.

toolkit/pallets/block-production-log/src/lib.rs
```diff
@@ -0,0 +1,109 @@
...
+			LatestBlock::<T>::put(current_block);
+			Ok(Log::<T>::try_append((slot, block_producer_id))
+				.map_err(|_| Error::<T>::StorageFull)?)
```
AmbientTea, line 98: This will brick the chain. Maybe we should think about some other way of handling that (eg. rotating out old log entries)

toolkit/pallets/block-production-log/src/test.rs
```diff
@@ -0,0 +1,142 @@
...
+		);
+		assert_eq!(
+			left_in_storage.to_vec(),
+			vec![(Slot::from(105), make_id(2)), (Slot::from(107), make_id(1)),]
+		);
+	})
+}
+
+#[test]
+fn take_prefix_when_there_are_two_entries_for_the_same_slot() {
```
AmbientTea, line 115: We also have a test for not allowing that to happen. :P

AmbientTea (approved):

toolkit/pallets/block-production-log/src/test.rs
LGLO: It can happen if consensus allows to have multiple blocks in one slot. There is no reason to disallow it, right?

toolkit/primitives/block-production-log/src/lib.rs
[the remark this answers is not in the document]
LGLO: This is one more powerful as it doesn't aggregate and lets users consume full log.

node/runtime/src/lib.rs
LGLO: Of course, thanks for spotting. I'll remove it.

toolkit/pallets/block-production-log/src/lib.rs
LGLO: As discussed, there exists some Polkadot documentation that suggests we should do it. We can revisit it and remove this check in runtime code. It won't be breaking change.

toolkit/pallets/block-production-log/src/lib.rs
LGLO: Changed to `#[pallet::unbounded]`.

PR #484, opened 2025-02-10
Revert "use fixed length array in MainchainSignature"

LGLO:
It is not backward compatible


AmbientTea:
Please don't revert it yet. We can write a backwards-compatible implementation of Encode/Decode and leave this change in.

PR #492, opened 2025-02-12
better errors when governance policy rejects transaction

LGLO:
# Description

Sketch of giving user a hint about governance policy failure


kpinter-iohk:
Regarding the generated Ogmios types:
It makes sense to have them. They should be in their own crate.
There should be a script that pieces together the required json schema with jq and executes typify on it so updating is less painful if there is a schema change. I'd be willing to write this script.

PR #494, opened 2025-02-13
add: envrc file exposing local environment parameters

AmbientTea:
# Description

Mounts the shared directory of the local env to a directory accessible from host and adds an `envrc` file that can be sourced to load the custom Cardano and PC parameters.

# Checklist

- [ ] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/dev/local-environment/configurations/cardano/entrypoint.sh
+++ b/dev/local-environment/configurations/cardano/entrypoint.sh
@@ -48,6 +48,7 @@ echo "Updated ConwayGenesisHash value in config files to: $conway_hash"
 
 byron_startTimeMillis=$(($byron_startTime * 1000))
 echo $byron_startTimeMillis > /shared/MC__FIRST_EPOCH_TIMESTAMP_MILLIS
+echo $byron_startTimeMillis > /runtime-values/MC__FIRST_EPOCH_TIMESTAMP_MILLIS
 echo "Created /shared/MC__FIRST_EPOCH_TIMESTAMP_MILLIS with value: $byron_startTimeMillis"
 
 echo "Current time is now: $(date +"%H:%M:%S.%3N"). Starting node..."
@@ -179,6 +180,7 @@ cat /shared/eve.utxo
 
 echo "Querying and saving the first UTXO details for new address to /shared/genesis.utxo:"
 cardano-cli latest query utxo --testnet-magic 42 --address "${new_address}" | /busybox awk 'NR>2 { print $1 "#" $2; exit }' > /shared/genesis.utxo
+cat /shared/genesis.utxo > /runtime-values/genesis.utxo
 cat /shared/genesis.utxo
 
 touch /shared/cardano.ready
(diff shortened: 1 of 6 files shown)


LGLO:
I think there is problem with `ready` files that will be left in the filesystem.

AmbientTea:
> I think there is problem with `ready` files that will be left in the filesystem.

You're right. Then probably the values need to be sourced by some `docker exec` or otherwise written somehow without persisting the files.

dev/local-environment/envrc
LGLO, line 1: Should the name be `.envrc` instead of `envrc` ?

dev/local-environment/configurations/cardano/entrypoint.sh
```diff
@@ -48,6 +48,7 @@ echo "Updated ConwayGenesisHash value in config files to: $conway_hash"
 
 byron_startTimeMillis=$(($byron_startTime * 1000))
 echo $byron_startTimeMillis > /shared/MC__FIRST_EPOCH_TIMESTAMP_MILLIS
+echo $byron_startTimeMillis > /runtime-values/MC__FIRST_EPOCH_TIMESTAMP_MILLIS
```
LGLO, line 51: We need /runtime-values in .gitignore

LGLO:
Having this, should we still
```
source dev/envs/devnet/.envrc
```
in *.envrc* in root directory?

AmbientTea:
> Having this, should we still
> 
> ```
> source dev/envs/devnet/.envrc
> ```
> 
> in _.envrc_ in root directory?

I don't have a strong opinion on this, but I often use this default to run a devnode without setting up the whole local env. So I'm open to change that but I would leave it as is.

dev/local-environment/configurations/cardano/entrypoint.sh
AmbientTea: done

dev/local-environment/envrc
AmbientTea: My idea was that this would be a file you intentionally source when you want to use the local env, so I named it without dot for visibility. I'm not married to either option.

dev/local-environment/setup.sh
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-PARTNER_CHAINS_NODE_IMAGE="ghcr.io/input-output-hk/partner-chains/partner-chains-node:v1.4.0"
+PARTNER_CHAINS_NODE_IMAGE="ghcr.io/input-output-hk/partner-chains/partner-chains-node:v1.5.0-rc8"
```
LGLO, line 3: IDK why one or another. Not this PR, but I would just remove it and require from user to pass it to setup.

dev/local-environment/envrc
```diff
@@ -0,0 +1,35 @@
+SCRIPT_DIR=$(dirname "$_")
+SHARED_DIR="$SCRIPT_DIR/data/shared"
+DOT_ENV="$SCRIPT_DIR/.env"
+SHELLEY_GENESIS_FILE="$SCRIPT_DIR/configurations/genesis/shelley/genesis.json"
+TIMESTAMP_FILE="$SCRIPT_DIR/runtime-values/MC__FIRST_EPOCH_TIMESTAMP_MILLIS"
+PC_GENESIS_FILE="$SCRIPT_DIR/runtime-values/genesis.utxo"
```
LGLO, line 6: Calling `genesis.utxo` a PC genesis file is confusing.

dev/local-environment/setup.sh
AmbientTea: Yeah, with some reasonable default maybe

PR #498, opened 2025-02-17
test: speed up block rewards tests

rsporny:
fixes:
- use cache instead of querying for the same block twice
- get validators once per pc epoch instead on every block

Refs: ETCM-9389


rsporny:
this reduces execution time by half (from 40 min to 20 min)

kpinter-iohk (approved):

PR #499, opened 2025-02-17
chore: set default local-env image to v1.5.0

rsporny:
chore: set default local-env image to v1.5.0


LGLO (approved):
Approved, but I really think we should just remove default PC node image. CI has to override it. When developing anything around node and using local-env for tests one also has to override it.

PR #506, opened 2025-02-19
change: Include candidate type and MC pubkey in committee data in the storage

AmbientTea:
# Description

IMPORTANT: This is a breaking change and must not be merged without a storage migration ready.

This PR's goal is to make previously missing committee member data available in the runtime: MC pubkey and type of candidate (permissioned, trustless)

* Makes committee member data a generic type in `pallet-session-validator-management::Config`
* Provides a `CommitteeMemberInfo` enum in the runtime to instantiate this generic type, which distinguishes between trustless and permissioned members and contains MC Public Key for the former

Follow up changes necessary:
- [x] storage migration: easiest way is to treat all members stored on runtime upgrade as permissioned 
- [ ] wizards update: at least build-spec
- [x] documentation / migration guide

# Checklist

- [x] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/node/runtime/Cargo.toml
+++ b/node/runtime/Cargo.toml
@@ -15,6 +15,7 @@ log = { workspace = true }
 parity-scale-codec = { workspace = true }
 scale-info = { workspace = true }
 derive-new = { workspace = true }
+serde = { workspace = true }
 serde_json = { workspace = true, default-features = false, features = [
 	"alloc",
 ] }
--- a/node/runtime/src/lib.rs
+++ b/node/runtime/src/lib.rs
@@ -8,7 +8,6 @@ extern crate frame_benchmarking;
 
 extern crate alloc;
 
-// A few exports that help ease life for downstream crates.
 use authority_selection_inherents::authority_selection_inputs::AuthoritySelectionInputs;
 use authority_selection_inherents::filter_invalid_candidates::{
 	validate_permissioned_candidate_data, PermissionedCandidateDataError, RegistrationDataError,
@@ -18,24 +17,14 @@ use authority_selection_inherents::select_authorities::select_authorities;
 use frame_support::genesis_builder_helper::{build_state, get_preset};
 use frame_support::weights::constants::RocksDbWeight as RuntimeDbWeight;
 use frame_support::BoundedVec;
-pub use frame_support::{
+use frame_support::{
 	construct_runtime, parameter_types,
-	traits::{
-		ConstBool, ConstU128, ConstU32, ConstU64, ConstU8, KeyOwnerProofSystem, Randomness,
-		StorageInfo,
-	},
-	weights::{
-		constants::{BlockExecutionWeight, ExtrinsicBaseWeight, WEIGHT_REF_TIME_PER_SECOND},
-		IdentityFee,
-	},
-	PalletId, StorageValue,
+	traits::{ConstBool, ConstU128, ConstU32, ConstU64, ConstU8},
+	weights::{constants::WEIGHT_REF_TIME_PER_SECOND, IdentityFee},
 };
-pub use frame_system::Call as SystemCall;
 use opaque::SessionKeys;
-pub use pallet_balances::Call as BalancesCall;
 use pallet_grandpa::AuthorityId as GrandpaId;
-pub use pallet_session_validator_management;
-pub use pallet_timestamp::Call as TimestampCall;
+use pallet_session_validator_management;
 use pallet_transaction_payment::{ConstFeeMultiplier, FungibleAdapter, Multiplier};
 use session_manager::ValidatorManagementSessionManager;
 use sidechain_domain::{
@@ -44,11 +33,7 @@ use sidechain_domain::{
 };
 use sp_api::impl_runtime_apis;
 use sp_consensus_aura::sr25519::AuthorityId as AuraId;
-use sp_core::hexdisplay::HexDisplay;
-use sp_core::{crypto::KeyTypeId, OpaqueMetadata};
-#[cfg(any(feature = "std", test))]
-pub use sp_runtime::BuildStorage;
-use sp_runtime::DispatchResult;
+use sp_core::{crypto::KeyTypeId, hexdisplay::HexDisplay, OpaqueMetadata};
 use sp_runtime::{
 	generic, impl_opaque_keys,
 	traits::{
@@ -56,9 +41,9 @@ use sp_runtime::{
 		Verify,
 	},
 	transaction_validity::{TransactionSource, TransactionValidity},
-	ApplyExtrinsicResult, MultiSignature,
+	ApplyExtrinsicResult, DispatchResult, MultiSignature, Perbill,
 };
-pub use sp_runtime::{Perbill, Permill};
+use sp_session_validator_management::CommitteeMember;
 use sp_sidechain::SidechainStatus;
 use sp_std::prelude::*;
 #[cfg(feature = "std")]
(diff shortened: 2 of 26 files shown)


node/runtime/src/lib.rs
```diff
@@ -405,6 +408,64 @@ parameter_types! {
...
+	Decode,
+	TypeInfo,
+	MaxEncodedLen,
+	Serialize,
+	Deserialize,
+	Debug,
+	PartialEq,
+	Eq,
+)]
+pub enum CommitteeMemberInfo<AuthorityId: Clone, AuthorityKeys: Clone> {
```
AmbientTea, line 423: I made this type parametric thinking about perhaps moving it out to another crate and making it reusable, but we'd be just fine with having concrete types in the fields.

node/runtime/src/lib.rs
```diff
@@ -405,6 +408,64 @@ parameter_types! {
...
+	Permissioned { id: AuthorityId, keys: AuthorityKeys },
+	Trustless { id: AuthorityId, keys: AuthorityKeys, main_chain_pub_key: StakePoolPublicKey },
+}
+
+impl<AuthorityId: Clone, AuthorityKeys: Clone> CommitteeMemberInfo<AuthorityId, AuthorityKeys> {
```
AmbientTea, line 428: Just like this info type implements an interface to be used by the validator management pallet, I imagine the production log pallet would define its own interface trait that this type here would implement as well.

node/runtime/src/lib.rs
```diff
@@ -405,6 +408,64 @@ parameter_types! {
...
+	MaxEncodedLen,
+	Serialize,
+	Deserialize,
+	Debug,
+	PartialEq,
+	Eq,
+)]
+pub enum CommitteeMemberInfo<AuthorityId: Clone, AuthorityKeys: Clone> {
+	Permissioned { id: AuthorityId, keys: AuthorityKeys },
+	Trustless { id: AuthorityId, keys: AuthorityKeys, main_chain_pub_key: StakePoolPublicKey },
```
LGLO, line 425: stake_pool_public_key, please

node/runtime/src/lib.rs
LGLO: If it is here, then as concrete as possible, please.

toolkit/pallets/session-validator-management/src/lib.rs
```diff
@@ -57,10 +72,16 @@ pub mod pallet {
 			+ From<u64>
 			+ Into<u64>;
 
+		type CommitteeMemberInfo: Parameter
+			+ Member
+			+ MaybeSerializeDeserialize
+			+ MaxEncodedLen
+			+ CommitteeMemberInfo<Self::AuthorityId, Self::AuthorityKeys>;
```
LGLO, line 79: We could demand `Into<Self::AuthorityId>` and `Into<Self::AuthorityKeys>` without adding new trait that basically expresses these two capabilities

AmbientTea: I considered this but I'm going to push back here. `Into` only tells us there's some cast between types. Having a trait shows that we're establishing a contract here and that types that implement it conform to the assumed logic of the pallet (a wittgensteinian language game if you will. yes, I checked wikipedia for that). In other words, I think its better to see `authority_keys()` and know what they mean immediately, than `into()` and hunt for the type.

AmbientTea: (I think of `Into` as meaning "this thing is already what I need here, just need to call this for the type checker to agree with me")

node/runtime/src/lib.rs
AmbientTea: Yeah, I made it concrete

toolkit/pallets/session-validator-management/src/migrations/v1.rs
```diff
@@ -0,0 +1,193 @@
...
+
+	#[storage_alias]
+	pub type CurrentCommittee<T: crate::pallet::Config> = StorageValue<
+		crate::Pallet<T>,
+		LegacyCommitteeInfo<
+			<T as crate::pallet::Config>::ScEpochNumber,
+			<T as crate::pallet::Config>::AuthorityId,
+			<T as crate::pallet::Config>::AuthorityKeys,
+			<T as crate::pallet::Config>::MaxValidators,
+		>,
```
AmbientTea, line 53: I couldn't compile it without this level of specificity.

toolkit/pallets/session-validator-management/README.md
```diff
@@ -0,0 +1,9 @@
+# Session Validator Management Pallet
+
+This pallet provides a way to rotate session validators based on arbitrary
+inputs and selection algorithm.
+
+## Migrations
+
+This pallet's storage has changed compared to its legacy version. See
```
LGLO, line 8: its legacy => 1.5.0

toolkit/pallets/session-validator-management/src/lib.rs
LGLO: :+1:

node/runtime/src/lib.rs
```diff
@@ -405,6 +408,66 @@ parameter_types! {
...
+	Decode,
+	TypeInfo,
+	MaxEncodedLen,
+	Serialize,
+	Deserialize,
+	Debug,
+	PartialEq,
+	Eq,
+)]
+pub enum CommitteeMemberInfo {
```
LGLO, line 423: Cosmetic: why "Info"? Isn't it just `CommitteeMember` (`Validator`)?

node/runtime/src/lib.rs
```diff
@@ -405,6 +408,66 @@ parameter_types! {
...
+	Permissioned { id: CrossChainPublic, keys: SessionKeys },
+	Registered { id: CrossChainPublic, keys: SessionKeys, stake_pool_pub_key: StakePoolPublicKey },
```
LGLO, line 425: `Permissioned` and `Registered` here will have to be copied to other runtimes. I think they belong to some other place. `authority-selection-inherents` already depend on `sp-session-validator-management`

node/runtime/Cargo.toml
```diff
@@ -15,6 +15,7 @@ log = { workspace = true }
 parity-scale-codec = { workspace = true }
 scale-info = { workspace = true }
 derive-new = { workspace = true }
+serde = { workspace = true }
```
LGLO, line 18: Is it required? Don't we use `sp_runtime::Serialize` et.al?

node/runtime/src/lib.rs
AmbientTea: Changed to `CommitteeMember`. `Validator` sounds too much technical imo (as in the actual node, not the identity behind running it).

node/runtime/Cargo.toml
AmbientTea: The derivations that these macros use still depend on the `serde` crate .

node/runtime/src/lib.rs
AmbientTea: Moved to the `authority-selection-inherents` crate which procides this data in the first place.

node/runtime/src/lib.rs
```diff
@@ -17,6 +16,8 @@ use authority_selection_inherents::filter_invalid_candidates::{
 use authority_selection_inherents::select_authorities::select_authorities;
 use frame_support::genesis_builder_helper::{build_state, get_preset};
 use frame_support::weights::constants::RocksDbWeight as RuntimeDbWeight;
+use sp_session_validator_management::CommitteeMember;
+// A few exports that help ease life for downstream crates.
```
LGLO, line 20: Cosmetic: this comment perhaps belongs to `pub use`, but what are downstream crates?

toolkit/pallets/session-validator-management/src/lib.rs
```diff
@@ -57,10 +63,16 @@ pub mod pallet {
 			+ From<u64>
 			+ Into<u64>;
 
+		type CommitteeMember: Parameter
+			+ Member
+			+ MaybeSerializeDeserialize
+			+ MaxEncodedLen
+			+ CommitteeMember<Self::AuthorityId, Self::AuthorityKeys>;
```
LGLO, line 70: Cosmetic: using qualified name would easier to comprehend for me.

node/runtime/src/lib.rs
AmbientTea: Turns out none of them was needed. x)

LGLO (approved):

PR #509, opened 2025-02-20
Update python to 3.12

gilligan:
# Description

Provide python 3.12 in the nix shell (previously we were on 3.11)

# Checklist

- [x] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/dev/nix/shell.nix
+++ b/dev/nix/shell.nix
@@ -40,9 +40,9 @@
 
               # tools for e2e testing
               docker-compose
-              python311
-              python311Packages.pip
-              python311Packages.virtualenv
+              python312
+              python312Packages.pip
+              python312Packages.virtualenv
               sops
 
               # local development tools


dev/nix/shell.nix
```diff
@@ -1,86 +1,87 @@
...
 
-          # tools for e2e testing
-          docker-compose
-          python311
-          python311Packages.pip
-          python311Packages.virtualenv
-          sops
+            # tools for e2e testing
+            docker-compose
+            python312
```
gilligan, line 46: Those few lines here are the relevant ones. Bit surprised by the changes all over the file - this must be due to an update of nixpkgs-fmt? 🤔

dev/nix/shell.nix
```diff
@@ -1,86 +1,87 @@
...
+            openssl
+            gnumake
 
-          # tools for e2e testing
-          docker-compose
-          python311
-          python311Packages.pip
-          python311Packages.virtualenv
-          sops
+            # tools for e2e testing
```
LGLO, line 44: This comment has confusion factor 100.

`docker-compose` is not needed by `e2e-tests` per se. It is at most needed for running local environment.

`python312` from devshell is not suitable for running `e2e-tests`, because numpy won't work in it.

dev/nix/shell.nix
```diff
@@ -1,86 +1,87 @@
...
-          # infra packages
-          earthly
-          awscli2
-          kubectl
-          kubernetes-helm
+            # infra packages
+            earthly
+            awscli2
+            kubectl
+            kubernetes-helm
```
LGLO, line 63: I'm wondering if we have a single Helm file in this repo.

LGLO (approved):

dev/nix/shell.nix
gilligan: I agree. The diff makes it look like I added the comment - i didn't

gilligan:
At this point I'm not entirely sure if I shouldn't just remove python from the nix shell completely seeing how it's currently causing problems with the use of numpy anyway.
I'm leaning towards removing it and just assuming that python needs to be installed individually.
@rsporny you probably don't mind either way?

rsporny (approved):

dev/nix/shell.nix
rsporny: `numpy` issue is not a global issue, it's caused by file system is missing something. I can try to replicate that @LGLO if you give me your specs. AFAIK 3.12 works fine, both in devshell and native installations, both on mac and linux.

rsporny:
@gilligan I don't mind, I checked out this branch and it works fine. I'll switch to using devshell's python if we merge it.

PR #513, opened 2025-02-24
change: Feed block production log from session validator management pallet

AmbientTea:
# Description

Reworks the block production log pallet to source the block author using runtime API and implements that API using the session validator management pallet.

# Checklist

- [ ] Self-reviewed the diff

# Note on CI
for you.


node/runtime/src/lib.rs
```diff
@@ -455,8 +458,45 @@ impl pallet_block_rewards::Config for Runtime {
...
+				f.write_str("Incentivized(")?;
+				f.write_str(&hex::encode(pc_key))?;
+				f.write_str(", ")?;
+				f.write_str(&hex::encode(&mc_key.0))?;
+				f.write_str(")")?;
+			},
+		};
+		Ok(())
+	}
+}
```
Klapeyron, line 485: why do we need to impl this macro manually?

node/runtime/src/lib.rs
```diff
@@ -871,6 +908,15 @@ impl_runtime_apis! {
...
 	}
+
+	impl sp_block_production_log::BlockProductionLogApi<Block, CommitteeMember<CrossChainPublic, SessionKeys>>  for Runtime {
+		fn get_current_author() -> CommitteeMember<CrossChainPublic, SessionKeys> {
+			let mut committee = SessionCommitteeManagement::get_current_committee().1;
+			let slot = pallet_aura::CurrentSlot::<Runtime>::get();
+			// this implementation is compatible with round-robin employed by Aura
+			committee.remove(*slot as usize % committee.len())
+		}
+	}
```
Klapeyron, line 919: I am not sure regarding the business logic, but it looks like we are fetching committee which is a vector which involves the copy of the underlying structure, but later we only take one element of if which we return. 
Maybe it will be good idea to implement some function for `SessionCommitteeManagement` that will make this single lookup for us

Klapeyron: or we can use `SessionCommitteeManagement::get_current_authority(slot)`, please check if it is doing the same

Klapeyron: also once we will be there, we can adjust the name of the function, maybe `current_authority` will be more idiomatic
https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

AmbientTea: Yeah, I don't want to change that. I don't see a way to get the committee size and get nth member modulo that size with one storage read, unless I move round-robin logic into the pallet and I don't think it belongs there.

Klapeyron: Yep, but we do not need to change the logic too much. We can use something like this with maybe some comment that we make a boundary check with modulo operation:
```suggestion
		fn get_current_author() -> CommitteeMember<CrossChainPublic, SessionKeys> {
			let committee = SessionCommitteeManagement::get_current_committee().1;
			let slot = pallet_aura::CurrentSlot::<Runtime>::get();
			// this implementation is compatible with round-robin employed by Aura
			committee.get_unchecked(*slot as usize % committee.len())
		}
	}
```
or for example use [swap_remove](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.swap_remove), which is O(1). It will reorder elements, but we drop this vec at the end of the function, so there will be no impact, but it may be less readable

AmbientTea: I'll remove the `remove` tho

kpinter-iohk (approved):

PR #515, opened 2025-02-24
Etcm 9319 payouts pallet

AmbientTea:
# Description

draft

to the issue.

# Checklist

- [ ] Self-reviewed the diff

# Note on CI
for you.


toolkit/pallets/block-rewards-payouts/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+
+	impl<T: Config> Pallet<T> {
+		fn get_rewards_inherent_data(
+			data: &InherentData,
+		) -> Option<BlockProductionPayoutInfo<T::BlockProducerId, T::BeneficiaryId>> {
+			data.get_data(&Self::INHERENT_IDENTIFIER)
+				.map_err(|err| {
+					format!("Reward payout inherent data is present but invalid: {err:?}")
+				})
+				.unwrap()
```
gilligan, line 75: maybe match on `data.get_data(&Self::INHERENT_IDENTIFIER)` here? You are handling the case of data being present that isn't valid but not that no data is present, right?

toolkit/pallets/block-rewards-payouts/src/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+			sp_block_rewards_payouts::INHERENT_IDENTIFIER;
+
+		fn create_inherent(data: &InherentData) -> Option<Self::Call> {
+			let up_to_slot = Self::get_rewards_inherent_data(data)?.up_to_slot;
+
+			Some(Call::note_payout { up_to_slot })
+		}
+
+		fn check_inherent(call: &Self::Call, data: &InherentData) -> Result<(), Self::Error> {
+			let Some(expected_inherent_data) = Self::get_rewards_inherent_data(data) else {
```
gilligan, line 39: This code assumes that inherent data is always available. You might want to check and conditionally return an Err?

toolkit/primitives/block-rewards-payouts/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+				let distribution = stake_distribution_data_source
+					.get_stake_pool_delegation_distribution_for_pools(mc_epoch, &pools)
+					.await
+					.map_err(InherentDataCreationError::DataSourceError)?;
+				for (producer, block_count) in producers.into_iter() {
+					let (beneficiary_total_share, beneficiaries) = match producer.clone().into() {
+						None => (0, vec![]),
+						Some(cardano_producer) => {
+							let PoolDelegation { total_stake, delegators } =
+								distribution.0.get(&cardano_producer).expect("");
```
gilligan, line 128: This assumes that there's always a value - maybe we want to be more careful to get a useful error instead of a panic..

```Rust
let PoolDelegation { total_stake, delegators } =
    distribution.0.get(&cardano_producer)
    .ok_or_else(|| InherentDataCreationError::DataSourceError("Missing delegation data".into()))?;
```

toolkit/primitives/block-rewards-payouts/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+
+		for (slot, producer) in slot_producers {
+			let mc_epoch = mc_epoch_config
+				.timestamp_to_mainchain_epoch(Timestamp::from_unix_millis(
+					slot.timestamp(slot_duration)
+						.expect("Timestamp for past slots can not overflow")
+						.as_millis(),
+				))
+				.expect("Mainchain epoch for past slots exists");
+			let mc_epoch = McEpochNumber(mc_epoch.0 - 2);
```
gilligan, line 177: Could you perhaps introduce a tiny function that provides/adjusts the epoch number along with a short comment that outlines _why_ we go 2 main chain epochs in the past. 

I'm mostly OK with being light on code comments but this seems like an opportunity where providing some context makes sense I believe.

toolkit/primitives/block-rewards-payouts/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+			Ok(())
+		}
+
+		async fn try_handle_error(
+			&self,
+			identifier: &InherentIdentifier,
+			mut error: &[u8],
+		) -> Option<Result<(), sp_inherents::Error>> {
+			if *identifier == INHERENT_IDENTIFIER {
+				let error = InherentError::decode(&mut error).ok()?;
```
gilligan, line 208: If the decoding fails this fails silently. Maybe you want to match this on `Ok` and log an error if there was a problem?

toolkit/primitives/block-rewards-payouts/src/lib.rs
```diff
@@ -0,0 +1,215 @@
...
+						beneficiary_total_share,
+						beneficiaries,
+					};
+					block_production_payout_info.production_rundowns.push(rundown);
+				}
+			}
+			block_production_payout_info.production_rundowns.sort();
+
+			for rundown in &mut block_production_payout_info.production_rundowns {
+				rundown.beneficiaries.sort()
```
gilligan, line 152: I believe in order to ensure consistency it might be good to be explicit about the ordering?
```Rust
for rundown in &mut block_production_payout_info.production_rundowns {
    rundown.beneficiaries.sort()
}
```
Tell me if i'm wrong but otherwise ordering might not be reproducible if the type changes?

gilligan (changes requested):
I have some suggestion about improved error handling but nothing of major concern

toolkit/pallets/block-rewards-payouts/src/lib.rs
AmbientTea: that's intended. It can be not present and that's okay

toolkit/primitives/block-rewards-payouts/src/lib.rs
AmbientTea: Changed

toolkit/primitives/block-rewards-payouts/src/lib.rs
AmbientTea: If the type changes the ordering will need to change anyway. Rust's derived ordering is field-by-field, so we're good using it here. Besides, the payouts should be the same for any ordering (this data doesn't go into transactions), so I'm sorting it more for consistency that out of necessity.

toolkit/primitives/block-rewards-payouts/src/lib.rs
AmbientTea: Added logging

node/runtime/src/lib.rs
```diff
@@ -491,6 +497,14 @@ impl From<CommitteeMember<CrossChainPublic, SessionKeys>> for BlockAuthor {
 		}
 	}
 }
+impl From<BlockAuthor> for Option<MainchainKeyHash> {
+	fn from(value: BlockAuthor) -> Self {
+		match value {
+			BlockAuthor::Incentivized(_, key) => Some(key.hash()),
+			_ => None,
```
Klapeyron, line 504: if there is only one more element, let's make it without wildcard, in case someone will add new element in the future, we will have an compilation error
```suggestion
			BlockAuthor::ProBono(_) => None,
```

toolkit/mainchain-follower/db-sync-follower/src/candidates/mod.rs
```diff
@@ -287,7 +288,7 @@ impl CandidatesDataSourceImpl {
 			))
 			.into())
 		} else {
-			Ok(McEpochNumber(epoch_of_data_usage.0 - 2))
+			Ok(McEpochNumber(epoch_of_data_usage.0 - Self::DATA_MC_EPOCH_OFFSET))
```
Klapeyron, line 291: we should also use constant in line 284, but we can also combine this sub with the previous check, consider something like this:
```rust
	fn get_epoch_of_data_storage(
		&self,
		epoch_of_data_usage: McEpochNumber,
	) -> Result<McEpochNumber, Box<dyn std::error::Error + Send + Sync>> {
		// Please add some note what it is
		const DATA_MC_EPOCH_OFFSET: u32 = 2;

		let epoch_number = epoch_of_data_usage.0.checked_sub(DATA_MC_EPOCH_OFFSET).ok_or_else(|| BadRequest(format!(
			"Minimum supported epoch of data usage is 2, but {} was provided",
			epoch_of_data_usage.0
		)))?;

		Ok(McEpochNumber(epoch_number))
	}
```

toolkit/pallets/block-participation/src/tests.rs
```diff
@@ -0,0 +1,95 @@
...
+	assert_eq!(inherent, crate::Call::<Test>::note_processing { up_to_slot: Slot::from(24) })
+}
+
+#[test]
+fn check_fails_if_data_is_missing() {
+	let inherent_data = InherentData::new();
+	let inherent = crate::Call::<Test>::note_processing { up_to_slot: Slot::from(25) };
+
+	let result = Payouts::check_inherent(&inherent, &inherent_data);
+	assert_eq!(result, Err(sp_block_participation::InherentError::UnexpectedInherent))
```
Klapeyron, line 44: there are also `assert_ok` and `assert_err` macros, consider them:
```suggestion
	assert_err!(
		Payouts::check_inherent(&inherent, &inherent_data),
		sp_block_participation::InherentError::UnexpectedInherent,
	);
```

toolkit/primitives/block-participation/src/lib.rs
```diff
@@ -0,0 +1,313 @@
...
+			let pools: Vec<_> = producers.flat_map(Into::<Option<_>>::into).collect();
+			data_source
+				.get_stake_pool_delegation_distribution_for_pools(mc_epoch, &pools)
+				.await
+				.map_err(InherentDataCreationError::DataSourceError)
+		}
+
+		// The number of main chain epochs back a Partner Chain queries for candidate stake delegation.
+		// This offset is necessary to fetch the same data that was used to select committee members.
+		const DATA_MC_EPOCH_OFFSET: u32 = 2;
```
Klapeyron, line 241: looks like duplication, is there an option to unify it?

toolkit/primitives/block-participation/src/lib.rs
```diff
@@ -0,0 +1,313 @@
...
+				inherent_data.put_data(INHERENT_IDENTIFIER, &block_production_data.up_to_slot)?;
+			}
+			Ok(())
+		}
+
+		async fn try_handle_error(
+			&self,
+			identifier: &InherentIdentifier,
+			mut error: &[u8],
+		) -> Option<Result<(), sp_inherents::Error>> {
```
Klapeyron, line 298: it will be not easier to make it `Result<Option<()>, sp_inherents::Error>>`?

toolkit/mainchain-follower/db-sync-follower/src/candidates/mod.rs
gilligan: @Klapeyron yeah that looks really nice and much like what i had in mind when i previously commented on the epoch offset

toolkit/pallets/block-participation/src/lib.rs
```diff
@@ -0,0 +1,113 @@
...
+				return Err(Self::Error::UnexpectedInherent);
+			};
+
+			let Self::Call::note_processing { up_to_slot } = call else {
+				unreachable!("There should be no other extrinsic in the pallet")
+			};
+
+			if *up_to_slot != expected_inherent_data {
+				return Err(Self::Error::IncorrectSlotBoundary);
+			}
```
Klapeyron, line 59:
```suggestion
			ensure!(*up_to_slot == expected_inherent_data, Self::Error::IncorrectSlotBoundary);
```

gilligan (approved):
@Klapeyron 's suggestions look worth following up on but otherwise I'm good with everything

toolkit/pallets/block-participation/src/lib.rs
```diff
@@ -0,0 +1,113 @@
...
+		fn is_inherent(call: &Self::Call) -> bool {
+			matches!(call, Call::note_processing { .. })
+		}
+
+		fn is_inherent_required(data: &InherentData) -> Result<Option<Self::Error>, Self::Error> {
+			if Self::decode_inherent_data(data).is_some() {
+				Ok(Some(Self::Error::InherentRequired))
+			} else {
+				Ok(None)
+			}
```
Klapeyron, line 73: consider `map`:
```suggestion
			Ok(Self::decode_inherent_data(data).map(|_slot| Self::Error::InherentRequired))
```

toolkit/pallets/block-participation/src/lib.rs
```diff
@@ -0,0 +1,113 @@
...
+				})
+				.unwrap()
+		}
+	}
+
+	#[pallet::call]
+	impl<T: Config> Pallet<T> {
+		/// Registers the fact that block participation data has been released and removes the handled data from block production log.
+		#[pallet::call_index(0)]
+		#[pallet::weight((0, DispatchClass::Mandatory))]
```
Klapeyron, line 91: don't forget about benchmark

toolkit/pallets/block-participation/src/tests.rs
```diff
@@ -0,0 +1,95 @@
...
+	let result = Payouts::is_inherent_required(&inherent_data).unwrap();
+	assert_eq!(result, Some(sp_block_participation::InherentError::InherentRequired))
+}
+
+#[test]
+fn inherent_is_not_required_when_data_is_not_present() {
+	let inherent_data = InherentData::new();
+
+	let result = Payouts::is_inherent_required(&inherent_data).unwrap();
+	assert_eq!(result, None)
```
Klapeyron, line 81:
```suggestion
	assert_ok!(Payouts::is_inherent_required(&inherent_data), None);
```

toolkit/primitives/block-participation/src/lib.rs
```diff
@@ -0,0 +1,313 @@
...
+			mc_epoch_config: &MainchainEpochConfig,
+		) -> McEpochNumber {
+			let mc_epoch = mc_epoch_config
+				.timestamp_to_mainchain_epoch(Timestamp::from_unix_millis(
+					slot.timestamp(slot_duration)
+						.expect("Timestamp for past slots can not overflow")
+						.as_millis(),
+				))
+				.expect("Mainchain epoch for past slots exists");
+			McEpochNumber(mc_epoch.0 - Self::DATA_MC_EPOCH_OFFSET)
```
Klapeyron, line 255: should we check here if sub is safe?

node/runtime/src/lib.rs
AmbientTea: yeah

toolkit/pallets/block-participation/src/tests.rs
AmbientTea: I honestly think `pretty_assertions::assert_eq` does a better job than those specialized assertions

toolkit/primitives/block-participation/src/lib.rs
AmbientTea: that signature is dictated by the substrate trait

toolkit/pallets/block-participation/src/lib.rs
```diff
@@ -0,0 +1,113 @@
...
+	}
+
+	impl<T: Config> Pallet<T> {
+		fn decode_inherent_data(data: &InherentData) -> Option<Slot> {
+			data.get_data(&Self::INHERENT_IDENTIFIER)
+				.map_err(|err| {
+					format!("Block participation inherent data is present but invalid: {err:?}")
+				})
+				.unwrap()
+		}
```
Klapeyron, line 84: here would be great to add new error type for this pallet with `#[pallet::error]`, it will allow us to change return type to `Result<Option<Slot>, Error<T>>` and remove unwrap
```suggestion
	impl<T: Config> Pallet<T> {
		fn decode_inherent_data(data: &InherentData) -> Result<Option<Slot>, Error<T>> {
			data.get_data(&Self::INHERENT_IDENTIFIER)
				.map_err(|err| Error::<T>::InvalidParticipationInherentData)
		}
	}
```

toolkit/pallets/block-participation/src/lib.rs
AmbientTea: I did, but I prefer to be more explicit

toolkit/primitives/block-participation/src/lib.rs
AmbientTea: yes, but I don't want to do it in this PR. I have it noted down

toolkit/mainchain-follower/db-sync-follower/src/candidates/mod.rs
AmbientTea: yeah, I'll do that in a separate PR tho

toolkit/primitives/block-participation/src/lib.rs
AmbientTea: in theory we shouldn't have to, because that subtraction has already been done when selecting the committee, but I we can be defensive here

toolkit/primitives/block-participation/src/lib.rs
```diff
@@ -0,0 +1,313 @@
...
+		fn production_breakdown_for(
+			mc_epoch: McEpochNumber,
+			block_producer: BlockProducer,
+			block_count: u32,
+			distribution: &StakeDistribution,
+		) -> Result<
+			BlockProducerParticipationData<BlockProducer, Beneficiary>,
+			InherentDataCreationError<BlockProducer>,
+		> {
+			let (beneficiary_total_share, beneficiaries) = match block_producer.clone().into() {
```
Klapeyron, line 201: hmm, I am not getting the moment, when we transfer `BlockProducer` into `Option`, why we want to do so?

toolkit/primitives/block-participation/src/lib.rs
```diff
@@ -0,0 +1,313 @@
...
+			mc_epoch_config: &MainchainEpochConfig,
+			slot_duration: SlotDuration,
+		) -> HashMap<McEpochNumber, HashMap<BlockProducer, u32>> {
+			let mut epoch_producers: HashMap<McEpochNumber, HashMap<BlockProducer, u32>> =
+				HashMap::new();
+
+			for (slot, producer) in slot_producers {
+				let mc_epoch = Self::data_mc_epoch_for_slot(slot, slot_duration, mc_epoch_config);
+
+				*epoch_producers.entry(mc_epoch).or_default().entry(producer).or_default() += 1;
```
Klapeyron, line 269: it is quite hard to decompose what this line does, can we refactor it a bit?

toolkit/mainchain-follower/db-sync-follower/src/candidates/mod.rs
Klapeyron: I do not like the concept of doing that in separate PR's, usually this stuff is gone

toolkit/primitives/block-participation/src/lib.rs
AmbientTea: we do it because on every block producer is a cardano SPO. we ignore the "pro-bono" (ie. permissioned) producers and only fetch stake for "incentivized" (ie. registered = SPO) producers

toolkit/primitives/block-participation/src/lib.rs
Klapeyron: how much effort it will be to do it in this PR? Can we give a try?

toolkit/pallets/block-participation/src/lib.rs
AmbientTea: added. that still requires unwrapping when proposing a block, but we can't do anything about it

toolkit/primitives/block-participation/src/lib.rs
Klapeyron: hmm, but there should be in this function some match on this enum that defines which variant we have, but I do not see it here

node/runtime/src/lib.rs
```diff
@@ -491,6 +497,14 @@ impl From<CommitteeMember<CrossChainPublic, SessionKeys>> for BlockAuthor {
...
 	}
 }
+impl From<BlockAuthor> for Option<MainchainKeyHash> {
+	fn from(value: BlockAuthor) -> Self {
+		match value {
+			BlockAuthor::Incentivized(_, key) => Some(key.hash()),
+			BlockAuthor::ProBono(_) => None,
+		}
+	}
+}
```
Klapeyron, line 507: ah ok, this conversion is here. We should not do impl of From for Option types here, I would rather propose to write some method for that converts one into another ore use TryFrom instead

Cargo.toml
```diff
@@ -241,3 +243,5 @@ partner-chains-cli = { path = "toolkit/partner-chains-cli", default-features = f
 partner-chains-plutus-data = { path = "toolkit/primitives/plutus-data", default-features = false }
 partner-chains-smart-contracts-commands = { path = "toolkit/cli/smart-contracts-commands", default-features = false }
 pallet-address-associations = { path = "toolkit/pallets/address-associations", default-features = false }
+pallet-block-participation = { path =	"toolkit/pallets/block-participation", default-features = false }
+sp-block-participation = { path =	"toolkit/primitives/block-participation", default-features = false }
```
Klapeyron, line 247: looks like a formatting issue

node/runtime/src/lib.rs
AmbientTea: `TryFrom`  introduces the notion that this cast can fail, we don't want that. I added a trait, good idea

toolkit/primitives/block-participation/src/lib.rs
AmbientTea: No, it's almost 1.5k LOC already and I have other follow-up changes I want to do anyway

toolkit/primitives/block-participation/src/lib.rs
```diff
@@ -0,0 +1,337 @@
...
+		fn production_breakdown_for(
+			mc_epoch: McEpochNumber,
+			block_producer: BlockProducer,
+			block_count: u32,
+			distribution: &StakeDistribution,
+		) -> Result<
+			BlockProducerParticipationData<BlockProducer, Beneficiary>,
+			InherentDataCreationError<BlockProducer>,
+		> {
+			let (beneficiary_total_share, beneficiaries) = match block_producer.as_cardano_spo() {
```
Klapeyron, line 213: I like it, definitely way to go. But `as_*` can be a bit misleading, due usually it represents cheap conversion that never fails (here we return Option). What if we will for example make a match on the `block_producer` enum and later convert only content of `Incentivized` case?

node/node/src/inherent_data.rs
```diff
@@ -102,8 +110,11 @@ where
 			mc_hash.mc_epoch(),
 		)
 		.await?;
+		#[cfg(test)]
```
AmbientTea, line 113: Since both production log IDP and block participation IDP depend on a runtime API being present, we need to create them using a constructor that checks the availability of the API. However, the mock runtime used in our tests does not implement the runtime versioning API (this is a limitation of substrate test utils), so we need to use a different constructor for tests. I don't like that really

Klapeyron: which target should be built to receive an error?

AmbientTea: `cargo test`

node/node/src/inherent_data.rs
```diff
@@ -8,6 +8,7 @@ use jsonrpsee::core::async_trait;
 use sc_consensus_aura::{find_pre_digest, SlotDuration};
 use sc_service::Arc;
 use sidechain_domain::mainchain_epoch::MainchainEpochConfig;
+use sidechain_domain::DelegatorKey;
```
LGLO, line 11: Cosmetic: grouping imports. I really need to bring unstable toolchain for this reason. We have it all over the place.

toolkit/primitives/block-production-log/src/lib.rs
```diff
@@ -30,11 +30,29 @@ impl IsFatalError for InherentError {
...
+	) -> Result<Self, Box<dyn Error + Send + Sync>>
+	where
+		Member: Decode,
+		Block: BlockT,
+		C: ProvideRuntimeApi<Block>,
+		C::Api: BlockProductionLogApi<Block, Member>,
+		Author: From<Member>,
+	{
+		let version = client.runtime_api().version(parent_hash)?;
+		if version.has_api_with(&<dyn BlockProductionLogApi<Block, Member>>::ID, |_| true) {
```
Klapeyron, line 50: there is shorter version with `has_api`, when we do not need to provide `with` predicate

toolkit/primitives/block-participation/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+						client,
+						data_source,
+						parent_hash,
+						current_slot,
+						mc_epoch_config,
+						slot_duration,
+					)
+					.await
+				},
+			}
```
Klapeyron, line 185:
```suggestion
			let has_api = client.runtime_api().has_api::<dyn BlockParticipationApi<Block, BlockProducer>>(parent_hash)?;

			if has_api {
				Self::new_cardano_stake_based(
					client,
					data_source,
					parent_hash,
					current_slot,
					mc_epoch_config,
					slot_duration,
				)
				.await
			}
			else {
				Ok(Self::Inert)
			}
```

toolkit/primitives/block-production-log/src/lib.rs
```diff
@@ -30,11 +30,29 @@ impl IsFatalError for InherentError {
...
+			Self::new(client, parent_hash)
+		} else {
+			Ok(Self { author: None })
+		}
```
Klapeyron, line 54:
```suggestion
		let has_api = client.runtime_api().has_api::<dyn BlockProductionLogApi<Block, Member>>(parent_hash)?;
		if has_api {
			Self::new(client, parent_hash)
		} else {
			Ok(Self { author: None })
		}
```

node/node/src/inherent_data.rs
Klapeyron: Ok, it looks like `mock_impl_runtime_apis!` implements automatically:

> Besides implementing the given traits, the [Core](https://paritytech.github.io/polkadot-sdk/master/sp_api/trait.Core.html) and [ApiExt](https://paritytech.github.io/polkadot-sdk/master/sp_api/trait.ApiExt.html) are implemented automatically.

so our call on `runtime_api().version()` is throwing: `Core::version not implemented for runtime api mocks`.
But it looks like `ApiExt` is implemented in some nicer way, so let's try to use this API instead:

```rust
let has_api = client.runtime_api().has_api::<dyn NativeTokenManagementApi<Block>>(parent_hash)?;
```

toolkit/primitives/block-participation/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+				api.blocks_produced_up_to_slot(parent_hash, up_to_slot)?;
+			let target_inherent_id = api.target_inherent_id(parent_hash)?;
+
+			let block_counts_by_epoch_and_producer = Self::count_blocks_by_epoch_and_producer(
+				blocks_produced_up_to_slot,
+				mc_epoch_config,
+				slot_duration,
+			)?;
+
+			let mut production_rundowns = vec![];
```
LGLO, line 214: Just about naming: I had to use dictionary to understand that this less formal of "summary". Also, breakdown for me is when something breaks down.

toolkit/primitives/block-participation/src/lib.rs
```diff
@@ -0,0 +1,370 @@
...
+pub const INHERENT_IDENTIFIER: InherentIdentifier = *b"blokpart";
+
+#[derive(Clone, Debug, PartialEq, Eq, Decode, Encode, TypeInfo, PartialOrd, Ord)]
+pub struct DelegatorBlockParticipationData<DelegatorId> {
+	id: DelegatorId,
+	share: u64,
+}
+
+#[derive(Clone, Debug, PartialEq, Eq, Decode, Encode, TypeInfo, PartialOrd, Ord)]
+pub struct BlockProducerParticipationData<BlockProducerId, DelegatorId> {
```
LGLO, line 24: Maybe I'm not on form, but I struggle to understand the code and additional type parameters don't help me at all. We provide Cardano data, where `BlockProducerId` is a **Stake Pool Key Hash** and `DelegatorId` is **Staking Key Hash**.

node/runtime/src/lib.rs
```diff
@@ -519,6 +535,21 @@ impl pallet_address_associations::Config for Runtime {
 	}
 }
 
+impl pallet_block_participation::Config for Runtime {
+	type WeightInfo = pallet_block_participation::weights::SubstrateWeight<Runtime>;
+	type DelegatorId = DelegatorKey;
+
+	fn should_release_data(slot: sidechain_slots::Slot) -> Option<sidechain_slots::Slot> {
+		if *slot % 3 == 0 {
```
LGLO, line 543: A follow up should be to make it configurable. This value won't help us spot any performance problems. OTOH we have only 8 different block producers in our networks, so making it much greater won't help anyway.

node/node/src/inherent_data.rs
AmbientTea: This works, thanks

toolkit/primitives/block-participation/src/lib.rs
AmbientTea: renamed to `production_summaries`

node/runtime/src/lib.rs
AmbientTea: I'll note that down for the future. It sure would be nice to have a better testability in general

toolkit/primitives/block-participation/src/lib.rs
```diff
@@ -0,0 +1,366 @@
...
+	where
+		DelegatorId: Encode + Send + Sync,
+		BlockProducerId: Encode + Send + Sync,
+	{
+		async fn provide_inherent_data(
+			&self,
+			inherent_data: &mut InherentData,
+		) -> Result<(), sp_inherents::Error> {
+			if let Self::Active { target_inherent_id, block_production_data } = &self {
+				inherent_data.put_data(target_inherent_id.clone(), block_production_data)?;
```
Klapeyron, line 341:
```
warning: using `clone` on type `[u8; 8]` which implements the `Copy` trait
   --> toolkit/primitives/block-participation/src/lib.rs:345:28
    |
345 | ...nt_data.put_data(target_inherent_id.clone(), block_production_data)?;
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^ help: try dereferencing it: `*target_inherent_id`

```

toolkit/primitives/domain/src/lib.rs
```diff
@@ -860,15 +862,36 @@ mod tests {
...
+				hex::encode(hash_raw),
+				hex::encode(script_hash)
+			),
+			Self::StakeKeyHash(hash) => alloc::format!("StakeKeyHash({})", hex::encode(hash)),
+		};
+
+		f.write_str(&s)
+	}
+}
+
```
Klapeyron, line 885: why do we use custom `Debug`?

node/node/src/inherent_data.rs
```diff
@@ -117,6 +125,27 @@ where
...
+		#[cfg(not(test))]
+		let payouts = BlockParticipationInherentDataProvider::new_cardano_stake_based_if_pallet_present(
+			client.as_ref(),
+			stake_distribution_data_source.as_ref(),
+			parent_hash,
+			*slot,
+			mc_epoch_config,
+			config.sc_slot_config.slot_duration,
+		)
+		.await?;
```
Klapeyron, line 147: I am not sure why we are losing comments history: https://github.com/input-output-hk/partner-chains/pull/515#discussion_r1976436874

Please unify this calls and remove macro

toolkit/primitives/block-participation/src/lib.rs
AmbientTea: block producer is not known to be an enum here

toolkit/primitives/domain/src/lib.rs
AmbientTea: byte arrays

Klapeyron: maybe `{:X}` will be helpfull or one of the other [formatting options](https://doc.rust-lang.org/std/fmt/#formatting-traits)

PR #519, opened 2025-02-25
change: Append block producer to log on block finalization

AmbientTea:
# Description

Adjustments for the block producer log pallet in anticipation of block reward payouts.

- Adds `peek_prefix` and `drop_prefix` operations to the block production log pallet
- Saves the block producer in a temporary storage and appends it to the block production log only in the `on_finalize` hook. This is to make the the prefix operations independent of pallet order in the runtime.

# Checklist

- [x] Self-reviewed the diff

# Note on CI
for you.


toolkit/pallets/block-production-log/src/test.rs
```diff
@@ -9,12 +13,20 @@ fn make_id(i: u64) -> [u8; 32] {
...
+
 #[test]
 fn first_append_should_succeed() {
 	new_test_ext().execute_with(|| {
 		let call = Call::<Test>::append { block_producer_id: make_id(1) };
 		assert_ok!(call.dispatch_bypass_filter(RuntimeOrigin::none()));
 
+		assert_eq!(CurrentProducer::<Test>::get(), Some(make_id(1)));
+
+		Pallet::<Test>::on_finalize(current_block());
```
Klapeyron, line 28: we can also get block number from system pallet just in case, probably should give the same result as frame_system call
```suggestion
		Pallet::<Test>::on_finalize(System::block_number());
```

toolkit/pallets/block-production-log/src/lib.rs
```diff
@@ -96,5 +112,15 @@ pub mod pallet {
...
+
+		pub fn peek_prefix<'a>(slot: Slot) -> impl Iterator<Item = (Slot, T::BlockProducerId)> {
+			Log::<T>::get().into_iter().filter(move |(s, _)| s <= &slot)
+		}
+
+		pub fn drop_prefix<'a>(slot: Slot) {
+			let entries_left: Vec<_> =
+				Log::<T>::get().into_iter().filter(move |(s, _)| s > &slot).collect();
+			Log::<T>::put(entries_left);
+		}
```
Klapeyron, line 124: looks like not used?

toolkit/pallets/block-production-log/src/lib.rs
```diff
@@ -86,7 +90,19 @@ pub mod pallet {
...
 			LatestBlock::<T>::put(current_block);
 
-			Ok(Log::<T>::append((T::current_slot(), block_producer_id)))
+			Ok(CurrentProducer::<T>::put(block_producer_id))
+		}
+	}
+
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_finalize(block: BlockNumberFor<T>) {
```
Klapeyron, line 99: there is probably small side effect, `Log` will not have `CurrentProducer` for given block, until all extrinsics will be processed and `on_finalize` will be executed, so if some extrininsic will be exeucted after this Mandatory call and try to query the state, data will be not there. The question is if it is relevant from business perspective. Would be also great to cover it in some testcase scenario

toolkit/pallets/block-production-log/src/test.rs
AmbientTea: yeah, that saves a couple of lines

toolkit/pallets/block-production-log/src/lib.rs
AmbientTea: it will

toolkit/pallets/block-production-log/src/lib.rs
AmbientTea: That's exactly what we want. The block hasn't been produced yet.

AmbientTea: covered with an assertion in tests

toolkit/pallets/block-production-log/src/lib.rs
```diff
@@ -86,7 +90,19 @@ pub mod pallet {
 			}?;
```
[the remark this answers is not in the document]
AmbientTea: updated

toolkit/pallets/block-production-log/src/lib.rs
```diff
@@ -86,7 +90,19 @@ pub mod pallet {
...
 			LatestBlock::<T>::put(current_block);
 
-			Ok(Log::<T>::append((T::current_slot(), block_producer_id)))
+			Ok(CurrentProducer::<T>::put(block_producer_id))
```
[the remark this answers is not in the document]
AmbientTea: updated

kpinter-iohk (approved):

toolkit/pallets/block-production-log/src/lib.rs
Klapeyron: ah, ok. Can we double check [this](https://github.com/input-output-hk/partner-chains/blob/master/node/runtime/src/lib.rs#L437) place as well? I think it is changing the behavior of `take_prefix`, so in that place, result may not contain this last element.

toolkit/pallets/block-production-log/src/lib.rs
Klapeyron: and you promise that there will be some tests added once it will be in use? :)

toolkit/pallets/block-production-log/src/lib.rs
```diff
@@ -96,5 +112,15 @@ pub mod pallet {
...
 		}
+
+		pub fn peek_prefix<'a>(slot: Slot) -> impl Iterator<Item = (Slot, T::BlockProducerId)> {
+			Log::<T>::get().into_iter().filter(move |(s, _)| s <= &slot)
+		}
+
+		pub fn drop_prefix<'a>(slot: Slot) {
+			let entries_left: Vec<_> =
+				Log::<T>::get().into_iter().filter(move |(s, _)| s > &slot).collect();
+			Log::<T>::put(entries_left);
```
Klapeyron, line 123: we can optimize it slightly, so try to mutate it and use retain directly on vector instead of call on iterator and collect (please double check due I have not tested it)
```suggestion
			Log::<T>::mutate(|log| log.retain(move |(s, _)| s > &slot));
```

toolkit/pallets/block-production-log/src/lib.rs
AmbientTea: doesn't matter. this is just a placeholder implementation to log to show stuff gets into the log

toolkit/pallets/block-production-log/src/lib.rs
```diff
@@ -96,5 +112,15 @@ pub mod pallet {
 			Log::<T>::put(to_retain);
 			to_return
 		}
+
+		pub fn peek_prefix<'a>(slot: Slot) -> impl Iterator<Item = (Slot, T::BlockProducerId)> {
+			Log::<T>::get().into_iter().filter(move |(s, _)| s <= &slot)
+		}
```
Klapeyron, line 118:
```suggestion
		pub fn peek_prefix<'a>(slot: &Slot) -> impl Iterator<Item = (Slot, T::BlockProducerId)> {
			Log::<T>::get().into_iter().filter(|(s, _)| s <= slot)
		}
```

toolkit/pallets/block-production-log/src/lib.rs
Klapeyron, line 124: we can optimize it slightly, so try to mutate it and use retain directly on vector instead of call on iterator and collect (please double check due I have not tested it)
```suggestion
		pub fn drop_prefix<'a>(slot: &Slot) {
			Log::<T>::mutate(|log| log.retain(|(s, _)| s > slot));
		}
```

AmbientTea: that worked

toolkit/pallets/block-production-log/src/lib.rs
```diff
@@ -96,5 +112,15 @@ pub mod pallet {
 			Log::<T>::put(to_retain);
 			to_return
 		}
+
+		pub fn peek_prefix<'a>(slot: Slot) -> impl Iterator<Item = (Slot, T::BlockProducerId)> {
+			Log::<T>::get().into_iter().filter(move |(s, _)| s <= &slot)
```
Klapeyron, line 117: if we assume that our collection is sorted, then we can utilize `skip_while`:
```suggestion
			Log::<T>::get().into_iter().skip_while(move |(s, _)| s <= &slot)
```

AmbientTea: take_while*. changed

toolkit/pallets/block-production-log/src/lib.rs
```diff
@@ -96,5 +112,15 @@ pub mod pallet {
 			Log::<T>::put(to_retain);
 			to_return
 		}
+
```
Klapeyron, line 115: similar can be applied to `take_prefix`, so we can use a fact that our collection is sorted, so `partition_point` shall be faster and use binary_search under the hood. Up to you if we would like to apply it if any
```rust
			let removed_prefix = Log::<T>::mutate(|log| {
				let pos = log.partition_point(|(s, _)| s <= slot);
				log.drain(..pos).collect()
			});
			removed_prefix
```

toolkit/pallets/block-production-log/src/lib.rs
```diff
@@ -96,5 +112,15 @@ pub mod pallet {
...
+		pub fn peek_prefix<'a>(slot: Slot) -> impl Iterator<Item = (Slot, T::BlockProducerId)> {
```
Klapeyron, line 116: lifetime is not needed 
```suggestion
		pub fn peek_prefix(slot: Slot) -> impl Iterator<Item = (Slot, T::BlockProducerId)> {
```

toolkit/pallets/block-production-log/src/lib.rs
```diff
@@ -96,5 +112,13 @@ pub mod pallet {
...
+			Log::<T>::get().into_iter().take_while(move |(s, _)| s <= &slot)
+		}
+
+		pub fn drop_prefix<'a>(slot: Slot) {
```
Klapeyron, line 120:
```suggestion
		pub fn drop_prefix(slot: Slot) {
```

toolkit/pallets/block-production-log/src/lib.rs
AmbientTea: this function most likely will not be needed, but changed just in case

Klapeyron: please double check if it has the same behavior as existing one, I wrote it on the fly and not tested xd

PR #520, opened 2025-02-25
e2e tests for address association

rsporny:
added:
- new e2e test verifying address association

changed:
- refactored test's logs obfuscation


e2e-tests/src/password_filter.py
```diff
@@ -1,13 +0,0 @@
-import re
-import logging
-
-
-class PasswordFilter(logging.Filter):
-    """Filter class to obscure sensitive information from logging."""
```
chrispalaskas, line 6: Did you remove this completely? I meant vkey can be public but not skey.

chrispalaskas: Oh I see, you moved the log_filter to src.

chrispalaskas (approved):

ladamesny (approved):

e2e-tests/src/password_filter.py
rsporny: yes, we had two filters, I merged them and moved to src

PR #522, opened 2025-02-26
Fix e2e python env in nix shell for linux

gilligan:
This change extends the LD_LIBRARY_PATH to include libs required by numpy. This wouldn't be necessary if we were providing python packages through Nix but we aren't so we need to make them visible

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/dev/nix/shell.nix
+++ b/dev/nix/shell.nix
@@ -9,18 +9,20 @@
           file = ../../rust-toolchain.toml;
           sha256 = "VZZnlyP69+Y3crrLHQyJirqlHrTtGTsyiSnZB8jEvVo=";
         };
-    in {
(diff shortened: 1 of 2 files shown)


rsporny (approved):

AmbientTea (approved):

alexromanov (approved):

tgunnoe (approved):
Thank you

kpinter-iohk (approved):
Works great, but since you updated the docs, [there is another place where it is stated that Python doesn't work from Nix](https://github.com/input-output-hk/partner-chains/blob/master/e2e-tests/docs/run-tests-on-local-env.md?plain=1#L6).

PR #527, opened 2025-03-04
fix: pallet-block-production-log benchmarks using BenchmarkHelper trait

LGLO:
# Description

## Problem:

Not every 32 bytes array is a valid seed for ECDSA. If invalid one is passed, the code panics with misleading error that array length was a problem.
This manifests only in benchmarks (we have 1), where we don't care about diverse data.

## Solution

Replace invalid `From<u8;32>` with the `BenchmarkHelper` trait that has one function to give `BlockProducerId` in benchmarks run.

`BenchmarkHelper` is used 10 times in polkadot-sdk, so it is a semi-pattern for me.

# Checklist

- [x] Self-reviewed the diff


node/runtime/src/lib.rs
```diff
@@ -491,11 +491,19 @@ impl From<CommitteeMember<CrossChainPublic, SessionKeys>> for BlockAuthor {
 		}
 	}
 }
+
+#[cfg(feature = "runtime-benchmarks")]
```
LGLO, line 495: I admit that I don't like additional 5 `#[cfg(feature = "runtime-benchmarks")]` flags.

kpinter-iohk (approved):

node/runtime/src/lib.rs
```diff
@@ -491,11 +491,19 @@ impl From<CommitteeMember<CrossChainPublic, SessionKeys>> for BlockAuthor {
...
+
 #[cfg(feature = "runtime-benchmarks")]
-impl From<[u8; 32]> for BlockAuthor {
-	fn from(arr: [u8; 32]) -> Self {
+impl pallet_block_production_log::benchmarking::BenchmarkHelper<BlockAuthor>
+	for PalletBlockProductionLogBenchmarkHelper
+{
+	fn producer_id() -> BlockAuthor {
+		let arr =
+			hex_literal::hex!("9d61b19deffd5a60ba844af492ec2cc44449c5697b326919703bac031cae7f60");
```
Klapeyron, line 504: it is somehow possible in other parts of the code to invject invalid ECDSA key in production part of the code?

toolkit/pallets/block-production-log/src/mock.rs
```diff
@@ -62,13 +62,26 @@ impl frame_system::Config for Test {
...
 }
 
+#[cfg(feature = "runtime-benchmarks")]
+pub struct PalletBlockProductionLogBenchmarkHelper;
+
+#[cfg(feature = "runtime-benchmarks")]
+impl crate::benchmarking::BenchmarkHelper<[u8; 32]> for PalletBlockProductionLogBenchmarkHelper {
+	fn producer_id() -> [u8; 32] {
+		Default::default()
+	}
```
Klapeyron, line 72: empty producer_id is considered valid?

toolkit/pallets/block-production-log/src/mock.rs
```diff
@@ -62,13 +62,26 @@ impl frame_system::Config for Test {
...
 impl crate::pallet::Config for Test {
 	type BlockProducerId = [u8; 32];
 	type WeightInfo = ();
 
 	fn current_slot() -> Slot {
 		Slot::from(1001000)
 	}
+
+	#[cfg(feature = "runtime-benchmarks")]
+	type BenchmarkHelper = PalletBlockProductionLogBenchmarkHelper;
```
Klapeyron, line 84: if it will have impact only on benchmarks, maybe there is no need to make a parametrization and just hardcode some valid address in benchmarks code

toolkit/pallets/block-production-log/src/mock.rs
LGLO: It is not empty. It is 32 bytes of zero value.

toolkit/pallets/block-production-log/src/mock.rs
LGLO: I would like to keep it like in polkadot-sdk, where every definition is parameterized. Maybe without parametrization it would signal some unused code?

node/runtime/src/lib.rs
LGLO: In CLIs but it is harmless, because invalid key will crash CLI only when someone gives invalid key.

Node doesn't sign with these keys. Node uses keys from keystore.

LGLO: But I'll try to construct public key in a cheaper way. Going from seed (private) to public is extra cost.

toolkit/pallets/block-production-log/src/mock.rs
Klapeyron: yep thats possible

node/runtime/src/lib.rs
AmbientTea: how about adding a `benchmark_helpers` module with one flag?

AmbientTea (approved):

toolkit/pallets/block-production-log/src/mock.rs
Klapeyron: yep, but this "zero" filled array which will be used will be considered valid?

LGLO: Yes, this is just to have some similar space occupation.

node/runtime/src/lib.rs
LGLO: I use them in different scopes, like pallet type definition. polkadot-sdk uses them like I do here

PR #529, opened 2025-03-04
test: fix non hex number in address association test

rsporny:
fixes:
- stake key hash conversion to hex rairly could be broken if the leading number was 0

added:
- create one log file for each thread

### DIFF
--- a/e2e-tests/src/substrate_api.py
+++ b/e2e-tests/src/substrate_api.py
@@ -750,6 +750,6 @@ class SubstrateApi(BlockchainApi):
         return tx
 
     def get_address_association(self, stake_key_hash):
-        result = self.substrate.query("AddressAssociations", "AddressAssociations", [hex(int(stake_key_hash, 16))])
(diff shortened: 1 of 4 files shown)


e2e-tests/src/substrate_api.py
```diff
@@ -750,6 +750,6 @@ def submit_address_association(self, signature, wallet):
         return tx
 
     def get_address_association(self, stake_key_hash):
-        result = self.substrate.query("AddressAssociations", "AddressAssociations", [hex(int(stake_key_hash, 16))])
+        result = self.substrate.query("AddressAssociations", "AddressAssociations", [f"0x{stake_key_hash}"])
```
AmbientTea, line 753: just a heads-up that depending on where this key has comes from, it can already have a 0x prefix (we're kinda inconsistent about it in the node RPC/CLI outputs)

AmbientTea (approved):

PR #530, opened 2025-03-04
change: require block production log pallet inherent only after data becomes available

AmbientTea:
# Description

- if `LastBlock` is empty, the pallet's inherent is NOT required. it becomes required, once the storage is set.
- changes the `get_current_author` API method to return Option, and the InherentDataProvider

The rationale for this is to allow the pallet to be added as part of a wider runtime upgrade and avoid issues on the first block after the upgrade, caused by access to yet-unmigrated storages. This is achieved by treating the `LastBlock` as an initialized state flag.

# Checklist

- [x] Self-reviewed the diff

# Note on CI
for you.

### DIFF
--- a/toolkit/pallets/block-production-log/src/lib.rs
+++ b/toolkit/pallets/block-production-log/src/lib.rs
@@ -43,6 +43,11 @@ pub mod pallet {
 	#[pallet::storage]
 	pub type LatestBlock<T: Config> = StorageValue<_, BlockNumberFor<T>, OptionQuery>;
 
+	/// This flag keeps track of whether the pallet has been initialized by receiving non-empty inherent data.
+	/// In initialized state, the inherent is requires for every block.
+	#[pallet::storage]
+	pub type Initialized<T: Config> = StorageValue<_, bool, ValueQuery>;
+
 	#[pallet::inherent]
 	impl<T: Config> ProvideInherent for Pallet<T> {
 		type Call = Call<T>;
@@ -50,19 +55,22 @@ pub mod pallet {
 		const INHERENT_IDENTIFIER: InherentIdentifier = INHERENT_IDENTIFIER;
 
 		fn create_inherent(data: &InherentData) -> Option<Self::Call> {
-			let block_producer_id = data
-				.get_data::<T::BlockProducerId>(&Self::INHERENT_IDENTIFIER)
-				.expect("Block Production Log inherent data not correctly encoded")
-				.expect("Block Production Log inherent data must be provided");
-			Some(Call::append { block_producer_id })
+			Self::decode_inherent_data(data)
+				.unwrap()
+				.map(|block_producer_id| Call::append { block_producer_id })
 		}
 
 		fn is_inherent(call: &Self::Call) -> bool {
 			matches!(call, Call::append { .. })
 		}
 
-		fn is_inherent_required(_: &InherentData) -> Result<Option<Self::Error>, Self::Error> {
-			Ok(Some(Self::Error::InherentRequired))
+		fn is_inherent_required(data: &InherentData) -> Result<Option<Self::Error>, Self::Error> {
+			let has_data = Self::decode_inherent_data(data)?.is_some();
+			if has_data || Initialized::<T>::get() {
+				Ok(Some(Self::Error::InherentRequired))
+			} else {
+				Ok(None)
+			}
 		}
 	}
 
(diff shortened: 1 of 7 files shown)


toolkit/pallets/block-production-log/src/lib.rs
```diff
@@ -43,26 +43,34 @@ pub mod pallet {
 	#[pallet::storage]
 	pub type LatestBlock<T: Config> = StorageValue<_, BlockNumberFor<T>, OptionQuery>;
 
+	/// This flag keeps track of whether the pallet has been initialized by receiving non-empty inherent data.
+	/// In initialized state, the inherent is requires for every block.
+	#[pallet::storage]
+	pub type Initialized<T: Config> = StorageValue<_, bool, ValueQuery>;
+
```
Klapeyron, line 50: you can maybe try to use other variable to check if it was already initialized, maintaining some booleans to represents if some other state is set or not may cause errors.
You can take a look on `block-production-log`, where instead of `ValueQuery` an `OptionQuery` is utilized

Klapeyron (changes requested):
.

AmbientTea:
> .

let me go mon

toolkit/pallets/block-production-log/src/lib.rs
```diff
@@ -43,26 +43,34 @@ pub mod pallet {
 	#[pallet::storage]
 	pub type LatestBlock<T: Config> = StorageValue<_, BlockNumberFor<T>, OptionQuery>;
 
+	/// This flag keeps track of whether the pallet has been initialized by receiving non-empty inherent data.
+	/// In initialized state, the inherent is requires for every block.
```
kpinter-iohk, line 47:
```suggestion
	/// In the initialized state, the inherent is required for every block.
```

kpinter-iohk (approved):

Klapeyron (dismissed):
Can you provide more context why we actually have an issue with first block after upgrade?

AmbientTea:
> Can you provide more context why we actually have an issue with first block after upgrade?

> Why there is "yet-unmigrated storage" after upgrade?

Migrations are executed during initialization of the block AFTER the block in which the runtime upgrade was performed. So the normal order of events is:
1. In block N, `setCode` is called and runtime is changed in the storage. The rest of the block weight is consumed.
2. Block N is finalized and propagated
3. Another block producer starts producing block N+1, selected based on the authority set in Aura pallet (not migrated)
4. Block producer constructs IDPs
5. Block producer invokes `on_initialization`, which performs necessary storage migrations before any other runtime code is run
6. Block is produced as normal

Now, our issue occurs at step 4, when BlockProductionInherentDataProvider tries to read SessionValidatorManagement pallet's committee data and fails, because it is using a runtime API of the new runtime, but the storage hasn't yet been migrated.

This is the issue this PR is fixing, by allowing for this data to be missing and the inherent to be skipped, but only if this is the beginning of the pallet's operation

Klapeyron:
> Block producer invokes on_initialization, which performs necessary storage migrations before any other runtime code is run

Docs say that `on_initialize` stuff is beeing executed after storage migration is performed:

> Storage migrations must be executed precisely during the runtime upgrade process to ensure data consistency and prevent [runtime panics](https://doc.rust-lang.org/std/macro.panic.html). The migration code needs to run as follows:
* After the new runtime is deployed
* Before any other code from the new runtime executes
* Before any [on_initialize](https://paritytech.github.io/polkadot-sdk/master/frame_support/traits/trait.Hooks.html#method.on_initialize) hooks run
* Before any transactions are processed

Have we verified this behavior via some dummy upgrade like `try-runtime` or something? If there is some race within the state, then it look more like a missing storage migration. I am afraid, that if we will introduce it in current form, then we will be not able to get rid of this "Initialized" feature. If for instance we will deploy a runtime upgrade that is adding this PR (lets say runtime_version 151) and later on we will remove it as part of 152 (due we will assume that state is initialized), then if somebody will invoke upgrade from 150 -> 152 without 151, then may have some issues

AmbientTea:
I checked the Substrate codebase. Migrations are run as part of block initialization (`initialize_block_impl` in  `frame/executive/src/lib.rs`). This is consistent with what I wrote: IDP creation uses runtime API, which has already been upgraded in previous block before the next one is initialized.

AmbientTea:
I've found a way to get rid of the flag, by just checking if `LastBlock` has been set.

LGLO:
So, is the order
* Block N: runtime code upgrade
* IDPs for Block N+1
* Migrations
* Rest of Block N+1
?

AmbientTea:
> So, is the order
> 
> * Block N: runtime code upgrade
> * IDPs for Block N+1
> * Migrations
> * Rest of Block N+1
>   ?

Exactly, the IDP get constructed before the block production begins

PR #535, opened 2025-03-05
fix: production log reading stale slot number

AmbientTea:
# Description

- Adds a slot parameter to `get_current_author` in block production log's runtime API; and changes the IDP to pass it.

# Checklist

- [ ] Self-reviewed the diff

# Note on CI
for you.


node/runtime/src/lib.rs
```diff
@@ -950,9 +950,8 @@ impl_runtime_apis! {
 	}
 
 	impl sp_block_production_log::BlockProductionLogApi<Block, CommitteeMember<CrossChainPublic, SessionKeys>>  for Runtime {
-		fn get_current_author() -> CommitteeMember<CrossChainPublic, SessionKeys> {
+		fn get_current_author(slot: Slot) -> CommitteeMember<CrossChainPublic, SessionKeys> {
```
Klapeyron, line 953: we should adjust call name as well, if we parametrize with slot, then it is not `current` author

AmbientTea: makes sense, changed

Klapeyron: what about simply `get_author`?

Klapeyron: we can add some comment for the trait doc if the name will be not fully meaningful

node/node/src/tests/runtime_api_mock.rs
```diff
@@ -114,7 +114,7 @@ sp_api::mock_impl_runtime_apis! {
 	}
 
 	impl sp_block_production_log::BlockProductionLogApi<Block, CommitteeMember<CrossChainPublic, SessionKeys>> for TestApi {
-		fn get_current_author() -> CommitteeMember<CrossChainPublic, SessionKeys> {
+		fn get_current_committee_member_for_slot(_slot: Slot) -> CommitteeMember<CrossChainPublic, SessionKeys> {
```
LGLO, line 117: I understand problem with "current" but we can make it `get_author`. There can be many committee member at given slot - basically all committee.

toolkit/primitives/block-production-log/src/test.rs
```diff
@@ -36,9 +37,12 @@ sp_api::mock_impl_runtime_apis! {
...
-			.expect("Should not fail");
+	let provider = BlockAuthorInherentProvider::<Author>::new(
+		&mock_api,
+		<Block as BlockT>::Hash::default(),
+		Slot::from(42),
+	)
+	.expect("Should not fail");
 
-	assert_eq!(provider.author.unwrap(), mock_api.author.into());
+	assert_eq!(provider.author.unwrap(), mock_api.author as Author);
```
Klapeyron, line 47: try to keep into if possible

AmbientTea: for some reason it doesn't work. `as` does the job

kpinter-iohk (approved):

toolkit/primitives/block-production-log/src/test.rs
Klapeyron: probably needs some additional type annotations, try:
```suggestion
	assert_eq!(provider.author.unwrap(), Author::from(mock_api.author));
```

PR #539, opened 2025-03-06
feat: simplify 'deposit' by removing token (asset id) parameter

LGLO:
# Description

Removed `token` parameter from the `smart-contracts reserve deposit` command, because asset id is already present in the reserve settings.

Ref: ETCM-9243 (Point 1 & 3)

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/docs/developer-guides/native-token-reserve-management.md
+++ b/docs/developer-guides/native-token-reserve-management.md
@@ -154,7 +154,7 @@ Command template:
 
 ### 1.4 Creating your token reserve
 
-Objective: Create a reserve for your token and define the release function `V`.
+Objective: Create a reserve for your token and define the release function `V`. There can be only one reserve, therefore one reserve token, for a partner chain.
 
 #### 1.4.1 Run the `reserve create` command
 
@@ -188,9 +188,9 @@ Command template:
 
 * `--total-accrued-function-script-hash`: The script hash of your `V` function
 
-* `--token`: Reserve token asset id encoded in form `<policy_id_hex>.<asset_name_hex>`
+* `--token`: Reserve token asset id encoded in form `<policy_id_hex>.<asset_name_hex>`, this sets the reserve token asset id
 
-* `--initial-deposit-amount`: The initial amount of tokens to deposit into the reserve
+* `--initial-deposit-amount`: The initial amount of tokens to deposit into the reserve, at least this amout of tokens has to available in the payment wallet
 
 * `--payment-key-file`: Your payment key file for transaction signing. This key has to have `initial-deposit-amount` in its wallet.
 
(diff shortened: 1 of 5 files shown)


toolkit/offchain/src/reserve/deposit.rs
```diff
@@ -53,10 +53,11 @@ pub async fn deposit_to_reserve<
...
 	let reserve = ReserveData::get(genesis_utxo, &ctx, client).await?;
-	let utxo = reserve.get_reserve_utxo(&ctx, client).await?.utxo;
-	let current_amount = get_token_amount(&utxo, &parameters.token);
-	let token_amount =
-		TokenAmount { token: parameters.token, amount: current_amount + parameters.amount };
+	let reserve_utxo = reserve.get_reserve_utxo(&ctx, client).await?;
+	let utxo = reserve_utxo.utxo;
+	let token = reserve_utxo.datum.immutable_settings.token;
+	let current_amount = get_token_amount(&utxo, &token);
+	let token_amount = TokenAmount { token, amount: current_amount + amount };
```
LGLO, line 60: I think we should refactor `TokenAmount` by renaming `token` to `asset_id`, but I want to keep it outside this PR.

Also I think that `ReserveUtxo` could benefit from some renaming, so we don't have `reserve_utxo.utxo` but also I don't want to mix refactorings with business logic change this time.

kpinter-iohk: Regrarding ReserveUtxo: IIRC this is on me and I agree. ReserveUtxo has the reserve utxo **and** the reserve datum extracted from that utxo. The reason we have utxo and datum together is that to find the correct utxo we need to decode the datum and even though we can get the datum from the utxo it just felt stupid to throw the datum away because we **do** need it later.
When naming  this we already had ReserveData for something else and ReserveUtxoWithDatum felt clumsy.

kpinter-iohk (approved):
:tada:

toolkit/offchain/src/reserve/deposit.rs
LGLO: Separately we'll see if we need OgmiosUtxo or maybe we should parse it completely (we parse datum already). I just want to have this discussion separately.

docs/developer-guides/native-token-reserve-management.md
```diff
@@ -188,9 +188,9 @@ Command template:
 
 * `--total-accrued-function-script-hash`: The script hash of your `V` function
 
-* `--token`: Reserve token asset id encoded in form `<policy_id_hex>.<asset_name_hex>`
+* `--token`: Reserve token asset id encoded in form `<policy_id_hex>.<asset_name_hex>`, this sets the reserve token asset id
 
-* `--initial-deposit-amount`: The initial amount of tokens to deposit into the reserve
+* `--initial-deposit-amount`: The initial amount of tokens to deposit into the reserve, at least this amout of tokens has to available in the payment wallet
```
AmbientTea, line 193:
```suggestion
* `--initial-deposit-amount`: The initial amount of tokens to deposit into the reserve, at least this amount of tokens has to be available in the payment wallet
```

AmbientTea (approved):

PR #549, opened 2025-03-07
multi-sig for d-param

jankun4:
# Description

[link](https://input-output.atlassian.net/browse/ETCM-9518)

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/toolkit/smart-contracts/commands/src/d_parameter.rs
+++ b/toolkit/smart-contracts/commands/src/d_parameter.rs
@@ -27,7 +27,7 @@ impl UpsertDParameterCmd {
 		};
 		let client = self.common_arguments.get_ogmios_client().await?;
 
-		upsert_d_param(
+		let result = upsert_d_param(
 			self.genesis_utxo,
 			&d_param,
 			&payment_key,
@@ -35,7 +35,10 @@ impl UpsertDParameterCmd {
 			&FixedDelayRetries::two_minutes(),
 		)
 		.await?;
-
+		match result {
+			Some(result) => println!("{}", serde_json::to_value(result)?),
+			None => println!("{{}}"),
+		}
 		Ok(())
 	}
 }
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,7 @@ This changelog is based on [Keep A Changelog](https://keepachangelog.com/en/1.1.
 
 ## Changed
 
+* `upsert-d-parameter` command now works with multi-sig.
 * `governance update` command now accepts multiple governance authority key hashes, not just one. It also takes `new-governance-threshold` parameter, which is the number of signatures required to perform governance action.
 * `governance init` and `governance update` will set Multisig policy implemented with ALeastN Native Script, instead of custom policy implemented as Plutus Script in partner-chains-smart-contracts. This policy doesn't require to set `required_signers` field in the transaction making it more user friendly.
 * Extracted the "Ariadne" committee selection algorithm to the `selection` crate.
--- a/toolkit/partner-chains-cli/src/setup_main_chain_state/tests.rs
+++ b/toolkit/partner-chains-cli/src/setup_main_chain_state/tests.rs
@@ -11,9 +11,7 @@ use crate::{verify_json, CmdRun};
 use hex_literal::hex;
 use partner_chains_cardano_offchain::multisig::MultiSigSmartContractResult;
 use serde_json::json;
-use sidechain_domain::{
-	AuraPublicKey, DParameter, GrandpaPublicKey, McTxHash, SidechainPublicKey, UtxoId,
-};
+use sidechain_domain::{AuraPublicKey, DParameter, GrandpaPublicKey, SidechainPublicKey, UtxoId};
 use sp_core::offchain::Timestamp;
 
 #[test]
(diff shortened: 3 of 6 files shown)


toolkit/offchain/src/d_param/mod.rs
```diff
@@ -130,27 +153,52 @@ async fn insert_d_param<C: QueryLedgerState + Transactions + QueryNetwork>(
...
+		log::info!("Transaction submitted: {}", hex::encode(tx_id.0));
+		Ok(TxHash(tx_id))
+	} else {
+		let tx_envelope = json!(
+			{ "type": "Unwitnessed Tx ConwayEra",
+			  "description": "",
+			  "cborHex": hex::encode(tx.to_bytes())
+			}
+		);
+		log::info!("Transaction envelope: {}", tx_envelope);
```
LGLO, line 199: WDYT about printing proper JSON to stdout? I think it has potential to help with other tooling and tests.

toolkit/offchain/tests/integration_tests.rs
```diff
@@ -273,24 +274,28 @@ async fn run_upsert_d_param<
...
 		&DParameter { num_permissioned_candidates, num_registered_candidates },
 		pkey,
+		governance,
 		client,
 		&FixedDelayRetries::new(Duration::from_millis(500), 100),
 	)
 	.await
 	.unwrap();
-	if let Some(tx_hash) = tx_hash {
+	if let Some(TxHash(tx_hash)) = tx_hash {
```
LGLO, line 290: Cosmetic: `TxHash(tx_hash)` is confusing, could you use qualified name for the outer one, please? `McSmartContractResult::TxHash`

toolkit/cli/smart-contracts-commands/src/d_parameter.rs
```diff
@@ -27,10 +30,17 @@ impl UpsertDParameterCmd {
 		};
 		let client = self.common_arguments.get_ogmios_client().await?;
 
+		let governance_authority = if self.governance_authority.len() == 0 {
```
kpinter-iohk, line 33:
```suggestion
		let governance_authority = if self.governance_authority.is_empty() {
```

toolkit/offchain/src/d_param/mod.rs
```diff
@@ -4,23 +4,33 @@
...
-use cardano_serialization_lib::{PlutusData, Transaction, TransactionBuilder, TxInputsBuilder};
+use cardano_serialization_lib::{
+	PlutusData, Transaction, TransactionBuilder, TxInputsBuilder,
+};
 use ogmios_client::query_ledger_state::QueryUtxoByUtxoId;
 use ogmios_client::{
 	query_ledger_state::QueryLedgerState, query_network::QueryNetwork, transactions::Transactions,
 	types::OgmiosUtxo,
 };
+// use pallas_primitives::conway::VKeyWitness;
```
kpinter-iohk, line 26: leftover

toolkit/cli/smart-contracts-commands/src/d_parameter.rs
kpinter-iohk: Actually this check should panic if `governance_authority` is empty since `clap` is supposed to ensure that Vec is not empty.

AmbientTea: I wouldn't panic here, since using the payment key's identity as the authority is a very convenient default

kpinter-iohk: It's not about what the default is, it's that this code path should be unreachable. Either this path should panic as impossible, or we should drop `num_args = 1..,` from the arg parser if we want to make it optional and allow for a default value to be used when `governance_authority` is not provided.

toolkit/offchain/src/d_param/mod.rs
jankun4: You mean printing directly to stdout, not through `log` crate, or you mean printing just the JSON, without the "Transaction envelope:" part?

toolkit/offchain/src/d_param/mod.rs
```diff
@@ -140,17 +147,41 @@ async fn insert_d_param<C: QueryLedgerState + Transactions + QueryNetwork>(
...
+				e,
+				hex::encode(signed_tx)
+			)
+		})?;
+		let tx_id = McTxHash(res.transaction.id);
+		log::info!("Transaction submitted: {}", hex::encode(tx_id.0));
+		Ok(TxHash(tx_id))
+	} else {
+		let authorities = gov_data.policy.key_hashes();
+		if authorities.contains(&ctx.payment_key_hash().to_bytes().try_into().unwrap()) {
```
LGLO, line 164: I would say it doesn't matter if `ctx.payment_key_hash` is an authority or not. Since it pays for the TX, it needs to sign it. I propose simplifying it by dropping this check and else case and dropping the `key_hashes()` combinator.

LGLO (approved):

toolkit/smart-contracts/commands/src/d_parameter.rs
```diff
@@ -27,15 +27,15 @@ impl UpsertDParameterCmd {
...
+		let result = upsert_d_param(
 			self.genesis_utxo,
 			&d_param,
 			&payment_key,
 			&client,
 			&FixedDelayRetries::two_minutes(),
 		)
 		.await?;
-
+		println!("{}", serde_json::to_value(result)?);
```
LGLO, line 38: This will print `null` could we have `{}` if option is none? Both are valid jsons but we use the latter one for permissioned candidates

PR #557, opened 2025-03-11
ETCM-9535 provide default for test_helper_pallet::participation_data_release_period

kpinter-iohk:
# Description

`participation_data_release_period` in `test_helper_pallet` does not have an explicit default, so it defaults to 0 (for example when upgrading the runtime). This causes test_helper_pallet to cause an execution failure because it is used on the right-hand side of a modulo operation.
This PR provides the default value of 30.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/node/runtime/src/genesis_config_presets.rs
+++ b/node/runtime/src/genesis_config_presets.rs
@@ -45,7 +45,17 @@ fn testnet_genesis(
 			..Default::default()
 		},
 		sudo: SudoConfig { key: Some(root) },
-		..Default::default()
+		test_helper_pallet: crate::test_helper_pallet::GenesisConfig {
+			participation_data_release_period: 30,
+			_phantom: core::marker::PhantomData,
+		},
+		system: Default::default(),
+		transaction_payment: Default::default(),
+		sidechain: Default::default(),
+		session_committee_management: Default::default(),
+		pallet_session: Default::default(),
+		session: Default::default(),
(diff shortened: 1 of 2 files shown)


node/runtime/src/genesis_config_presets.rs
```diff
@@ -45,7 +45,17 @@ fn testnet_genesis(
 			..Default::default()
 		},
 		sudo: SudoConfig { key: Some(root) },
-		..Default::default()
+		test_helper_pallet: crate::test_helper_pallet::GenesisConfig {
+			participation_data_release_period: 30,
```
kpinter-iohk, line 49: This was the first attempt at fixing the issue but turned out not to be necessary. I'm not sure when this is used but seems like we should have it. Please let me know if I should remove this.

node/runtime/src/genesis_config_presets.rs
```diff
@@ -45,7 +45,17 @@ fn testnet_genesis(
 			..Default::default()
 		},
 		sudo: SudoConfig { key: Some(root) },
-		..Default::default()
```
kpinter-iohk, line 48: Seems like a bad idea to me to have all fields default initialized here. When adding a new pallet this makes it easier to forget to consider what defaults are needed.

gilligan:
extra kudos for providing a good PR description and additional comments on the changes 👍

PR #564, opened 2025-03-13
remove: block-rewards feature (ETCM-9577)

AmbientTea:
# Description

Removes the `block-rewards` pallet and its primitives crate. This feature was superseded by the `block-production-log` and `block-participation` instead.

# Checklist

- [x] Self-reviewed the diff


node/runtime/src/test_helper_pallet.rs
```diff
@@ -62,10 +62,7 @@ pub mod pallet {
 
 	impl<T: Config> sp_sidechain::OnNewEpoch for Pallet<T> {
 		fn on_new_epoch(old_epoch: ScEpochNumber, _new_epoch: ScEpochNumber) -> sp_weights::Weight {
-			let rewards = pallet_block_rewards::Pallet::<T>::get_rewards_and_clear();
-			log::info!("Rewards accrued in epoch {old_epoch}: {rewards:?}");
-
-			crate::RuntimeDbWeight::get().reads_writes(1, 1)
+			crate::RuntimeDbWeight::get().reads_writes(0, 0)
```
AmbientTea, line 65: I'm leaving this handler empty here so it's clear later that this helper pallet is where eventual handler logic for other features should go

Klapeyron: we can try to add some `#[deprecated]` information

kpinter-iohk (approved):

node/runtime/src/test_helper_pallet.rs
AmbientTea: But it's not deprecated, it's just not used for now.

PR #578, opened 2025-03-17
change: ETCM-9592 Move "Ariadne" to selection crate

AmbientTea:
# Description

- disentangles the "Ariadne" algorithm from the authority-selection-inherents crate's code and moves into the selection crate
- adds property tests for it
- moves substrate-aware helper functions out of the selection crate, so it can be a pure Rust crate

There's quite a lot LOC changed here, but a big chunk of it is tests and moving stuff around. Main action is between `select_authorities.rs` and newly added `ariadne.rs`, where I had to make some types generic in the moved code.

# Checklist

- [x] Self-reviewed the diff


toolkit/primitives/authority-selection-inherents/src/filter_invalid_candidates.rs
```diff
@@ -31,18 +31,29 @@ pub struct CandidateWithStake<TAccountId, TAccountKeys> {
...
 }
 
-#[derive(Clone, Debug, Encode, Decode, PartialEq)]
+#[derive(Clone, Debug, Encode, Decode, PartialEq, Eq)]
 pub enum Candidate<TAccountId, TAccountKeys> {
 	Permissioned(PermissionedCandidate<TAccountId, TAccountKeys>),
 	Registered(CandidateWithStake<TAccountId, TAccountKeys>),
 }
 
+impl<TAccountId: Ord + Eq, TAccountKeys: Eq> PartialOrd for Candidate<TAccountId, TAccountKeys> {
```
AmbientTea, line 46: This manual `Ord` is so the ordering stays the same as it was previously when using our types, but I don't think that's really necessary, since any total ordering will do just fine (and changing it wouldn't break anything since the selection function is kept in the runtime)

toolkit/primitives/authority-selection-inherents/src/select_authorities.rs
```diff
@@ -3,107 +3,74 @@
...
-	} else {
-		1 // if there are no permissioned candidates, trustless candidates should be selected using unmodified stake
-	};
-	trustless_candidates
-		.iter()
-		.map(|c| {
-			(Candidate::Registered(c.clone()), u128::from(c.stake_delegation.0) * weight_factor)
-		})
-		.collect()
+pub fn seed_from_nonce_and_sc_epoch(
```
AmbientTea, line 48: moved from `selection`

kpinter-iohk (approved):

toolkit/primitives/authority-selection-inherents/src/filter_invalid_candidates.rs
Klapeyron: yep, if it will have no impact on existing data and just for calculation of new commitee it may be ordered slightly different, then I will also vote for keeping the Ord generated with a macro.
But also maybe it is worth to look into `select_authorities` itself, maybe we can remove Ord/PartialOrd requirement and use a trait `Candidate` instead, so `select_authorities` will `sort_by` candidate.account_id(), which may be more readable from the algorithm perspective.

node/runtime/src/mock.rs
```diff
@@ -156,12 +156,12 @@ impl pallet_session_validator_management::Config for Test {
 		input: AuthoritySelectionInputs,
 		_sidechain_epoch: ScEpochNumber,
 	) -> Option<BoundedVec<(Self::AuthorityId, Self::AuthorityKeys), Self::MaxValidators>> {
-		let candidates: Vec<_> = filter_trustless_candidates_registrations(
-			input.registered_candidates,
-			Sidechain::genesis_utxo(),
-		)
+		let candidates: Vec<_> = filter_trustless_candidates_registrations::<
```
Klapeyron, line 159: general question regarding candidates, do we have a warranty, that Vec from `filter_trustless_candidates_registrations` will fit into 32 elements boundary?
What about further restricting of `filter_trustless_candidates_registrations ` and `select_authorities`, so they will be operating on BoundedVec instead of Vec?

toolkit/primitives/selection/src/ariadne.rs
```diff
@@ -0,0 +1,189 @@
...
+	use crate::tests::assert_subset;
+	use crate::tests::*;
+	use quickcheck::*;
+	use quickcheck_macros::quickcheck;
+	use std::collections::HashSet;
+
+	const MAX_CANDIDATES: usize = 50;
+
+	#[quickcheck]
+	fn ariadne_selection(
```
LGLO, line 100: would be nice to have tests name reflect tested property

toolkit/primitives/authority-selection-inherents/src/filter_invalid_candidates.rs
AmbientTea: That's what I started with, but I didn't really like that, because it assumes that we sort by ID etc. and expose that detail. I was thinking about moving the sorting out of the selection completely, and expecting the caller to supply data in a stable order, but that introduces a possibility for a bug, so I dropped that too. I think `Ord` strikes a good balance for now.

LGLO (approved):

toolkit/primitives/authority-selection-inherents/src/filter_invalid_candidates.rs
```diff
@@ -31,18 +31,29 @@ pub struct CandidateWithStake<TAccountId, TAccountKeys> {
...
+	fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
+		Some(self.account_id().cmp(other.account_id()))
```
Klapeyron, line 48: this `cmp` looks like copy pasted

AmbientTea: reused `cmp` instead

node/runtime/src/mock.rs
AmbientTea: There's no boundary for candidates, only for the committee.

toolkit/primitives/authority-selection-inherents/src/filter_invalid_candidates.rs
```diff
@@ -31,18 +31,29 @@ pub struct CandidateWithStake<TAccountId, TAccountKeys> {
...
+impl<TAccountId: Ord + Eq, TAccountKeys: Eq> PartialOrd for Candidate<TAccountId, TAccountKeys> {
+	fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
+		Some(self.cmp(other))
+	}
+}
+impl<TAccountId: Ord + Eq, TAccountKeys: Eq> Ord for Candidate<TAccountId, TAccountKeys> {
+	fn cmp(&self, other: &Self) -> core::cmp::Ordering {
+		self.account_id().cmp(other.account_id())
+	}
+}
```
Klapeyron, line 55: there is no need to restrict Eq, PartialOrd/Ord implies that
```suggestion
impl<TAccountId: Ord, TAccountKeys: Eq> PartialOrd for Candidate<TAccountId, TAccountKeys> {
	fn partial_cmp(&self, other: &Self) -> Option<core::cmp::Ordering> {
		Some(self.cmp(other))
	}
}
impl<TAccountId: Ord, TAccountKeys: Eq> Ord for Candidate<TAccountId, TAccountKeys> {
	fn cmp(&self, other: &Self) -> core::cmp::Ordering {
		self.account_id().cmp(other.account_id())
	}
}
```

AmbientTea:
Successfully tested this on my local 3-node chain (upgrade from 1.5). Will merge

toolkit/primitives/authority-selection-inherents/src/filter_invalid_candidates.rs
```diff
@@ -31,18 +31,29 @@ pub struct CandidateWithStake<TAccountId, TAccountKeys> {
...
 
-#[derive(Clone, Debug, Encode, Decode, PartialEq)]
+#[derive(Clone, Debug, Encode, Decode, PartialEq, Eq)]
 pub struct PermissionedCandidate<TAccountId, TAccountKeys> {
 	pub account_id: TAccountId,
 	pub account_keys: TAccountKeys,
 }
 
-#[derive(Clone, Debug, Encode, Decode, PartialEq)]
+#[derive(Clone, Debug, Encode, Decode, PartialEq, Eq)]
```
Klapeyron, line 40: here is some additional overhead, so from the perspective of our Ord, we will be comparing entire structure (not a account_id only), which may be fine if types will be trivial, but account_keys may be quite big and complex

AmbientTea: it's the other way around, we should order the entries by all the fields, on the off chance that there's entries with the same account id but different session keys (there shouldn't, but we don't want that to possibly interfere with chain's operation). I stick to derived `Ord` as well, even if it's a little heavier.

Klapeyron: ah, it is a tuple, so we are fine

PR #582, opened 2025-03-17
tests: set new chain for ci-preview

rsporny:
fixes:
- ci-preview env was using staging's addresses and was impacting nightly tests; I initialized separate chain for ci

### DIFF
--- a/.github/workflows/staging.yml
+++ b/.github/workflows/staging.yml
@@ -60,7 +60,7 @@ jobs:
     - name: checkout code
       uses: actions/checkout@v4
       with:
-        ref: ${{ github.event_name == 'workflow_dispatch' && github.ref_name || 'master' }}
+        ref: ${{ github.event_name == 'workflow_dispatch' && github.ref_name || 'release-v1.6.0' }}
 
     - name: Set deployment_version as env variable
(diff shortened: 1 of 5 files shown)


.github/workflows/staging.yml
```diff
@@ -60,7 +60,7 @@ jobs:
     - name: checkout code
       uses: actions/checkout@v4
       with:
-        ref: ${{ github.event_name == 'workflow_dispatch' && github.ref_name || 'master' }}
+        ref: ${{ github.event_name == 'workflow_dispatch' && github.ref_name || 'release-v1.6.0' }}
```
kpinter-iohk, line 63: Do I understand correctly that this combined with `no-wipe: true` means that we want to catch breaking node/runtime changes in ci-preview?

rsporny: yes and no: this line only affects staging nightly branch. `no-wipe` will indeed fail job if there will be breaking node/runtime changes, but since it's failing now we don't block anyone. Hopefully, we can support this in CI quickly and be prone to upgrade/runtime issues.

kpinter-iohk (approved):

PR #583, opened 2025-03-18
Build legacy multisig transaction programatically

LGLO:
Modifies test. Transaction for setting the governance mechanism with single key "partner-chains-smart-contracts MultiSig policy" is now created programatically. This will enable using it should the transaction done before have changed.


AmbientTea:
>  still requires updating genesis utxo buried in applied scripts

can't we also programmatically apply the utxo?

LGLO:
> > still requires updating genesis utxo buried in applied scripts
> 
> can't we also programmatically apply the utxo?

We could by copy-pasting some code from `csl` module here or by making some `pub(crate)` functions/structs `pub`. WDYT?

kpinter-iohk:
`csl` (and `plutus_scripts` for that matter) should be in their own crate (crates?) as they have useful things. I'm pretty sure I wanted to use something from them and ended up having to copypaste it. If you look at those modules they are pretty generic and would even make sense outside of PC (they do reference PC in comments or one error message but nothing relevant).

toolkit/offchain/tests/integration_tests.rs
```diff
@@ -210,15 +216,75 @@ async fn update_legacy_governance() {
...
...44424466600200a0080066eb4d5d08011bad357420026eb4d5d09aba200135744002260309201035054310035573c0046aae74004dd50009aba1357440044646464a666ae68cdc3a400000426424460040066eb8d5d08008a999ab9a3370e900100109909118008019bae357420022602e921035054310035573c0046aae74004dd500089809a49035054310035573c0046aae74004dd5000911919192999ab9a3370e90010010a8040a999ab9a3370e90000010980498029aba1001130134901035054310035573c0046aae74004dd5000899800bae75a4464460046eac004c04088cccd55cf800900811919a8081980798031aab9d001300535573c00260086ae8800cd5d0801008909118010018891000980591299a800880511099a8058011802000980511299a800880491099a8050011802000980491299a800880411099a80499a8029a980300111000a4000600800226444a666ae68cdc4000a4000260129210350543600133003001002300822253350011300949103505437002215333573466e1d20000041002133005337020089001000919198021aab9d0013233004200100135573c0026ea80048c88c008004c01c88cccd55cf8009003919a80318021aba10023003357440049311091980080180109100109109119800802001919319ab9c001002120012323001001223300330020020014c012bd8799fd8799f5820f8fbe7316561e57de9ecd1c86ee8f8b512a314ba86499ba9a584bfa8fe2edc8dff00ff004c0129d8799fd87a9f581c7eccc25232d07fdc848d7b653465fc6c23a32d8dfb266181ecd26f0fffd87a80ff0001");
```
gilligan, line 257: Any thoughts on putting blobs like that in external files and including them via `include_str` or `include_bytes` (depending on how the content is stored in the file obviously) ?

Not a major concern but something like the following seems appealing:
```Rust
let applied_version_oracle_script = !include_bytes("./applied_version_oracle_script.bin")
plutus_scripts.add(&PlutusScript::new_v2(applied_version_oracle_script.to_vec()));
```
Maybe I'm ignoring some downsides though? WDYT?

toolkit/offchain/tests/integration_tests.rs
```diff
@@ -210,15 +216,75 @@ async fn update_legacy_governance() {
...
+
+	let genesis_utxo =
```
gilligan, line 220: There is quite a lot going on in this test - I wonder if it would be feasible to maybe extract some pieces and get a bit closer to the common structure of setup / "action" / assertion? 

```
let something = setup_something();
let whatever = build_whatever();

let result = run_upsert_d_param(...)

assert!(result, ...)
```

Might help to make it clearer what is being asserted? WDYT?

LGLO:
> `csl` (and `plutus_scripts` for that matter) should be in their own crate (crates?) as they have useful things. I'm pretty sure it happened more than once that I wanted to use something from them and ended up having to copy-paste it. If you look at those modules they are pretty generic and would even make sense outside of PC (they do reference PC in comments or one error message but nothing relevant).

I don't think so. We take only `offchain` code needs into account there and we also take advantage of having things there usually `pub(crate)`.
Especially `csl` is a set of `offchain` needs regarding mostly CSL and also our Ogmios implementation. I don't think we are in business of providing (and if we are, we lack workforce) general purpose improvements to CSL.

jankun4 (approved):

toolkit/offchain/tests/integration_tests.rs
LGLO: Of course, but this is polishing. The main question for this draft was if it is worth to build this programatically.

toolkit/offchain/tests/integration_tests.rs
LGLO: There is a question if we want to use applied script or we prefer to apply parameters to the base script in a programatic manner. In the latter case we can import it `use raw_scripts::VERSION_ORACLE_POLICY`

toolkit/offchain/tests/integration_tests.rs
gilligan: Indeed, makes perfect sense. Ignore my comment :)

gilligan (approved):

PR #591, opened 2025-03-19
fix: Remove all IP addressing from repo

skylar-simoncelli:
AWS IP addresses are subject to change, we should not be using them anywhere in tests. 

K8s service FQDNs are fixed and reliable

### DIFF
--- a/e2e-tests/config/substrate/ci_nodes.json
+++ b/e2e-tests/config/substrate/ci_nodes.json
@@ -7,8 +7,8 @@
   "nodes_config": {
     "nodes": {
       "validator-1": {
-        "host": "10.0.10.7",
-        "port": "30621",
+        "host": "ci-preview-validator-1-service.ci-preview.svc.cluster.local",
+        "port": "9933",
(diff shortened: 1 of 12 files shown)


kpinter-iohk (approved):
YES!

e2e-tests/config/substrate/ci_nodes.json
```diff
@@ -7,8 +7,8 @@
   "nodes_config": {
     "nodes": {
       "validator-1": {
-        "host": "10.0.10.7",
-        "port": "30621",
+        "host": "ci-preview-validator-1-service.ci-preview.svc.cluster.local",
```
rsporny, line 10: FQDN only works from cluster, right? This config is to run tests from local. When we run tests in CI we use `--ci-run` and a separate config overrides IPs with FQDNs, see `*-ci.json`

rsporny: the other thing is that we're missing `ci-ci.json` config

PR #592, opened 2025-03-19
change: move session-manager under pallet-session-validator-management

AmbientTea:
# Description

Removes the crate `session-manager` and moves its logic into `pallet-session-validator-management` as a submodule behind feature flag `pallet-session-compat`.

# Checklist

- [x] Self-reviewed the diff


toolkit/pallets/session-validator-management/Cargo.toml
```diff
@@ -22,6 +22,10 @@ sp-session-validator-management = { workspace = true }
 sp-io = { workspace = true, optional = true }
 sidechain-domain = { workspace = true }
 
+derive-new = { workspace = true, optional = true }
+pallet-session = { workspace = true, optional = true }
```
LGLO, line 26: Is pallet-session really optional for PC? I don't think you can get rid of it anyway with the code proposed in this PR.

Are you able to get rid of grandpa (and other pallets requiring pallet-session) and build a functional PC runtime?

So, I'm okay with moving the files around but I'm not okay with moving them under feature `pallet-session-compat`

AmbientTea: pallet-session is used with Aura, Grandpa and Babe. In principle, if a chain rolls its own consensus, it's not bound to use pallet-session at all with it. I remember there were mentions of potentially implementing and switching to FastBFT. In that case, there wouldn't really be any reason to keep using pallet-session, because the committee data can be sourced directly from pallet-session-validator-management.

LGLO (approved):

PR #593, opened 2025-03-19
test: remove active and passive flow tests

rsporny:
fixed:
- return types in conftest.py

changed:
- run-e2e-tests action has no default markers, all tests will be run

removed:
- deleted active and passive flow tests along with its corresponding configuration
- eliminated unused methods from the BlockchainApi and SubstrateApi classes, including burn_tokens, get_incoming_txs, and wait_for_next_mc_block


kpinter-iohk (approved):
:tada:

PR #602, opened 2025-03-25
add: OnNewAssociation hook for address-associations pallet

AmbientTea:
# Description

Adds a simple hook to address-associations pallet to inject handler logic for new associations. This is done with a situation in mind, where some block rewards are put aside for beneficiaries whose addresses are not associated and need to be processed in the future when those associations are provided to the chain.

# Checklist

- [ ] Self-reviewed the diff


toolkit/address-associations/pallet/src/lib.rs
```diff
@@ -92,7 +105,10 @@ pub mod pallet {
 
 			ensure!(is_valid_signature, Error::<T>::InvalidMainchainSignature);
 
-			AddressAssociations::<T>::insert(stake_key_hash, partnerchain_address);
+			AddressAssociations::<T>::insert(stake_key_hash.clone(), partnerchain_address.clone());
+
+			T::OnNewAssociation::on_new_association(partnerchain_address, stake_key_hash);
+
```
Klapeyron, line 111: Are we handling the case, when somebody calls this extrinsic with some arbitrary stake_key_hashes that are unique between executions? I am thinking about successful executions, so all signature verifications etc will pass.
Also shall we add some pruning mechanics or deposit requirements for keeping stuff on-chain?

AmbientTea: I don't understand the arbitrary hashes question. If someone can produce a signature, they can save associate their key.

We don't want to prune anything. It might be good to think about some better fee or deposit mechanism to prevent some malicious behaviour, but that's for another PR I guess.

Klapeyron: Yes, so regarding hashes I meant as you wrote, that anybody can store on-chain associations as long as signature is valid. Should I create some issue, so we can address this stuff later on?

Also do we have some docs regarding this pallet functionality? I am trying to understand what kind of issue we are trying to address with it

AmbientTea: We are fine with that. The documentation is in the works. This is to provide mapping from Cardano accounts to Partner Chain accounts.

toolkit/address-associations/pallet/src/lib.rs
```diff
@@ -65,7 +78,7 @@ pub mod pallet {
 	#[pallet::call]
 	impl<T: Config> Pallet<T> {
 		#[pallet::call_index(0)]
-		#[pallet::weight(T::WeightInfo::associate_address())]
+		#[pallet::weight(T::WeightInfo::associate_address() + T::OnNewAssociation::weight())]
```
Klapeyron, line 81: I am not sure if it is necessary, we deliver benchmarks, so in theory end user can execute it with it's own `OnNewAssociation` implementation and use adjusted weights.

AmbientTea: good comment. should we actually expose our own weights in the crates, or would it be better to leave that for PC builders to generate their own?

toolkit/address-associations/pallet/src/lib.rs
```diff
@@ -25,6 +27,15 @@ pub struct AddressAssociationSignedMessage<PartnerChainAddress> {
 	pub genesis_utxo: UtxoId,
 }
 
+pub trait OnNewAssociation<PartnerChainAddress> {
```
Klapeyron, line 30: also can we provide here some default implementation for `()` that does nothing? It will allow end-user to disable it if needed

PR #613, opened 2025-03-27
ETCM-9597 block producer metadata pallet

kpinter-iohk:
# Description

Adds block producer metadata pallet - https://input-output.atlassian.net/browse/ETCM-9597

# Checklist

- [ ] Self-reviewed the diff


toolkit/block-producer-metadata/pallet/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+			let is_valid_signature = signature
+				.verify(&cross_chain_pub_key.into(), &metadata_message.encode())
+				.is_ok();
+
+			ensure!(is_valid_signature, Error::<T>::InvalidMainchainSignature);
+
+			BlockProducerMetadataStorage::<T>::insert(cross_chain_key_hash, metadata);
+			Ok(())
+		}
+	}
```
Klapeyron, line 90: this pallet looks like a copy-paste of pallet-address-associations. What is a purpose of having it?

gilligan: Feel free to read up on the context in the related JIRA ticket

Klapeyron: Right, but it still is the same logic as in pallet-address-associations, underlying metadata is different. 
In associated Jira ticket I found that this registration is for validators, should we verify it here as well?
Also I have similar doubts as with pallet-address-associations, so if we will be pruning at some point this metadata, maybe taking some deposit fee for storing this data on-chain or associating their existence with lifetime of the account (what if it's balance will drop below maintenance level?).

toolkit/block-producer-metadata/primitives/src/lib.rs
```diff
@@ -12,25 +11,18 @@ pub struct MetadataSignedMessage<Metadata> {
 }
 
 impl<M: Encode> MetadataSignedMessage<M> {
```
AmbientTea, line 13: I'd prefer `MetadataSignedMessage` to have both `sign_with_key` and `verify_for_key`. I know it's trivial that we encode the message for signing so we also encode it for verifying the signature, but it would be a little better to have dual functions that encapsulate that.

toolkit/block-producer-metadata/primitives/src/lib.rs
```diff
@@ -12,25 +11,18 @@ pub struct MetadataSignedMessage<Metadata> {
...
-		signature: CrossChainSignature,
-	) -> Result<(), secp256k1::Error> {
-		let data = self.encode();
-		let data_hash = Message::from_hashed_data::<sha256::Hash>(&data);
-
-		let signature = secp256k1::ecdsa::Signature::from_der(&signature.0)
-			.or_else(|_| secp256k1::ecdsa::Signature::from_compact(&signature.0))
-			.expect("ecdsa::Signature from CrossChainSignature should always succeed");
+		let digest =
+			<k256::Secp256k1 as k256::ecdsa::hazmat::DigestPrimitive>::Digest::new_with_prefix(
```
AmbientTea, line 20: I'd `use` that to shorted the line

toolkit/cli/commands/Cargo.toml
```diff
@@ -21,10 +21,12 @@ hex-literal = { workspace = true }
 plutus = { workspace = true, default-features = true }
 plutus-datum-derive = { workspace = true, default-features = true }
 secp256k1 = { workspace = true, features = ["std", "global-context"] }
+k256 = { workspace = true, features = ["serde"] }
 sidechain-domain = { workspace = true, default-features = true, features = [
 	"std",
 ] }
 sp-io = { workspace = true, default-features = true }
+# sp-core = { workspace = true, default-features = true }
```
AmbientTea, line 29: to remove if not needed

toolkit/sidechain/domain/src/lib.rs
```diff
@@ -471,10 +471,72 @@ pub struct SidechainSignature(pub Vec<u8>);
...
+#[cfg_attr(feature = "std", byte_string(to_hex_string, decode_hex))]
+#[cfg_attr(feature = "serde", byte_string(hex_serialize, hex_deserialize))]
+pub struct CrossChainKeyHash(pub [u8; CROSS_CHAIN_KEY_HASH_LEN]);
+
+impl Display for CrossChainKeyHash {
+	fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
+		let hash = sp_core::hexdisplay::HexDisplay::from(&self.0);
+		write!(f, "0x{}", hash)
+	}
+}
```
AmbientTea, line 520:
```suggestion
#[cfg_attr(feature = "serde", byte_string(hex_serialize, hex_deserialize))]
#[byte_string(to_hex_string)]
pub struct CrossChainKeyHash(pub [u8; CROSS_CHAIN_KEY_HASH_LEN]);

impl Display for CrossChainKeyHash {
	fn fmt(&self, f: &mut Formatter<'_>) -> core::fmt::Result {
	        f.write_str(hash.to_hex_string())
	}
}
```
This should work

AmbientTea (approved):

toolkit/block-producer-metadata/pallet/src/lib.rs
chrispalaskas: Lets address this concern. Can anyone call this function upsert_metadata or only validators?

chrispalaskas: I don't understand the rush to merge, when there is a valid concern. If anyone can call this function, they can spam your storage for a very small fee, just paying tx fee and having a huge metadata to store. Is that thinking right? If yes, isn't it a valid concern that needs addressed?

kpinter-iohk: These concerns can be addressed in a follow up PR.

chrispalaskas: https://input-output.atlassian.net/wiki/spaces/SID/pages/4932042804/Partner+Chains+Risk+Assessment
Captured here by @AmbientTea

PR #616, opened 2025-03-27
ETCM-9598 block producer metadata rpc

kpinter-iohk:
# Description

RPC for block producer metadata pallet

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/toolkit/block-producer-metadata/primitives/src/lib.rs
+++ b/toolkit/block-producer-metadata/primitives/src/lib.rs
@@ -1,7 +1,9 @@
 #![cfg_attr(not(feature = "std"), no_std)]
 
-use parity_scale_codec::Encode;
+use parity_scale_codec::{Decode, Encode};
 use sidechain_domain::*;
+use sp_api;
+extern crate alloc;
 
 #[derive(Debug, Clone, Encode)]
 pub struct MetadataSignedMessage<Metadata> {
@@ -56,5 +58,16 @@ mod tests {
 		let signature = message.sign_with_key(&skey);
 
 		assert!(message.verify_signature(&vkey.into(), signature).is_ok());
+	}
+}
+
+sp_api::decl_runtime_apis! {
+	pub trait BlockProducerMetadataApi<Metadata>
+	where Metadata:Decode
+	{
+		/// Retrieves the metadata for a given SPO public key if it exists.
+		fn get_metadata_for(
+			cross_chain_pub_key: &CrossChainPublicKey,
+	) -> Option<Metadata>;
 	}
 }
--- a/toolkit/block-producer-metadata/primitives/src/lib.rs
+++ b/toolkit/block-producer-metadata/primitives/src/lib.rs
@@ -68,6 +68,6 @@ sp_api::decl_runtime_apis! {
 		/// Retrieves the metadata for a given SPO public key if it exists.
 		fn get_metadata_for(
 			cross_chain_pub_key: &CrossChainPublicKey,
-	) -> Option<Metadata>;
+		) -> Option<Metadata>;
 	}
 }
--- a/toolkit/block-producer-metadata/rpc/Cargo.toml
+++ b/toolkit/block-producer-metadata/rpc/Cargo.toml
@@ -1,8 +1,14 @@
 [package]
 name = "pallet-block-producer-metadata-rpc"
-version = "1.6.0"
-edition = "2021"
+version.workspace = true
 license = "Apache-2.0"
+authors.workspace = true
+edition.workspace = true
+homepage.workspace = true
+repository.workspace = true
+
+[lints]
+workspace = true
 
 [dependencies]
 parity-scale-codec = { workspace = true, features = ['std'] }
--- a/toolkit/block-producer-metadata/rpc/Cargo.toml
+++ b/toolkit/block-producer-metadata/rpc/Cargo.toml
@@ -1,0 +1,28 @@
+[package]
+name = "pallet-block-producer-metadata-rpc"
+version = "1.6.0"
+edition = "2021"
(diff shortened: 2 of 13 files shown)


toolkit/sidechain/domain/src/byte_string.rs
```diff
@@ -78,6 +79,17 @@ impl<'a, const N: u32> Deserialize<'a> for BoundedString<N> {
...
+impl<const N: u32> Serialize for BoundedString<N> {
+	fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+	where
+		S: serde::Serializer,
+	{
+		let str = alloc::string::String::from_utf8(self.0.to_vec())
+			.map_err(|_| S::Error::custom("String is not valid UTF-8"))?;
+		serializer.serialize_str(&str)
+	}
+}
```
Klapeyron, line 91: probably we do not need to implement it manually, why we do not want to derive it?

Klapeyron (changes requested):

toolkit/sidechain/domain/src/byte_string.rs
AmbientTea: This is a byte-encoded string. Derived serialize will encode it as an array of number, which I don't think is what we want.

Also, please don't use "request changes" unless it's something that really shouldn't make it to master. All suggestions are welcome, but nits and questions shouldn't block merging and not every PR author feels comfortable dismissing reviews. We're fine being 90% perfect.

Klapeyron: there are some [crates](https://crates.io/crates/serde_with) that allow to adjust this behavior without manual implementation, in general it is not a standard to write serialize/deserialize traits manually. Also [serde](https://serde.rs/field-attrs.html) has dedicated macros to do some conversions just before serialization

toolkit/block-producer-metadata/primitives/src/lib.rs
```diff
@@ -58,3 +60,14 @@ mod tests {
...
+
+sp_api::decl_runtime_apis! {
+	pub trait BlockProducerMetadataApi<Metadata>
+	where Metadata:Decode
+	{
+		/// Retrieves the metadata for a given SPO public key if it exists.
+		fn get_metadata_for(
+			cross_chain_pub_key: &CrossChainPublicKey,
+	) -> Option<Metadata>;
+	}
```
AmbientTea, line 72: something wrong with formatting

toolkit/block-producer-metadata/rpc/Cargo.toml
AmbientTea, line 1: I have nothing against grouping dependencies like that, but we should probably be consistent first

toolkit/block-producer-metadata/rpc/src/lib.rs
```diff
@@ -0,0 +1,56 @@
...
+		&self,
+		cross_chain_public_key: CrossChainPublicKey,
+	) -> RpcResult<Option<Metadata>>;
+}
+
+#[derive(new)]
+pub struct BlockProducerMetadataRpc<C, Block, Metadata> {
+	client: Arc<C>,
+	_marker: std::marker::PhantomData<Block>,
+	_marker1: std::marker::PhantomData<Metadata>,
```
AmbientTea, line 28:
```suggestion
	_phantom: std::marker::PhantomData<(Block, Metadata)>,
```

AmbientTea (approved):

toolkit/block-producer-metadata/primitives/src/lib.rs
kpinter-iohk: Thank you, I assume cargo fmt didn't catch this because it's inside a macro.

toolkit/block-producer-metadata/rpc/Cargo.toml
kpinter-iohk: [It's consistent with our inconsistency.](https://github.com/input-output-hk/partner-chains/blob/master/toolkit/sidechain/rpc/Cargo.toml)
I will fix both.

toolkit/block-producer-metadata/rpc/Cargo.toml
```diff
@@ -0,0 +1,26 @@
+[package]
+name = "pallet-block-producer-metadata-rpc"
+version = "1.6.0"
+edition = "2021"
+license = "Apache-2.0"
+
```
Klapeyron, line 6: please unify it with rest of the .toml files

toolkit/sidechain/domain/src/byte_string.rs
kpinter-iohk: I tried `serde_with` but it was not cooperating. I don't want to go down this rabbit hole right now so I created a ticket for this.

Klapeyron: my additional thought there was also like if we want to store under the hood UTF-8 based array, then maybe we should store String instead of BoundedVec and verify it's lenght in TryFrom. But we can address that later ofc

kpinter-iohk: We use BoundedVec because it has a MaxEncodedLen instance. AFAIK String doesn't (can't?) have one. From what I've found Substrate recommends BoundedVec<u8> for storing Strings on the chain.

Klapeyron: frontier uses something similar
https://github.com/polkadot-evm/frontier/blob/9377029ff4a4182c7f6a7eaf632f0fcb135a7f3f/precompiles/src/solidity/codec/bytes.rs#L79
`pub type BoundedString<S> = BoundedBytesString<StringKind, S>;`

PR #617, opened 2025-03-27
local-env-governance-utxo-with-plutus-script

rsporny:
added:
- governance authority in local env has plutus script attached to mimic real life scenarios

Refs: ETCM-9630

### DIFF
--- a/e2e-tests/config/substrate/local_nodes.json
+++ b/e2e-tests/config/substrate/local_nodes.json
@@ -1,6 +1,6 @@
 {
   "deployment_mc_epoch": 2,
-  "genesis_utxo": "f8fbe7316561e57de9ecd1c86ee8f8b512a314ba86499ba9a584bfa8fe2edc8d#0",
+  "genesis_utxo": "83d513c0571bb0ad2dce3ec7cb1e40966e8a9e0b1610a57021932b98683582c1#0",
   "main_chain": {
     "network": "--testnet-magic 42",
     "epoch_length": 120,
--- a/dev/local-environment/configurations/cardano/entrypoint.sh
+++ b/dev/local-environment/configurations/cardano/entrypoint.sh
@@ -121,6 +121,7 @@ cardano-cli latest transaction build-raw \
   --tx-out "$dave_address+$tx_out3" \
   --tx-out "$eve_address+$tx_out4" \
   --tx-out "$new_address+$change" \
+  --tx-out-reference-script-file /shared/v-function.script \
   --tx-out "$new_address+$tx_out5_lovelace+$tx_out5_reward_token" \
   --tx-out "$vfunction_address+$tx_out6" \
   --tx-out-reference-script-file /shared/v-function.script \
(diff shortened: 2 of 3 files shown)


LGLO:
I think it should be a test case in e2e-tests:
* given an utxo with script attached
* when governance init is run for such case
* then init completes successfully

How many test scenarios are we going to squeeze into local-env setup?

rsporny:
@LGLO you've already covered it in unit tests, so it won't even get pass through `cargo check`. What you're proposing is not e2e test, because it would only test command's response and/or main chain state. There is no e2e test for `governance init` because chain builder is interested in starting a chain, `governance init` is only one of the needed steps to achieve this.

What you're asking for is an integration level coverage which we can address separately. For now, it's not a bad idea to mimic real-life scenarios while we set up local-env. In case of failure the reason may not be clear, but it's a clear win-win solution, because we'll catch this bug (and probably others too) in CI.

LGLO:
> Current e2e test coverage is implicitly testing it in local-env entrypoint.
Yes, and we have obviously 1 entry point that calls `governance init`, `upsert-d-parameter`, `upsert-permissioned-candidates` all of them with single key governance.

`governance init` is "implicitly tested" with a UTXO that has no script attached.
This pull request changes this to implicitly testing `governance init` with UTXO that has Plutus Script attached.

The bad news is that this won't cover scenario where UTXO has Native Script attached. I can assure you that we have a bug with `governance init` using genesis utxo with a native script attached. So, how are you going to test both case with this approach?

If e2e-tests are going to cover multiple scenarios for `governance init`: no script, PlutusScript, NativeScript, multi-assets present or not, then `governance init` should be called from e2e-tests.
If e2e-tests are only interested in one of the choices, let it be the simplest choice and then we can either settle on adding unit tests (asserting of lack of failures, or having fee over some value) or we can add Rust integration tests that assert us that `governance init` transaction is successful.

rsporny:
@LGLO The approach to test such cases is to use integration tests. e2e-tests are not going to cover all of the above, but as I said, there is nothing wrong with enhancing initial setup with additional data, because the goal is to catch bugs. It serves now as a second safety net. Are you suggesting that we should not increase our coverage now (therefore increase the risk of unnoticed bugs) and wait until we have integration tests, because you don't like the fact that we're testing `governance init` indirectly?

LGLO:
I'm saying local-env is not a place to make changes when we need another test scenario. We can merge this change now. The next day we'll find out that it does not cover ETCM-9632 and we do another update? And as I said before, this way we can only test one case. I do not block this PR and I don't want to be complicit in going in wrong direction.

e2e-tests/config/substrate/local_nodes.json
```diff
@@ -1,6 +1,6 @@
 {
   "deployment_mc_epoch": 2,
-  "genesis_utxo": "f8fbe7316561e57de9ecd1c86ee8f8b512a314ba86499ba9a584bfa8fe2edc8d#0",
+  "genesis_utxo": "83d513c0571bb0ad2dce3ec7cb1e40966e8a9e0b1610a57021932b98683582c1#0",
```
LGLO, line 3: This UTOX does not have script attached.

AmbientTea:
Closing due to no activity

PR #622, opened 2025-03-31
BoundedString to BoundedVec

Klapeyron:
# Description

Proposition of small refactoring

# Checklist

- [ ] Self-reviewed the diff


node/runtime/src/lib.rs
```diff
@@ -461,13 +461,15 @@ impl AsCardanoSPO for BlockAuthor {
 	}
 }
 
-pub const MAX_METADATA_URL_LENGTH: u32 = 512;
+parameter_types! {
+	pub const MaxMetadataUrlLength: u32 = 512;
```
Klapeyron, line 465: looks quite long, should we short it?

gilligan:
> Proposition of small refactoring

Could you add a bit context in the description like the rationale/benefit of the refactoring? I might learn something :)
Using BoundedVec means various traits already exist and don't have to be manually generated right? Doesn't that also imply changes on the call-site? (I guess that's what Lech's comment might also be hinting at)

Klapeyron:
I put some details regarding changes in this PR in associated Jira item

node/runtime/src/lib.rs
```diff
@@ -461,13 +461,15 @@ impl AsCardanoSPO for BlockAuthor {
...
+}
 
 #[derive(
 	Clone, Debug, MaxEncodedLen, Encode, Decode, Serialize, Deserialize, PartialEq, Eq, TypeInfo,
 )]
 pub struct BlockProducerMetadataType {
-	pub url: BoundedString<MAX_METADATA_URL_LENGTH>,
```
AmbientTea, line 470: This is our own reference runtime implementation, not some code that anyone other than us is forced to use. We can make any assumptions that are convenient for us to do here. I went with a string-like type, because that's easy and intuitive for us to work with.

AmbientTea (commented):
I don't think there is a problem to be solved here

PR #624, opened 2025-03-31
chore: add GH action to run e2e-tests on ci-preview

rsporny:
added:
- env dropdown in GH action to run tests against specific environment


kpinter-iohk (approved):
:tada:

PR #629, opened 2025-04-01
fix: Release block participation based on block not slot number

AmbientTea:
# Description

This should make it much easier to test the feature. Now the block participation inherent should always release data for 30 (defaul) blocks at a time.

# Checklist

- [ ] Self-reviewed the diff


LGLO:
I don't see how it makes it any simpler.  Still the output data will span multiple epochs and it is okay. I would refrain from making this unless it is proven it fixes a bug.

PR #642, opened 2025-04-03
ETCM-9557-Implement direct tests

skylar-simoncelli:


e2e-tests/src/partner_chain_rpc.py
```diff
@@ -47,51 +47,51 @@ def __get_body(self, jsonrpc="2.0", method="", params=[], id=1):
...
+        
+        # Use kubectl exec method
+        pod = os.environ["KUBECTL_EXEC_POD"]
+        namespace = os.environ.get("K8S_NAMESPACE", "default")
+        
+        payload = json.dumps(self.__get_body(method=method, params=params))
+        cmd = [
+            "kubectl", "exec", pod, "-n", namespace, "--",
+            "curl", "-s", "-H", "Content-Type: application/json",
+            "-d", payload, "http://localhost:9933"
```
rsporny, line 76:
```suggestion
            "-d", payload, self.url
```
I would change that. Users can select which node they want to connect to/test in config:
```
"selected_node": "validator-4",
```
and in future we want to test random nodes, so I’d like to keep it that way. It should be as easy as updating hosts in `<env>-ci.json`. I assume fqdn that we can only use within aws network are no longer be needed since we're switching to kubectl.

.github/actions/tests/k8s-e2e-tests/action.yml
```diff
@@ -73,6 +73,27 @@ runs:
         kubectl config use-context my-context
       shell: bash
 
+    - name: Copy all decrypted keys from binary-host pod
```
rsporny, line 76: We want to remove `binary-host` completely, and we need secrets to be available by `partner-chains-node` binary so we need copy them into the pod directly. The best way to do that is leveraging existing pytest fixtures, since they allow us to clean them up after tests are done.

AmbientTea:
Closing as inactive

PR #638, opened 2025-04-02
ETCM-6659 allow unset permissioned candidates if P=0

kpinter-iohk:
# Description

When getting Ariadne parameters, if `num_permissioned_candidates` in d_params is 0, having not inserted a permissioned candidates list is not an error. If a chain does not wish to have any permissioned candidates, this way they won't need to spend ADA submitting an empty list of them.

# Checklist

- [ ] Commit sequence broadly makes sense and commits have useful messages.
- [ ] The size limit of 400 LOC isn't needlessly exceeded
- [ ] The PR refers to a JIRA ticket (if one exists)
- [ ] New tests are added if needed and existing tests are updated.
- [ ] Relevant logging and metrics added
- [ ] Any changes are noted in the `changelog.md` for affected crate
- [ ] Self-reviewed the diff

PR #643, opened 2025-04-03
ETCM-6659 allow unset permissioned candidates if P=0

kpinter-iohk:
# Description

:rotating_light: This PR is similar to #638 but it moves the logic out of the data source layer.

When getting Ariadne parameters, if num_permissioned_candidates in d_params is 0, having not inserted a permissioned candidates list is not an error. If a chain does not wish to have any permissioned candidates, this way they won't need to spend ADA submitting an empty list of them.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/toolkit/committee-selection/authority-selection-inherents/src/ariadne_inherent_data_provider.rs
+++ b/toolkit/committee-selection/authority-selection-inherents/src/ariadne_inherent_data_provider.rs
@@ -207,27 +207,11 @@ mod tests {
 	async fn return_empty_ariadne_cidp_if_runtime_requests_too_new_epoch() {
 		// This is the epoch number that is too new
 		let next_unset_epoch_number = ScEpochNumber(42);
-		let client = TestApi { next_unset_epoch_number };
-		let sc_slot_config = ScSlotConfig {
-			slots_per_epoch: SlotsPerEpoch(10),
-			slot_duration: SlotDuration::from_millis(1000),
-		};
-		let epoch_config = MainchainEpochConfig {
-			first_epoch_timestamp_millis: Timestamp::from_unix_millis(0),
-			first_epoch_number: 0,
-			epoch_duration_millis: Duration::from_millis(
-				u64::from(sc_slot_config.slots_per_epoch.0)
-					* sc_slot_config.slot_duration.as_millis()
-					* 10,
-			),
-			first_slot_number: 0,
-			slot_duration_millis: Duration::from_millis(1000),
-		};
 		let mc_reference_epoch = McEpochNumber(1);
 		let empty_ariadne_idp = AriadneInherentDataProvider::new(
-			&client,
-			&sc_slot_config,
-			&epoch_config,
+			&client(next_unset_epoch_number),
+			&sc_slot_config(),
+			&epoch_config(),
 			H256::zero(),
 			// This is the slot that will be used to calculate current_epoch_number
 			Slot::from(400u64),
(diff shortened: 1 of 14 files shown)


toolkit/committee-selection/authority-selection-inherents/src/tests.rs
```diff
@@ -347,6 +347,57 @@ fn ariadne_does_not_return_empty_committee() {
...
+	assert!(calculated_committee.is_some());
+
+	let committee = calculated_committee.unwrap();
+	let committee_names = committee
+		.iter()
+		.map(|candidate| account_id_to_name(candidate.account_id()))
+		.collect::<Vec<_>>();
+	let expected_committee_names = vec!["greg", "kim", "henry", "ferdie", "ida", "henry", "greg"];
+
+	assert_eq!(committee_names, expected_committee_names);
```
LGLO, line 398: I know it looks like other tests in this file, but after having some reflection, I think it would be better to replace this exact expected with something that checks property, like that each of candidate in committee_names comes from registered validators (IOW they are not made up) and that length is equal to R.

toolkit/committee-selection/authority-selection-inherents/src/authority_selection_inputs.rs
```diff
@@ -8,7 +8,7 @@ use sidechain_domain::*;
 #[derive(Clone, Debug, Encode, Decode, TypeInfo, PartialEq, Eq)]
 pub struct AuthoritySelectionInputs {
 	pub d_parameter: DParameter,
-	pub permissioned_candidates: Vec<PermissionedCandidateData>,
+	pub permissioned_candidates: Option<Vec<PermissionedCandidateData>>,
```
LGLO, line 11: At his point this PR looks like a breaking change to me. It changes type used in the runtime.

Users do runtime update first: IDPs give old data format, but runtime expects new - error.
Users upgrade their nodes first: IDPs give new data format, but runtime expects old - error.

LGLO:
I'm afraid that #638 is better, because it makes the change local to data source and doesn't break things. All the knowledge needed for the case we want to support in this PR description is available in data source.

toolkit/committee-selection/authority-selection-inherents/src/tests.rs
kpinter-iohk: This test has been removed, please let me know if you think the new tests would work better as property based.

toolkit/committee-selection/authority-selection-inherents/src/authority_selection_inputs.rs
kpinter-iohk: This has been changed to hopefully avoid making this  a breaking change.

toolkit/data-sources/db-sync/src/candidates/tests.rs
```diff
@@ -89,13 +89,31 @@ async fn test_get_ariadne_parameters_returns_the_latest_candidates_if_there_were
...
+		.get_ariadne_parameters(
+			McEpochNumber(191),
+			d_parameter_policy(),
+			permissioned_candidates_policy(),
+		)
+		.await
+		.unwrap();
+	assert_eq!(
+		result.d_parameter,
+		DParameter { num_permissioned_candidates: 1, num_registered_candidates: 3 }
```
LGLO, line 112:
```
        INSERT INTO datum ( id   , hash        , tx_id , value                                         )
        VALUES            ( 6000 , datum1_hash , tx1_id, '{"list": [{"int": 1}, {"int": 2}]}' 
```

toolkit/data-sources/db-sync/testdata/migrations/6_insert_transactions.sql
```diff
@@ -180,7 +180,7 @@ DO $$
         d_parameter_policy hash28type := decode('500000000000000000000000000000000000434845434b504f494e69', 'hex');
     BEGIN
         INSERT INTO tx ( id     , hash     , block_id , block_index , out_sum , fee , deposit , size , invalid_before , invalid_hereafter , valid_contract , script_size )
-        VALUES         ( tx1_id , tx1_hash , 3        , 0           , 0       , 0   , 0       , 1024 , NULL           , NULL              , TRUE           , 1024        )
```
LGLO, line 183: Is it a required change? Is moving "insert D-param" transaction away from "insert permissioned candidates" to have epoch 191 observing the former but not the latter?

kpinter-iohk: That was the intention.

toolkit/committee-selection/authority-selection-inherents/src/ariadne_inherent_data_provider.rs
```diff
@@ -239,4 +239,86 @@ mod tests {
...
+			first_epoch_timestamp_millis: Timestamp::from_unix_millis(0),
+			first_epoch_number: 0,
+			epoch_duration_millis: Duration::from_millis(
+				u64::from(sc_slot_config.slots_per_epoch.0)
+					* sc_slot_config.slot_duration.as_millis()
+					* 10,
+			),
+			first_slot_number: 0,
+			slot_duration_millis: Duration::from_millis(1000),
+		};
```
LGLO, line 305: Cosmetic: I think 3x same config could be factored out.

LGLO (approved):

PR #654, opened 2025-04-09
chore: change edition to 2024

LGLO:
# Description

Changes Rust edition to `"2024"`. It mostly changes rules of ordering `use` and formatting.
This PR doesn't add any non-default in this regard.

# Checklist

- [x] Self-reviewed the diff


AmbientTea (commented):
there's a lot of lines being broken up in this PR. maybe we should set a higher line length for cargo fmt

LGLO:
It adds approx 300 lines. Making max line length 120 would decrease number of lines by approx 3800 lines. I don't know if  I want to mix these two changes. I was rather going to play with some formatting rules later. The question here is if we want to use edition 2024 or we stick to 2021.

gilligan:
> The question here is if we want to use edition 2024 or we stick to 2021.

I guess there is no urgent need to update to 2024 but that's also a good reason to just do so anyway because we don't want to wait for having to do so urgently 👍

kpinter-iohk:
> I was rather going to play with some formatting rules later.

I think this is the case where it would be better if this was done in one PR. Reformatting many lines and then reformatting them again (potentially many to their original state) pollutes git history, and there are 2 large PRs to review.

kpinter-iohk:
I see that all `if let` expressions were converted to `match`, because of this: https://doc.rust-lang.org/edition-guide/rust-2024/temporary-if-let-scope.html
I would revert these as I would hazard a guess that 100% of our `if let`s are fine in 2024 edition, and I find `if let` less noisy than `match` (one less indentation level and block, no unnecessary match arm).

kpinter-iohk:
There are lots of `expr_2021`s in macros, have these been checked if they work with `expr`? ([Example](https://github.com/input-output-hk/partner-chains/pull/654/files#diff-3fd8bf68e003c766606f2bebde848d4e8574beb75fa08f04eb0bae7afe4906e9R663))
See https://doc.rust-lang.org/edition-guide/rust-2024/macro-fragment-specifiers.html

AmbientTea (approved):

PR #655, opened 2025-04-09
add: ETCM-9700 Governed Map inherent data provider

AmbientTea:
# Description

__WIP___

# Checklist

- [ ] Self-reviewed the diff


toolkit/governed-map/primitives/src/lib.rs
```diff
@@ -50,3 +69,166 @@ impl IsFatalError for InherentError {
...
+
+#[cfg(feature = "std")]
+#[async_trait::async_trait]
+impl sp_inherents::InherentDataProvider for GovernedMapInherentDataProvider {
+	async fn provide_inherent_data(
+		&self,
+		inherent_data: &mut sp_inherents::InherentData,
+	) -> Result<(), sp_inherents::Error> {
+		match self {
+			Self::ActiveV1 { changes } if !changes.is_empty() => {
```
AmbientTea, line 93: Should I log some warning that an active IDP was created with empty change list? This should not happen in theory

kpinter-iohk: Could we use a non-empty vec type for the changes?

AmbientTea: is there one?

toolkit/governed-map/primitives/src/lib.rs
```diff
@@ -50,3 +69,167 @@ impl IsFatalError for InherentError {
...
+	}
+}
+
+sp_api::decl_runtime_apis! {
+	/// Runtime API exposing data required for the [GovernedMapInherentDataProvider] to operate.
+	#[api_version(1)]
+	pub trait GovernedMapApi
+	{
+		/// Returns all key-value mappings currently stored in the pallet
+		fn get_stored_mappings() -> BTreeMap<String, ByteString>;
```
LGLO, line 229: Should we have "get by key" as well or am I making up requirements?

toolkit/governed-map/primitives/src/lib.rs
LGLO: Or make your life simpler by removing `Inert` and corresponding `if-else` in `new_v1`. `Inert` should equal to `changes.is_empty()`.

toolkit/governed-map/primitives/src/lib.rs
LGLO: `HashMap` cannot be used because of runtime not having it available?

toolkit/governed-map/primitives/src/lib.rs
```diff
@@ -50,3 +69,167 @@ impl IsFatalError for InherentError {
...
+		let entries_in_storage = api.get_stored_mappings(parent_hash)?;
+		let current_entries = data_source
+			.get_current_mappings(mc_hash, main_chain_script)
+			.await
+			.map_err(InherentProviderCreationError::DataSourceError)?;
+
+		let mut changes: ChangesV1 = ChangesV1::new();
+
+		for (key, value) in current_entries.iter() {
+			if !entries_in_storage.contains_key(key) || entries_in_storage[key] != *value {
```
LGLO, line 202:
```suggestion
			if entries_in_storage.get(key) != Some(value) {
```
to avoid double lookup for the same key

LGLO (approved):

toolkit/governed-map/primitives/src/lib.rs
AmbientTea: You're right that I can remove the if-else in `new_v1`, but I think keeping `Inert` is still a principled thing to do here, so `new` has a version-neutral variant to return and so does `new_v1` when the addresses are unset

toolkit/governed-map/primitives/src/lib.rs
AmbientTea: Yes, `HashMap` is unavailable in no-std context. We don't need get-by-key for the IDP. We'll probably add it when we add an RPC endpoint, but I think the RPC should then have its own runtime API definition anyway

toolkit/governed-map/primitives/src/lib.rs
AmbientTea: Good idea, thanks

toolkit/governed-map/primitives/src/lib.rs
AmbientTea: I removed the if-else and let the IDP have empty vector, but left `Inert` for situations where the pallet or addresses are not set in the runtime yet

PR #663, opened 2025-04-10
ETCM-9703 governed map data source

kpinter-iohk:
# Description

This PR adds data source and query for governed map

# Checklist

- [ ] Commit sequence broadly makes sense and commits have useful messages.
- [ ] The size limit of 400 LOC isn't needlessly exceeded
- [ ] The PR refers to a JIRA ticket (if one exists)
- [ ] New tests are added if needed and existing tests are updated.
- [ ] Relevant logging and metrics added
- [ ] Any changes are noted in the `changelog.md` for affected crate
- [ ] Self-reviewed the diff

PR #671, opened 2025-04-14
add: ETCM-9713 Governed Map pallet on-change hooks

AmbientTea:
# Description

- adds `OnGovernedMapChange` trait and uses it as a hook in the pallet
- adds a mock data source (shamelessly stolen from https://github.com/input-output-hk/partner-chains/pull/663 )
- wires the IDP and data source (just mock) into the demo node

# Checklist

- [x] Self-reviewed the diff


demo/runtime/src/test_helper_pallet.rs
```diff
@@ -6,11 +6,15 @@ pub use pallet::*;
 pub mod pallet {
 	use crate::AccountId;
 	use crate::BlockAuthor;
+	use crate::MaxKeyLength;
+	use crate::MaxValueLength;
```
LGLO, line 10: Cosmetic: group them please

demo/runtime/src/lib.rs
```diff
@@ -1009,6 +1022,20 @@ impl_runtime_apis! {
 			<Runtime as pallet_block_participation::Config>::TARGET_INHERENT_ID
 		}
 	}
+
+	impl sp_governed_map::GovernedMapApi<Block> for Runtime {
+		fn get_stored_mappings() -> BTreeMap<String, ByteString> {
```
LGLO, line 1027: I'm wondering why we decided to lose the information in this public API. We could keep size constraints and let consumers to drop them. We assumed that users won't use it.

demo/runtime/src/test_helper_pallet.rs
```diff
@@ -88,6 +92,38 @@ pub mod pallet {
...
+						to_hex(&new_value, false)
+					)
+				},
+
+				(None, Some(old_value)) => {
+					log::info!(
+						"Governed Map value for key '{key}' deleted, old value: {}",
+						to_hex(&old_value, false)
+					)
+				},
```
LGLO, line 121: Are logs enough for testing this feature? I think it could write to unbounded map storage with block hash as key and value being a Vec<(key, new_value, old_value)> and let e2e-tests assert on storage values.

toolkit/governed-map/pallet/src/lib.rs
```diff
@@ -177,7 +187,9 @@ pub mod pallet {
 			log::info!("💾 Registering {} Governed Map changes", changes.len(),);
 
 			for (key, value) in changes {
-				Mapping::<T>::set(key, value);
+				let old_value = Mapping::<T>::get(&key);
```
LGLO, line 190: Run bechmarking, please

toolkit/governed-map/pallet/src/lib.rs
```diff
@@ -214,5 +226,10 @@ pub mod pallet {
 		pub fn get_version() -> u32 {
 			PALLET_VERSION
 		}
+
+		///
```
kpinter-iohk, line 230: missing comment

demo/runtime/src/lib.rs
AmbientTea: We are the users. I can rename this trait to reflect that, but this trait exists only for the IDP to use and the IDP doesn't know anything about those constraints

demo/runtime/src/test_helper_pallet.rs
AmbientTea: The pallet is already saving the current state of the mapping in the storage and that can be queried for tests

LGLO: Right, should be enough if we run archive nodes.

LGLO (approved):

PR #672, opened 2025-04-14
test: add ariadne v2 tests

rsporny:
added:
- guaranteed seats tests
- seats distribution matching DParam tests

changed:
- participation tolerance will use absolute value if calculated one is too small

Refs: ETCM-8934

### DIFF
--- a/e2e-tests/tests/committee/conftest.py
+++ b/e2e-tests/tests/committee/conftest.py
@@ -16,7 +16,6 @@ from src.run_command import RunnerFactory
 
 block_finder: BlockFinder = None
 
-PERMISSIONED_CANDIDATE_WEIGHT = 1
 CANDIDATES_STABILITY_OFFSET_IN_MC_EPOCHS = 2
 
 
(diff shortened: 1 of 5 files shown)


e2e-tests/tests/committee/conftest.py
```diff
@@ -441,14 +442,21 @@ def _inner(mc_epoch):
                 probability_of_selection_in_single_epoch = (permissioned_seats / total_committee_seats) * (
                     PERMISSIONED_CANDIDATE_WEIGHT / permissioned_candidates_count
                 )
+                expected_attendance = permissioned_seats * (
+                    PERMISSIONED_CANDIDATE_WEIGHT / permissioned_candidates_count
```
LGLO, line 446: IMO `PERMISSIONED_CANDIDATE_WEIGHT` should just disappear. It is one, the neutral element of multiplication used in formulas that are only multiplications and divisions.

What is more:
```
probability_of_selection_in_single_epoch = (permissioned_seats / total_committee_seats) * (
                    PERMISSIONED_CANDIDATE_WEIGHT / permissioned_candidates_count
                )
```
is now invalid. If P=1, R=2, and there is 1 permissioned candidate, his probability is 1, but according to this formula it would be `(1/3) * (1/1)`, which is `1/3`.

e2e-tests/tests/committee/conftest.py
```diff
@@ -441,14 +442,21 @@ def _inner(mc_epoch):
...
+                )
+                guaranteed_seats = math.floor(expected_attendance)
             else:
                 probability_of_selection_in_single_epoch = (trustless_seats / total_committee_seats) * (
                     candidate.stake_delegation / total_stake
                 )
+                expected_attendance = trustless_seats * (candidate.stake_delegation / total_stake)
+                guaranteed_seats = math.floor(expected_attendance)
             expected_attendance_in_single_epoch = total_committee_seats * probability_of_selection_in_single_epoch
```
LGLO, line 455: I think this should be removed

e2e-tests/tests/committee/conftest.py
```diff
@@ -441,14 +442,21 @@ def _inner(mc_epoch):
...
             total_expected_attendance = epochs_num * expected_attendance_in_single_epoch
```
LGLO, line 456:
```suggestion
            total_expected_attendance = epochs_num * expected_attendance
```

e2e-tests/tests/committee/conftest.py
```diff
@@ -441,14 +442,21 @@ def _inner(mc_epoch):
...
             candidate.probability = probability_of_selection_in_single_epoch
```
LGLO, line 457: If you really need `.probability`, then I think it is `min(1, expected_attendance)`

e2e-tests/tests/committee/test_committee.py
```diff
@@ -3,37 +3,80 @@
...
+    def p_candidates_count(
+        self,
+        current_mc_epoch,
+        get_pc_epoch_committee,
+        p_candidates_available,
+        pc_epoch_calculator: PartnerChainEpochCalculator,
+    ):
+        candidates_count = {}
 
+        def _p_candidates_count(pc_epoch):
```
LGLO, line 36: I don't think it returns candidates count in a pc_epoch. It returns number of committee seats assigned to permissioned candidates. Committee size is not same as PC epoch length.

LGLO: I think it is a matter of rename or doc comment. Reading further I understood it is about committee of pc_epoch.

LGLO (approved):

PR #674, opened 2025-04-15
fix: Ignore decoding errors when filtering governed map utxos

AmbientTea:
# Description

This causes the offchain to not fail if some utxos contain datums that do not conform to the expected schema and instead to ignore those datums completely.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/toolkit/smart-contracts/offchain/src/governed_map/mod.rs
+++ b/toolkit/smart-contracts/offchain/src/governed_map/mod.rs
@@ -67,7 +67,6 @@ pub async fn run_insert<
 	Ok(tx_hash_opt)
 }
 
-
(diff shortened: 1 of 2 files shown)


LGLO:
I don't object this. I just thought that since it is secured by governance, like the D-param, we treat it in the same way.

AmbientTea:
> I don't object this. I just thought that since it is secured by governance, like the D-param, we treat it in the same way.

I'm not sure myself but I think we should be resilient to some trash datums being at the address, even if they're marked with the token. Maybe @jankun4 has some strong arguments here

jankun4 (approved):

PR #675, opened 2025-04-15
add: ETCM-9697 Governed Map list command

AmbientTea:
# Description

Adds the `smart-contracts governed-map list` command and wires it into the demo node.

Output:
```json
[{"key":"key1","value":"0xaaaabbbb0000"},{"key":"key2","value":"0x01"}]
```

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/toolkit/smart-contracts/commands/src/governed_map.rs
+++ b/toolkit/smart-contracts/commands/src/governed_map.rs
@@ -55,14 +55,8 @@ impl InsertCmd {
 			&FixedDelayRetries::five_minutes(),
 		)
 		.await?;
-		Ok(serde_json::json!(result).into())
+		Ok(serde_json::json!(result))
 	}
-}
-
-#[derive(Clone, Debug, clap::ValueEnum)]
-pub enum OutputFormat {
-	Json,
-	Table,
 }
 
 #[derive(Clone, Debug, clap::Parser)]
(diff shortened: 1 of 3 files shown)


kpinter-iohk:
I appreciate `GovernedMapDisplay` and the table implementation but do we really want to take this on ourselves? All of our commands use json and one can pipe to `jq` for better readability or use many tools for dealing with it. This adds complexity that we don't really need; the demo node should provide an example for how to use our framework, not how to print data in a table.

AmbientTea:
> the demo node should provide an example for how to use our framework, not how to print data in a table.

The demo node doesn't implement any tabling, we get that from a library. It implements a single function from bytes to string, which is quite simple

toolkit/smart-contracts/commands/src/lib.rs
```diff
@@ -59,12 +63,18 @@ impl CommonArguments {
 	}
 }
 
+#[derive(From)]
+pub enum CmdOuputValue {
```
LGLO, line 67: I don't understand why can we keep things simple and output only JSON. This ticket was doable with touching 3 files perhaps.

AmbientTea: Alright, I dropped the tabling.

kpinter-iohk (approved):

PR #683, opened 2025-04-16
fix: extend default timeout to 5 minutes

rsporny:
Refs: ETCM-9741


LGLO (approved):
Actually 5 minutes is 59 retries, because the first try is not a retry.

gilligan (approved):

PR #685, opened 2025-04-17
ETCM-9695 governed map - update command

kpinter-iohk:
# Description

Adds update command for governed map

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/toolkit/smart-contracts/commands/src/governed_map.rs
+++ b/toolkit/smart-contracts/commands/src/governed_map.rs
@@ -1,13 +1,15 @@
 use crate::{GenesisUtxo, PaymentFilePath};
 use partner_chains_cardano_offchain::await_tx::FixedDelayRetries;
-use partner_chains_cardano_offchain::governed_map::{run_get, run_insert, run_list, run_remove};
+use partner_chains_cardano_offchain::governed_map::{
+	run_get, run_insert, run_list, run_remove, run_update,
+};
 use serde_json::json;
 use sidechain_domain::byte_string::ByteString;
 
 #[derive(Clone, Debug, clap::Subcommand)]
 #[allow(clippy::large_enum_variant)]
 pub enum GovernedMapCmd {
-	/// Inserts a key-value pair into the Governed Gap. If a value for the key already exists it won't be updated.
+	/// Inserts a key-value pair into the Governed Map. If a value for the key already exists it won't be updated.
 	///
 	/// NOTE: In rare cases, race conditions may occur, and two inserts with the same key will both succeed.
 	/// In that case the second one in terms of block and transaction number is considered valid.
@@ -18,6 +20,8 @@ pub enum GovernedMapCmd {
 	Remove(RemoveCmd),
 	/// Retrieves the value stored in the Governed Map for the given key
 	Get(GetCmd),
+	/// Updates a key-value pair in the Governed Map. If the key is missing it won't be inserted.
+	Update(UpdateCmd),
 }
 
 impl GovernedMapCmd {
@@ -27,6 +31,7 @@ impl GovernedMapCmd {
 			Self::List(cmd) => cmd.execute().await,
 			Self::Remove(cmd) => cmd.execute().await,
 			Self::Get(cmd) => cmd.execute().await,
+			Self::Update(cmd) => cmd.execute().await,
 		}
 	}
 }
@@ -84,6 +89,25 @@ pub struct RemoveCmd {
 	genesis_utxo: GenesisUtxo,
 }
 
+#[derive(Clone, Debug, clap::Parser)]
+pub struct UpdateCmd {
+	#[clap(flatten)]
+	common_arguments: crate::CommonArguments,
+	#[arg(long, help = "The key of the entry, UTF-8 encodable string.")]
+	key: String,
+	#[arg(long, help = "The value of the entry, hex encoded bytes.")]
+	value: ByteString,
+	#[arg(
+		long,
+		help = "If provided update will fail unless the current value matches the one on the ledger."
+	)]
+	current_value: Option<ByteString>,
+	#[clap(flatten)]
+	payment_key_file: PaymentFilePath,
+	#[clap(flatten)]
+	genesis_utxo: GenesisUtxo,
+}
+
 impl ListCmd {
 	pub async fn execute(self) -> crate::SubCmdResult {
 		let client = self.common_arguments.get_ogmios_client().await?;
@@ -133,5 +157,24 @@ impl GetCmd {
 		};
 
 		Ok(json!(value.to_hex_string()).into())
+	}
+}
+impl UpdateCmd {
+	pub async fn execute(self) -> crate::SubCmdResult {
+		let payment_key = self.payment_key_file.read_key()?;
+
+		let client = self.common_arguments.get_ogmios_client().await?;
+
+		let result = run_update(
+			self.genesis_utxo.into(),
+			self.key,
+			self.value,
+			self.current_value,
+			&payment_key,
+			&client,
+			&FixedDelayRetries::five_minutes(),
+		)
+		.await?;
+		Ok(serde_json::json!(result))
 	}
 }
--- a/toolkit/smart-contracts/offchain/src/governed_map/mod.rs
+++ b/toolkit/smart-contracts/offchain/src/governed_map/mod.rs
@@ -27,6 +27,7 @@ use partner_chains_plutus_data::governed_map::{
 };
 use sidechain_domain::byte_string::ByteString;
 use sidechain_domain::{PolicyId, UtxoId};
+use std::ops::Neg;
 
 pub async fn run_insert<
 	C: QueryLedgerState + QueryNetwork + Transactions + QueryUtxoByUtxoId,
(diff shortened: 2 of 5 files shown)


toolkit/smart-contracts/offchain/src/governed_map/mod.rs
```diff
@@ -157,3 +157,160 @@ fn insert_key_value_tx(
...
+	tx_builder.add_mint_one_script_token_using_reference_script(
+		&governance_data.policy.script(),
+		&gov_tx_input,
+		&costs,
+	)?;
+
+	Ok(tx_builder.balance_update_and_build(ctx)?.remove_native_script_witnesses())
+}
+
+fn get_current_value_and_utxo(
```
kpinter-iohk, line 292: This function is a modified version of `get_current_value` and should be consolidated after all commands have been merged.

toolkit/smart-contracts/offchain/src/governed_map/mod.rs
```diff
@@ -316,3 +316,134 @@ pub async fn run_insert_with_force<
...
+	payment_signing_key: &CardanoPaymentSigningKey,
+	ogmios_client: &C,
+	await_tx: &A,
+) -> anyhow::Result<Option<MultiSigSmartContractResult>> {
+	let ctx = TransactionContext::for_payment_key(payment_signing_key, ogmios_client).await?;
+	let (validator, policy) = crate::scripts_data::governed_map_scripts(genesis_utxo, ctx.network)?;
+	let validator_address = validator.address_bech32(ctx.network)?;
+	let validator_utxos = ogmios_client.query_utxos(&[validator_address]).await?;
+
+	let tx_hash_opt = if let Some(actual_current_value) =
```
AmbientTea, line 337: This would be more readable as a `match` like this:
```
let maybe_current_value = get_current_value(validator_utxos.clone(), key.clone(), policy.policy_id());
let tx_hash_opt = match (maybe_current_value, expected_current_value) {
    // wrong current case
    // update case
}

File: toolkit/smart-contracts/offchain/src/governed_map/mod.rs [hunk 3]
```diff
@@ -316,3 +316,134 @@ pub async fn run_insert_with_force<
...
+					ogmios_client,
+				)
+				.await?,
+			)
+		} else {
+			log::info!("Value for key '{key}' is already set to {actual_current_value:?}. Skipping update.");
+			None
+		}
+	} else {
+		return Err(anyhow!("Cannot update nonexistent key :'{key}'."));
```
AmbientTea, line 368: Perhaps if the user doesn't specify `expected_current_value` we should just insert

toolkit/smart-contracts/offchain/src/governed_map/mod.rs
kpinter-iohk: After going this way first, and adding conditions that check if `value == current` to avoid unneeded updates, and checking for `current == expected` and all cases of Some/None, I strongly disagree. Your comment did prompt me to make the conditional logic flatter however.

toolkit/smart-contracts/offchain/src/governed_map/mod.rs
kpinter-iohk: In the discussion about the map operations we decided against upsert.

toolkit/smart-contracts/commands/src/governed_map.rs
```diff
@@ -85,6 +90,22 @@ pub struct RemoveCmd {
...
+	common_arguments: crate::CommonArguments,
+	#[arg(long)]
+	key: String,
+	#[arg(long)]
+	value: ByteString,
+	#[arg(long)]
+	current_value: Option<ByteString>,
+	#[clap(flatten)]
+	payment_key_file: PaymentFilePath,
+	#[arg(long, short('g'))]
```
LGLO, line 105:
```suggestion
	#[arg(long, short('c'))]
```
Please run `./partner-chains-demo-node smart-contracts governance init --help`
We use `c` for it. We use `g` for governance authority parameter.

toolkit/smart-contracts/commands/src/governed_map.rs
```diff
@@ -85,6 +90,22 @@ pub struct RemoveCmd {
...
+#[derive(Clone, Debug, clap::Parser)]
+pub struct UpdateCmd {
+	#[clap(flatten)]
+	common_arguments: crate::CommonArguments,
+	#[arg(long)]
+	key: String,
+	#[arg(long)]
+	value: ByteString,
+	#[arg(long)]
+	current_value: Option<ByteString>,
```
LGLO, line 102: I think `key` and `value` could have a help message, for example about formatting. The first one should be UTF-8 encodable string (with length constrained by the chain runtime settings), the second one should be hex encoded bytes.

This optional parameter for me requires a help message that explains how it affects execution of the command.

toolkit/smart-contracts/offchain/src/governed_map/mod.rs
```diff
@@ -316,3 +316,148 @@ pub async fn run_insert_with_force<
...
+	submit_or_create_tx_to_sign(
+		&governance_data,
+		ctx,
+		|costs, ctx| {
+			update_key_value_tx(
+				validator,
+				policy,
+				key.clone(),
+				value.clone(),
+				validator_utxos,
```
LGLO, line 399: In `update_key_value_tx` it becomes `utxos_for_key`. Where does filtering happen?

toolkit/smart-contracts/offchain/src/governed_map/mod.rs
```diff
@@ -316,3 +316,148 @@ pub async fn run_insert_with_force<
...
+					value,
+					&validator_utxos,
+					ctx,
+					genesis_utxo,
+					ogmios_client,
+				)
+				.await?,
+			)
+		} else {
+			log::info!("Value for key '{key}' is already set to {actual_current_value:?}. Skipping update.");
```
LGLO, line 367: Cosmetic: it will display: "... is set to ByteString(<some hex>). Skipping ...". I mean, we throw Debug format at users.

toolkit/smart-contracts/commands/src/governed_map.rs
```diff
@@ -81,7 +86,26 @@ pub struct RemoveCmd {
...
+pub struct UpdateCmd {
+	#[clap(flatten)]
+	common_arguments: crate::CommonArguments,
+	#[arg(long, help = "The key of the entry, UTF-8 encodable string.")]
+	key: String,
+	#[arg(long, help = "The value of the entry, hex encoded bytes.")]
+	value: ByteString,
+	#[arg(
+		long,
+		help = "If provided update will fail unless the current value matches the one on the ledger. This check is performed off-chain only."
```
jankun4, line 103: What if there are 2 UTxOs for a given key, with different values? I think it would be reasonable if the update consumed both of these utxos then, but maybe there is a reason for why it should fail if even one utxo has wrong value?

toolkit/smart-contracts/offchain/src/governed_map/mod.rs
```diff
@@ -316,3 +317,147 @@ pub async fn run_insert_with_force<
...
+	expected_current_value: Option<ByteString>,
+	payment_signing_key: &CardanoPaymentSigningKey,
+	ogmios_client: &C,
+	await_tx: &A,
+) -> anyhow::Result<Option<MultiSigSmartContractResult>> {
+	let ctx = TransactionContext::for_payment_key(payment_signing_key, ogmios_client).await?;
+	let (validator, policy) = crate::scripts_data::governed_map_scripts(genesis_utxo, ctx.network)?;
+	let validator_address = validator.address_bech32(ctx.network)?;
+	let validator_utxos = ogmios_client.query_utxos(&[validator_address]).await?;
+	let utxos_for_key = get_utxos_for_key(validator_utxos.clone(), key.clone(), policy.policy_id());
```
jankun4, line 337: In rare case, when there are 2 UTxOs for a given key, but with different values I think we should spend both of these utxos.

toolkit/smart-contracts/offchain/src/governed_map/mod.rs
jankun4: Filtering happens in `run_update`

toolkit/smart-contracts/offchain/src/governed_map/mod.rs
kpinter-iohk: I think we do.

toolkit/smart-contracts/commands/src/governed_map.rs
kpinter-iohk: It checks the one returned by `get_current_value`, and then consumes all.

LGLO: When there are many, we behave like the latest one or highest tx hash one (I don't remember). I think we should behave consistently and error if that one is different.
But we are now solving a problem we invented by adding `current_value` parameter.

toolkit/smart-contracts/offchain/src/governed_map/mod.rs
kpinter-iohk: There was no filtering when Lech made his comment, his comment prompted me to add it, as I forgot to add it after some code restructuring. (Thanks Lech)

toolkit/smart-contracts/offchain/src/governed_map/mod.rs
```diff
@@ -316,3 +317,147 @@ pub async fn run_insert_with_force<
...
+		tx_builder.add_mint_script_tokens(
+			policy,
+			&empty_asset_name(),
+			&unit_plutus_data(),
+			&costs.get_mint(&policy.clone()),
+			&Int::new_i32(burn_amount),
+		)?;
+	}
+
+	tx_builder.add_output_with_one_script_token(
```
LGLO, line 455: Don't we have primitive for adding output without minting/burning?

kpinter-iohk: `add_output_with_one_script_token` doesn't mint or burn

PR #687, opened 2025-04-17
Add cmd line args for configuring retry params

kpinter-iohk:
# Description

Adds command line arguments to allow setting retry parameters, while preserving the currently set defaults.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/toolkit/partner-chains-cli/src/deregister/mod.rs
+++ b/toolkit/partner-chains-cli/src/deregister/mod.rs
@@ -15,7 +15,10 @@ use anyhow::anyhow;
 use partner_chains_cardano_offchain::register::Deregister;
 
 #[derive(Clone, Debug, clap::Parser)]
-pub struct DeregisterCmd;
+pub struct DeregisterCmd {
+	#[clap(flatten)]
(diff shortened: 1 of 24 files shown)


AmbientTea (approved):

toolkit/partner-chains-cli/src/deregister/mod.rs
```diff
@@ -15,7 +15,10 @@ use anyhow::anyhow;
 use partner_chains_cardano_offchain::register::Deregister;
 
 #[derive(Clone, Debug, clap::Parser)]
-pub struct DeregisterCmd;
+pub struct DeregisterCmd {
+	#[clap(flatten)]
+	common_arguments: crate::CommonArguments,
```
LGLO, line 20: This somehow surprised me, because we have no other command line parameter in wizards. They use `pc-resources-config.json` file for "operational parameters", like Ogmios URL constituents.

kpinter-iohk: @LGLO I was wondering if I should move these params to the config json. Do you think that would be better?

PR #694, opened 2025-04-22
fix: Mount chains-specs from PVC instead of secrets

skylar-simoncelli:


.github/actions/deploy/deploy-ci-preview/action.yml
```diff
@@ -84,11 +84,11 @@ runs:
...
+        helm upgrade --install ci-preview-validator-1 . -f values/chains/ci-preview.yaml -f values/nodes/ci-preview/validator/ci-preview-validator-1 --set images.substrateNode="${{ inputs.image }}"
+        helm upgrade --install ci-preview-validator-2 . -f values/chains/ci-preview.yaml -f values/nodes/ci-preview/validator/ci-preview-validator-2 --set images.substrateNode="${{ inputs.image }}"
+        helm upgrade --install ci-preview-validator-3 . -f values/chains/ci-preview.yaml -f values/nodes/ci-preview/validator/ci-preview-validator-3 --set images.substrateNode="${{ inputs.image }}"
+        helm upgrade --install ci-preview-validator-4 . -f values/chains/ci-preview.yaml -f values/nodes/ci-preview/validator/ci-preview-validator-4 --set images.substrateNode="${{ inputs.image }}"
+        helm upgrade --install ci-preview-validator-5 . -f values/chains/ci-preview.yaml -f values/nodes/ci-preview/validator/ci-preview-validator-5 --set images.substrateNode="${{ inputs.image }}"
```
LGLO, line 91: Is ci-preview redeployed each time we merge?
This env should not start from genesis and therefore the chain spec file will be there but should not be used by nodes.
When someone wipes one node, then it would not be able to join the rest of nodes.

LGLO (approved):

.github/actions/deploy/deploy-ci-preview/action.yml
skylar-simoncelli: Yes it is redeployed on every PR merge, and PVCs are not wiped:
https://github.com/input-output-hk/partner-chains/blob/4afe61352db9440ad272ece61c2a399b899db7a6/.github/workflows/ci.yml#L585

I understand the edge case you have spotted, but I can't reconcile a way for CI to have knowledge of the previous chain-spec that was in use.

We have no requirement for this environment to support manually wiping one, so for now they are all wiped and redeployed together multiple times per day so this will not be a problem

PR #702, opened 2025-04-23
Tidy up governed map implementation

kpinter-iohk:
# Description

* Rearranges governed map operations to make more sense, both in the code and the help text. (Split to separate commit for easier review.)
* Adds missing help for arguments.
* Removes debug printing and printing of entry values from logs/errors. As entries can be potentially quite large it's better to leave to the user to query them if needed.
* Makes integration tests shorter and adds another check in the test for update.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/toolkit/smart-contracts/commands/src/governed_map.rs
+++ b/toolkit/smart-contracts/commands/src/governed_map.rs
@@ -39,9 +39,9 @@ impl GovernedMapCmd {
 pub struct InsertCmd {
 	#[clap(flatten)]
 	common_arguments: crate::CommonArguments,
-	#[arg(long)]
+	#[arg(long, help = "The key of the entry, UTF-8 encodable string.")]
 	key: String,
-	#[arg(long)]
+	#[arg(long, help = "The value of the entry, hex encoded bytes.")]
 	value: ByteString,
 	#[clap(flatten)]
 	payment_key_file: PaymentFilePath,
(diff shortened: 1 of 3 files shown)


toolkit/smart-contracts/commands/src/governed_map.rs
```diff
@@ -98,7 +78,7 @@ pub struct UpdateCmd {
 	value: ByteString,
 	#[arg(
 		long,
-		help = "If provided update will fail unless the current value matches the one on the ledger."
+		help = "If provided, update will fail unless the current value matches the one on the ledger."
```
LGLO, line 81: While we are here, I'm 51% sure it is "in the ledger", not "on the ledger".

LGLO (approved):

toolkit/smart-contracts/commands/src/governed_map.rs
kpinter-iohk: I've looked into this a bit now. "In the ledger" is appropriate when talking about ledgers in the classical sense as a book in which things are recorded. However since distributed ledgers are on computers, for those "on the ledger" would be more appropriate. I recall only ever seeing "on the ledger" previously in this context (ctrl+f for example [here](https://blog.digitalasset.com/blog/smart-contract-language-the-real-arbiter-of-truth) or [here](https://iohk.io/en/blog/posts/2020/12/08/native-tokens-on-cardano/)).
Note that "in the ledger" also occurs ([for example](https://developers.cardano.org/docs/get-started/cardano-node/cardano-components/)) but in the context of describing components of the ledger software.
So I'd think for example data is recorded *on* a distributed ledger, and smart contract logic is implemented *in* a distributed ledger.
cc @joseph-fajen

PR #726, opened 2025-04-29
feat: add block producer fees pallet

LGLO:
# Description

Adds `block-producer-fees` pallet.

Requirements were trimmed down to one setting for margin fee, that should be in range between 0% and 100%, with precision of 1/100th of a percent.

The configurable number of historical items and storing the slot of a change is added to satisfy requirements:
* "Changes to the fee in epoch n MUST take effect in epoch n+1" - rewards will be able to use value that is not the latest
* "There must be a mechanism that allows for delegators to be notified of a change in margin fee"

RPC and benchmarking will be done separately.

Ticket with requirements: https://input-output.atlassian.net/browse/ETCM-9326

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/toolkit/block-producer-fees/pallet/src/lib.rs
+++ b/toolkit/block-producer-fees/pallet/src/lib.rs
@@ -1,4 +1,4 @@
-//! Pallet to store Block Producer Fees settings that are important in context of rewards payments.
+//! Pallet to store Block Producer Fees settings that are relevant to rewards payments.
 //!
 //! Margin fee is percent of block rewards that will be paid to the block producer before
 //! distributing the rest of rewards to his stakers. Precision of the margin fee setting is bounded
@@ -79,7 +79,6 @@ pub mod pallet {
 					let _ = fees_log.pop_back();
 				}
 				fees_log.push_front((T::current_slot(), fee_numerator));
-				fees_log.clone()
 			});
 			Ok(())
 		}
--- a/toolkit/block-producer-fees/pallet/src/lib.rs
+++ b/toolkit/block-producer-fees/pallet/src/lib.rs
@@ -1,0 +1,116 @@
+//! Pallet to store Block Producer Fees settings that are important in context of rewards payments.
+//!
+//! Margin fee is percent of block rewards that will be paid to the block producer before
+//! distributing the rest of rewards to his stakers. Precision of the margin fee setting is bounded
+//! to 1/100 of a percent.
+//!
+//! The margin fees are stored together with the slot at which change occurred, so this data can be
+//! exposed to rewards calculation.
+//!
+//! Log of changes per account is bounded. The oldest entries are dropped when new ones are added.
+//! Intention is to discourage users from too frequent changes and there is an assumption
+//! that rewards calculation algorithm will account for it.
+
+#![cfg_attr(not(feature = "std"), no_std)]
+#![deny(missing_docs)]
+
+#[cfg(test)]
+mod mock;
+#[cfg(test)]
+mod tests;
+pub mod weights;
+
+pub use pallet::*;
+pub use weights::WeightInfo;
+
+#[frame_support::pallet]
+pub mod pallet {
+	use super::*;
+	use frame_support::pallet_prelude::*;
+	use frame_system::pallet_prelude::*;
+	use sp_consensus_slots::Slot;
+	use sp_std::collections::vec_deque::VecDeque;
+
+	/// Current version of the pallet
+	pub const PALLET_VERSION: u32 = 1;
+
+	#[pallet::pallet]
+	pub struct Pallet<T>(_);
+
+	#[pallet::config]
+	pub trait Config: frame_system::Config {
+		/// Maximum number of past changes per one block producer kept in the storage.
+		type HistoricalChangesPerProducer: Get<u16>;
+
+		/// Weight information on extrinsic in the pallet. For convenience weights in [weights] module can be used.
+		type WeightInfo: WeightInfo;
+
+		/// The slot number of the current block.
+		fn current_slot() -> Slot;
+	}
+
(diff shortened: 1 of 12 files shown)


Klapeyron:
we can try to utilize [OnChargeTransaction](https://docs.rs/pallet-transaction-payment/latest/pallet_transaction_payment/pallet/trait.Config.html#associatedtype.OnChargeTransaction) from `pallet_transaction_payment`, I think that additional pallet is unnecessary.

toolkit/block-producer-fees/pallet/src/weights.rs
```diff
@@ -0,0 +1,14 @@
+//! Weights for runtime calls, output of benchmarking
+use frame_support::weights::Weight;
+
+/// Weight functions needed for the pallet.
+pub trait WeightInfo {
+	/// Yes
```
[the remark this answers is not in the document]
LGLO: I'll fix it. Dummy doc to pass compilation.

toolkit/block-producer-fees/pallet/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+
+		/// Weight information on extrinsic in the pallet. For convenience weights in [weights] module can be used.
+		type WeightInfo: WeightInfo;
+
+		/// The slot number of the current block.
+		fn current_slot() -> Slot;
+	}
+
+	// Margin Fee precision is 0.01 of a percent, so use 1/10000 as unit.
+	type PerTenThousands = u16;
```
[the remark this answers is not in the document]
LGLO: Substrate uses `Permill`, `Perbill`, `Percent`. I'll stick to "Per"

toolkit/block-producer-fees/pallet/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+			})
+		}
+
+		/// Retrieves block producers fees settings.
+		pub fn get(id: T::AccountId) -> VecDeque<FeeChange> {
+			FeesChanges::<T>::get(id)
+		}
+
+		/// Gets the latest fee setting for the given account.
+		pub fn get_latest(id: T::AccountId) -> Option<FeeChange> {
```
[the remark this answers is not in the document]
LGLO: No, because there can be no collection for given `id`.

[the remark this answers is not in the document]
kpinter-iohk:
```rust
FeesChanges::<T>::get(id).front().cloned()
```
I still don't understand this. Shouldn't this be something like:
```rust
FeesChanges::<T>::try_get(id)
	.ok()
	.map(|deq| *deq.front().expect("There are no empty collections in storage"))
```
If there exists an invalid entry where an account id is in the storage map but the associated deque is empty, wouldn't the current code just return `None`, when in fact there is an invalid state? Or is it the case that the invalid state is impossible (we are just not proving it with types) and therefore we are ignoring it?

toolkit/block-producer-fees/pallet/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+			let account_id = ensure_signed(origin)?;
+			if fee_numerator > 10000 {
+				return Err(DispatchError::Other("fee numerator must be in range from 0 to 10000"));
+			}
+			FeesChanges::<T>::mutate(account_id, |fees_log| {
+				if fees_log.len() > T::HistoricalChangesPerProducer::get().into() {
+					let _ = fees_log.pop_back();
+				}
+				fees_log.push_front((T::current_slot(), fee_numerator));
+				fees_log.clone()
```
AmbientTea, line 82: is this clone needed here?

toolkit/block-producer-fees/pallet/src/lib.rs
```diff
@@ -0,0 +1,116 @@
...
+		pub fn set_fee(origin: OriginFor<T>, fee_numerator: PerTenThousands) -> DispatchResult {
+			let account_id = ensure_signed(origin)?;
+			if fee_numerator > 10000 {
+				return Err(DispatchError::Other("fee numerator must be in range from 0 to 10000"));
+			}
+			FeesChanges::<T>::mutate(account_id, |fees_log| {
+				if fees_log.len() > T::HistoricalChangesPerProducer::get().into() {
+					let _ = fees_log.pop_back();
+				}
+				fees_log.push_front((T::current_slot(), fee_numerator));
```
AmbientTea, line 81: I see a potential issue here, where any entry can be wiped from history by invoking this extrinsic enough times. If the assumption is that the fee settings used should come from eg. current epoch - 1 etc., what is the expected behaviour here? I know that falls on the payout rewards implementer to decide, but I don't see a good solution here. The mechanism could default to some pre-defined fee values when there's no settings old enough in the storage - but then, unless the default is a 0% fee, a validator could set the fee lower than the default to attract more delegators and then call the extrinsic enough times to wipe the settings and bump the fee back to default.

AmbientTea: I think we could avoid this issue by eg. allowing `fees_log.pop_back()` only if it still leaves an entry in the history that is old enough to be applied by the payout. The value of "old enough" could be configured in the pallet config.

LGLO:
```
//! Log of changes per account is bounded. The oldest entries are dropped when new ones are added.
//! Intention is to discourage users from too frequent changes and there is an assumption
//! that rewards calculation algorithm will account for it.
```

toolkit/block-producer-fees/pallet/src/lib.rs
LGLO: No, it is not. This line will be removed.

PR #732, opened 2025-04-29
add: ETCM-9734 caching for governed map data

jankun4:
# Description

[link](https://input-output.atlassian.net/browse/ETCM-9734)
Adds Cached GovernedMap DataSource implementation, and uses it in demo-node

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/demo/node/src/data_sources.rs
+++ b/demo/node/src/data_sources.rs
@@ -2,7 +2,7 @@ use authority_selection_inherents::authority_selection_inputs::AuthoritySelectio
 use pallet_sidechain_rpc::SidechainRpcDataSource;
 use partner_chains_db_sync_data_sources::{
 	block::BlockDataSourceImpl, candidates::CandidatesDataSourceImpl,
-	governed_map::GovernedMapDataSourceImpl, mc_hash::McHashDataSourceImpl,
+	governed_map::GovernedMapDataSourceCachedImpl, mc_hash::McHashDataSourceImpl,
 	metrics::McFollowerMetrics, native_token::NativeTokenManagementDataSourceImpl,
 	sidechain_rpc::SidechainRpcDataSourceImpl, stake_distribution::StakeDistributionDataSourceImpl,
 };
@@ -69,7 +69,6 @@ pub fn create_mock_data_sources()
 
 pub const CANDIDATES_FOR_EPOCH_CACHE_SIZE: usize = 64;
 pub const STAKE_CACHE_SIZE: usize = 100;
-
 pub async fn create_cached_db_sync_data_sources(
 	metrics_opt: Option<McFollowerMetrics>,
 ) -> Result<DataSources, Box<dyn Error + Send + Sync + 'static>> {
@@ -96,6 +95,9 @@ pub async fn create_cached_db_sync_data_sources(
 			metrics_opt.clone(),
 			STAKE_CACHE_SIZE,
 		)),
-		governed_map: Arc::new(GovernedMapDataSourceImpl::new(pool, metrics_opt)),
+		governed_map: Arc::new(GovernedMapDataSourceCachedImpl::new_from_env(
+			pool,
+			metrics_opt.clone(),
+		)?),
 	})
 }
--- a/demo/node/src/data_sources.rs
+++ b/demo/node/src/data_sources.rs
@@ -69,6 +69,7 @@ pub fn create_mock_data_sources()
 
 pub const CANDIDATES_FOR_EPOCH_CACHE_SIZE: usize = 64;
 pub const STAKE_CACHE_SIZE: usize = 100;
+pub const GOVERNED_MAP_CACHE_SIZE: u16 = 100;
 pub async fn create_cached_db_sync_data_sources(
 	metrics_opt: Option<McFollowerMetrics>,
 ) -> Result<DataSources, Box<dyn Error + Send + Sync + 'static>> {
@@ -80,7 +81,7 @@ pub async fn create_cached_db_sync_data_sources(
 			block.clone(),
 			metrics_opt.clone(),
 		)),
-		mc_hash: Arc::new(McHashDataSourceImpl::new(block, metrics_opt.clone())),
+		mc_hash: Arc::new(McHashDataSourceImpl::new(block.clone(), metrics_opt.clone())),
 		authority_selection: Arc::new(
 			CandidatesDataSourceImpl::new(pool.clone(), metrics_opt.clone())
 				.await?
@@ -95,9 +96,11 @@ pub async fn create_cached_db_sync_data_sources(
 			metrics_opt.clone(),
 			STAKE_CACHE_SIZE,
 		)),
-		governed_map: Arc::new(GovernedMapDataSourceCachedImpl::new_from_env(
+		governed_map: Arc::new(GovernedMapDataSourceCachedImpl::new(
 			pool,
 			metrics_opt.clone(),
-		)?),
+			GOVERNED_MAP_CACHE_SIZE,
+			block,
+		)),
 	})
 }
--- a/toolkit/data-sources/db-sync/src/db_model.rs
+++ b/toolkit/data-sources/db-sync/src/db_model.rs
@@ -114,6 +114,14 @@ pub(crate) struct DatumOutput {
 	pub datum: DbDatum,
 }
 
+#[derive(Debug, Clone, sqlx::FromRow, PartialEq)]
+pub(crate) struct DatumChangeOutput {
+	pub datum: DbDatum,
+	pub block_no: BlockNumber,
+	pub block_index: TxIndexInBlock,
+	pub action: String,
+}
+
 #[derive(Debug, Clone, PartialEq)]
 pub(crate) struct NativeTokenAmount(pub u128);
 impl From<NativeTokenAmount> for sidechain_domain::NativeTokenAmount {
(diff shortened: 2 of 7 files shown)


toolkit/governed-map/primitives/src/lib.rs
```diff
@@ -125,8 +125,8 @@ pub const INHERENT_IDENTIFIER: InherentIdentifier = *b"govrnmap";
 pub struct MainChainScriptsV1 {
 	/// Cardano address of the Governed Map validator, at which UTXOs containig key-value pairs are located
 	pub validator_address: MainchainAddress,
-	/// Policy of the asset used to mark the UTXOs containing the Governed Map's key-value pairs
-	pub asset_policy_id: PolicyId,
+	/// Asset used to mark the UTXOs containing the Governed Map's key-value pairs
+	pub asset: AssetId,
```
LGLO, line 129: Do you just revert @AmbientTea fix from today?

toolkit/data-sources/db-sync/src/governed_map/tests.rs
```diff
@@ -70,13 +106,35 @@ async fn test_governed_map_upsert(pool: PgPool) {
...
+async fn test_cached_governed_map_upsert(pool: PgPool) {
+	let source = make_cached_source(pool);
+	let mut result = source.get_mapping_changes(Some(BLOCK_6), BLOCK_7, scripts()).await.unwrap();
+	result.sort();
+}
+
 fn scripts() -> MainChainScriptsV1 {
 	MainChainScriptsV1 {
-		asset_policy_id: PolicyId(hex!("500000000000000000000000000000000000434845434b504f494e69")),
+		asset: AssetId {
```
LGLO, line 118: Don't revert https://github.com/input-output-hk/partner-chains/pull/730/files#diff-ebbcafa36b97f2598b8cbbbf6f5f72e65399212eb90b34f7b154a4518f110a78L75

toolkit/data-sources/db-sync/src/db_model.rs
```diff
@@ -367,6 +375,53 @@ pub(crate) async fn get_token_utxo_for_epoch(
 		.await?)
 }
 
+#[cfg(feature = "governed-map")]
+pub(crate) async fn get_changes(
+	pool: &Pool<Postgres>,
+	address: &Address,
+	last_block: Option<BlockNumber>,
+	new_block: BlockNumber,
```
LGLO, line 383: I propose "after_block" (or "since_block", but "after" is more precise to me, because "since" isn't that clear about inclusion) and "to_block" instead of "last_block" and "new_block" in this layer.

toolkit/data-sources/db-sync/src/governed_map/mod.rs
```diff
@@ -88,3 +92,229 @@ impl GovernedMapDataSourceImpl {
...
+		) -> std::result::Result<
+			Vec<(String, Option<ByteString>)>,
+			Box<dyn std::error::Error + Send + Sync>,
+		> {
+			if let Ok(mut cache) = self.cache.lock() {
+				cache.set_address(scripts.validator_address.clone().into());
+				cache.set_asset(scripts.asset_policy_id.clone());
+			};
+
+			let since_block_number = match since_mc_block {
```
LGLO, line 122: We have block data source in this crate that has caching of stable blocks. Shouldn't it be used here?

toolkit/data-sources/db-sync/src/db_model.rs
```diff
@@ -367,6 +375,53 @@ pub(crate) async fn get_token_utxo_for_epoch(
...
+		LEFT JOIN block consuming_block	ON consuming_tx.block_id = consuming_block.id
+		INNER JOIN datum				ON tx_out.data_hash = datum.hash
+		INNER JOIN ma_tx_out			ON tx_out.id = ma_tx_out.tx_out_id
+		INNER JOIN multi_asset			ON multi_asset.id = ma_tx_out.ident
+		WHERE
+			tx_out.address = $1
+			AND (consuming_tx_in.id IS NOT NULL AND ($2 IS NULL OR consuming_block.block_no > $2) AND consuming_block.block_no <= $3)
+			AND multi_asset.policy = $4
+			AND multi_asset.name = $5)
+		ORDER BY block_no ASC";
```
LGLO, line 414: Order by block_index as well?

toolkit/data-sources/db-sync/src/governed_map/mod.rs
```diff
@@ -88,3 +92,229 @@ impl GovernedMapDataSourceImpl {
...
+	address: Option<MainchainAddress>,
+	policy_id: Option<PolicyId>,
+}
+
+fn deduplicate_changes(
+	mut changes: Vec<(String, Option<ByteString>, BlockNumber, TxIndexInBlock)>,
+) -> Vec<(String, Option<ByteString>)> {
+	let mut result = HashMap::new();
+
+	changes.sort_by_key(|(_, _, block_number, block_index)| (*block_number, *block_index));
```
LGLO, line 239: I think they can be sorted by the SQL query and don't need sorting again.

toolkit/data-sources/db-sync/src/governed_map/mod.rs
```diff
@@ -88,3 +92,229 @@ impl GovernedMapDataSourceImpl {
...
+
+	for (key, value, _, _) in changes {
+		if result.contains_key(&key) {
```
LGLO, line 242: Is it necessary? Wouldn't `insert` overwrite anyway?

toolkit/data-sources/db-sync/src/governed_map/mod.rs
```diff
@@ -88,3 +92,229 @@ impl GovernedMapDataSourceImpl {
...
+			result.remove(&key);
+		}
+		result.insert(key, value);
+	}
+
+	result.into_iter().collect()
```
LGLO, line 248: Ordering is lost after putting to the hash map and getting it back, so the result is invalid.

toolkit/data-sources/db-sync/src/governed_map/mod.rs
LGLO: I think there is one liner for creating this hash map:
`let m: HashMap<_, _> = changes.into_iter().map(|(key, value, _, _)|(key, value)).collect()`

toolkit/data-sources/db-sync/src/governed_map/tests.rs
```diff
@@ -70,6 +106,13 @@ async fn test_governed_map_upsert(pool: PgPool) {
 	assert_eq!(result, expected);
 }
 
+#[sqlx::test(migrations = "./testdata/governed-map/migrations")]
+async fn test_cached_governed_map_upsert(pool: PgPool) {
+	let source = make_cached_source(pool);
+	let mut result = source.get_mapping_changes(Some(BLOCK_6), BLOCK_7, scripts()).await.unwrap();
+	result.sort();
```
LGLO, line 113: This test has no assertion

demo/node/src/data_sources.rs
```diff
@@ -96,6 +95,9 @@ pub async fn create_cached_db_sync_data_sources(
 			metrics_opt.clone(),
 			STAKE_CACHE_SIZE,
 		)),
-		governed_map: Arc::new(GovernedMapDataSourceImpl::new(pool, metrics_opt)),
```
LGLO, line 99: Should we cleanup codebase from `GovernedMapDataSourceImpl` ?

toolkit/data-sources/db-sync/src/governed_map/mod.rs
```diff
@@ -88,3 +92,229 @@ impl GovernedMapDataSourceImpl {
...
+}
+
+fn filter_changes_in_range(
+	changes: Vec<(String, Option<ByteString>, BlockNumber, TxIndexInBlock)>,
+	since_block: Option<BlockNumber>,
+	up_to_block: BlockNumber,
+) -> Vec<(String, Option<ByteString>, BlockNumber, TxIndexInBlock)> {
+	changes
+		.iter()
```
AmbientTea, line 257: `into_iter` to avoid cloning later

toolkit/data-sources/db-sync/src/governed_map/mod.rs
```diff
@@ -88,3 +92,229 @@ impl GovernedMapDataSourceImpl {
...
+				&& since_block.map(|b| block_number.0 > b.0).unwrap_or(true)
+		})
+		.map(|x| x.clone())
+		.collect()
+}
+
+async fn get_latest_block(pool: &PgPool) -> Result<Block> {
+	crate::db_model::get_latest_block_info(pool).await?.ok_or(
+		DataSourceError::ExpectedDataNotFound(
+			"The latest block not found when querying for native token transfers".to_string(),
```
AmbientTea, line 269: "native token"

toolkit/data-sources/db-sync/src/governed_map/mod.rs
AmbientTea: It should.

toolkit/data-sources/db-sync/src/governed_map/mod.rs
```diff
@@ -88,3 +92,229 @@ impl GovernedMapDataSourceImpl {
...
+			else {
+				return Err(Box::new(ExpectedDataNotFound(format!("Block hash: {up_to_mc_block}"))));
+			};
+
+			if let Some(cached_changes) = self.get_changes_from_cache(since_block_number, up_to_block.block_no).await? {
+				return Ok(cached_changes);
+			}
+
+			let latest_block = get_latest_block(&self.pool).await?;
+			let latest_stable_block = BlockNumber(latest_block.block_no.0.saturating_sub(self.security_parameter));
```
AmbientTea, line 142: `BlockDataSourceImpl` solves this already

toolkit/data-sources/db-sync/src/governed_map/mod.rs
```diff
@@ -88,3 +92,229 @@ impl GovernedMapDataSourceImpl {
...
+			let since_block_plus = BlockNumber(since_block_number.unwrap_or(BlockNumber(0)).0 + self.cache_size as u32);
+			let max_search_block = min(latest_stable_block, max(up_to_block.block_no, since_block_plus));
+
+			let changes = self.get_changes_in_range_to_cache(since_block_number, max_search_block, scripts).await?;
+
+			if let Ok(mut cache) = self.cache.lock() {
+				cache.add_changes(changes.clone());
+			}
```
AmbientTea, line 150: I think this is unsafe. If the data source is called with two block ranges that are spread apart, the cache will be missing any changes that happened between them.

We should either unconditionally replace the cache here, or have a check that the queried range and the cache overlap to make sure the extension is safe (and we need to drop the overlaping changes to avoid duplicates)

toolkit/data-sources/db-sync/src/governed_map/mod.rs
```diff
@@ -88,3 +92,229 @@ impl GovernedMapDataSourceImpl {
...
+			.and_then(|s| s.parse().ok())
+			.ok_or("Couldn't read env variable CARDANO_SECURITY_PARAMETER as u32")?;
+		Ok(Self {
+			pool,
+			metrics_opt,
+			security_parameter,
+			cache_size: 1000,
+			cache: Default::default(),
+		})
+	}
```
AmbientTea, line 173: I think this constructor should accept `Arc<dyn BlockDataSource>` so we don't need to do block stability calculations manually in this data source.

AmbientTea (changes requested):
See my comment on the cache safety. I think we're missing some checks. My suggestion is to just replace the cache on every miss, this is what we're doing with native tokens and I don't see a scenario where having more complex cache extension logic makes has any substantial benefits.

demo/node/src/data_sources.rs
jankun4: @AmbientTea asked to keep it for now, to test if old implementation can work hand in hand with the new one

toolkit/data-sources/db-sync/src/governed_map/mod.rs
```diff
@@ -88,3 +92,204 @@ impl GovernedMapDataSourceImpl {
...
+			Vec<(String, Option<ByteString>)>,
+			Box<dyn std::error::Error + Send + Sync>,
+		> {
+			if let Ok(mut cache) = self.cache.lock() {
+				cache.set_address(scripts.validator_address.clone().into());
+				cache.set_asset(scripts.asset_policy_id.clone());
+			};
+
+			let since_block_number = match since_mc_block {
+				Some(hash) =>
```
kpinter-iohk, line 123: Looks like a `map`.

jankun4: Not really, because there is async function in there

PR #734, opened 2025-04-30
ETCM-9687 docs for committee selection rpc, query, primitives

kpinter-iohk:
# Description

My rustdoc is terse, please comment with ideas on what I could expand upon if you have any.

# Checklist

- [ ] Self-reviewed the diff


toolkit/committee-selection/query/src/commands.rs
```diff
@@ -1,6 +1,11 @@
+//! Implementations for CLI query commands for getting Ariadne params and registration status
 use crate::SessionValidatorManagementQueryApi;
 use sidechain_domain::{McEpochNumber, StakePoolPublicKey};
 
+/// Returns Ariadne parameters in JSON format
+/// Arguments:
+///  - query: query api implementing [SessionValidatorManagementQueryApi] trait
+///  - mc_epoch_number: mainchain epoch for which to get Ariadne parameters
```
LGLO, line 8: If I query for epoch 800, will it return Ariadne Parameters that are set in epoch 800 or that are used by committee selection during Cardano epoch 800?

toolkit/committee-selection/query/src/types/ariadne.rs
```diff
@@ -6,16 +6,23 @@ use sidechain_domain::{AuraPublicKey, GrandpaPublicKey, SidechainPublicKey};
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
+/// Parameters for Ariadne committee selection
 pub struct AriadneParameters {
+	/// The d-parameter
```
LGLO, line 11: I'm 51% sure that it is "D-parameter" (uppercase)

toolkit/committee-selection/query/src/types/mod.rs
```diff
@@ -43,6 +48,7 @@ impl GetCommitteeResponse {
 		GetCommitteeResponse { sidechain_epoch: sidechain_epoch.0, committee }
 	}
 
+	/// Legacy constructor
```
LGLO, line 51:
```suggestion
	/// Legacy constructor, used for input coming from old versions of the pallet runtime
```

toolkit/committee-selection/query/src/types/registrations.rs
```diff
@@ -15,41 +15,55 @@ use sp_runtime::{MultiSigner, traits::IdentifyAccount};
...
+	/// A wrapped [StakeError]
 	StakeError(#[from] StakeError),
 	#[error("{0}")]
+	/// A wrapped [RegistrationDataError]
 	InvalidRegistrationData(#[from] RegistrationDataError),
 }
 
 #[derive(Default, Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(rename_all = "camelCase")]
+/// Candidate Registration Entry
```
LGLO, line 30: I have a problem with this one, because these docs can satisfy cargo, but they don't explain peculiarities of the type itself. It doesn't really have to be explain for this type, but it should be documented somewhere:
* what is "sidechain_pub_key" (I would say, that "sidechain_signature" is than easy to figure out)
* what is "cross_chain_pub_key"
* what is relation of these two?

toolkit/committee-selection/query/src/types/registrations.rs
```diff
@@ -15,41 +15,55 @@ use sp_runtime::{MultiSigner, traits::IdentifyAccount};
...
+	/// Sidechain signature of the candidate
 	pub sidechain_signature: String,
 	/// Signature made with Stake Pool key
 	pub mainchain_signature: String,
+	/// Cross chain signature of the candidate
 	pub cross_chain_signature: String,
 	/// Data of Utxo that contained this registration"
 	pub utxo: UtxoInfo,
 	#[serde(skip_serializing_if = "Option::is_none")]
+	/// Optional stake delegation amount
```
LGLO, line 53: When is it None? Could you point that it is total stake delegated to the pool identified by `mainchain_pub_key`, please?

toolkit/committee-selection/query/src/types/registrations.rs
```diff
@@ -15,41 +15,55 @@ use sp_runtime::{MultiSigner, traits::IdentifyAccount};
...
 	pub cross_chain_pub_key: String,
+	/// Aura pub key of the candidate
 	pub aura_pub_key: String,
+	/// Grandpa pub key of the candidate
 	pub grandpa_pub_key: String,
+	/// Sidechain signature of the candidate
 	pub sidechain_signature: String,
 	/// Signature made with Stake Pool key
 	pub mainchain_signature: String,
+	/// Cross chain signature of the candidate
```
LGLO, line 48:
```suggestion
	/// Cross chain key signature for the registration message
```

toolkit/committee-selection/query/src/types/registrations.rs
kpinter-iohk: I've investigated when this is None: when there are no entries returned by [`get_stake_distribution` query](https://github.com/input-output-hk/partner-chains/blob/ETCM-9687-committee-selection-docs-1/toolkit/data-sources/db-sync/src/db_model.rs#L323-L332) which is odd to me as it queries all stake pools and is not specific to the candidate registration.
[The code later](https://github.com/input-output-hk/partner-chains/blob/ETCM-9687-committee-selection-docs-1/toolkit/data-sources/db-sync/src/candidates/mod.rs#L192-L206) queries for a specific stake pool key, and if it's not present defaults to 0.
This makes no sense to me, I think this field should not be optional and `None` should map to 0 instead.

toolkit/committee-selection/query/src/types/registrations.rs
kpinter-iohk: I think a good solution is to reference the actual domain types in the rustdoc which I've done now.

gilligan (approved):

PR #736, opened 2025-04-30
change: ETCM-9800 track and handle initialization of the governed map pallet

AmbientTea:
# Description

* Correctly handles changing main chain scripts by tracking "initialized" state in the pallet and resetting it on script change
* Modifes the IDP to work differently for initialized and uninitialized pallet state:
    * uninitialized: fetches full Governed Map state, compares it with the pallet's storage and provides the changes as inherent data. This is necessary when the pallet needs to be re-initialized after main chain script change, but it already contains some data.
    * initialized: only fetches changes since last MC hash and provides them as inherent data
* Modifies the runtime API to also provide initialization status and current mappings
* Modifies the data source API to provide methods to both 1) query full state at block; 2) query changes between blocks
* Uses BTreeMap for changes in the pallet to force one-key-one-change invariant

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/toolkit/data-sources/db-sync/src/governed_map/mod.rs
+++ b/toolkit/data-sources/db-sync/src/governed_map/mod.rs
@@ -13,7 +13,6 @@ use sidechain_domain::*;
 use sp_governed_map::{GovernedMapDataSource, MainChainScriptsV1};
 use sqlx::PgPool;
 use std::cmp::{max, min};
-use std::collections::HashMap;
 use std::sync::{Arc, Mutex};
 
 #[cfg(test)]
@@ -27,6 +26,15 @@ pub struct GovernedMapDataSourceImpl {
 
 observed_async_trait!(
 impl GovernedMapDataSource for GovernedMapDataSourceImpl {
+	async fn get_state_at_block(
+		&self,
+		mc_block: McBlockHash,
+		scripts: MainChainScriptsV1,
+	) -> std::result::Result<BTreeMap<String, ByteString>, Box<dyn std::error::Error + Send + Sync>>
+	{
+		Ok(get_mappings_entries(&self.pool, mc_block, scripts).await?.into())
+	}
+
 	async fn get_mapping_changes(
 		&self,
 		since_mc_block: Option<McBlockHash>,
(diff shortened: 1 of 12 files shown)


toolkit/data-sources/db-sync/src/governed_map/mod.rs
```diff
@@ -22,6 +21,15 @@ pub struct GovernedMapDataSourceImpl {
 
 observed_async_trait!(
 impl GovernedMapDataSource for GovernedMapDataSourceImpl {
+	async fn get_current_mappings(
```
LGLO, line 24: I wouldn't say it is "current", it is more "at_block".

toolkit/governed-map/pallet/src/lib.rs
```diff
@@ -233,18 +246,22 @@ pub mod pallet {
...
 			let mut changes = Changes::<T>::new();
-			for GovernedMapChangeV1 { key, new_value } in raw_changes {
+			for (key, new_value) in raw_changes {
 				let new_value = match new_value {
 					Some(new_value) => Some(Self::bound_value(&key, new_value)?),
 					None => None,
 				};
 				let key = Self::bound_key(&key)?;
-				changes.try_push((key, new_value)).expect("Number of changes is below maximum");
+				changes.try_insert(key, new_value).expect("Number of changes is below maximum");
```
LGLO, line 264: Late spot, but why is it expected to not panic? What makes us sure that "Number of changes is below maximum"?

toolkit/governed-map/primitives/src/lib.rs
```diff
@@ -403,6 +423,10 @@ sp_api::decl_runtime_apis! {
 	#[api_version(1)]
 	pub trait GovernedMapIDPApi
 	{
+		/// Returns initialization state of the pallet
+		fn is_initialized() -> bool;
+		/// Returns an iterator over all mappings currently stored in the pallet
```
LGLO, line 428: Not an iterator but BTreeMap, which I like more btw. Being strict at output is a good thing.

toolkit/governed-map/primitives/src/lib.rs
```diff
@@ -284,6 +279,13 @@ impl sp_inherents::InherentDataProvider for GovernedMapInherentDataProvider {
 #[cfg(feature = "std")]
 #[async_trait::async_trait]
 pub trait GovernedMapDataSource {
+	/// Returns all key-value mappings stored in the Governed Map on Cardano after execution of `mc_block`.
+	async fn get_current_mappings(
```
LGLO, line 283:
```suggestion
	async fn get_state_at_block(
```
state vs mappings => I would like to have one name for this across all layers. Runtime api has "get_current_state".

"current" vs "at_block" => because current could mean at the time of invoking this function

toolkit/governed-map/pallet/src/lib.rs
```diff
@@ -209,22 +217,24 @@ pub mod pallet {
 		}
 
 		fn is_inherent(call: &Self::Call) -> bool {
-			matches!(call, Call::register_changes { .. })
+			match call {
```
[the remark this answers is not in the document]
AmbientTea: Absolutely nothing, at some point I had a separate inherent for initialization so I had to use match here and didn't switch back when I unified them again.

toolkit/governed-map/pallet/src/lib.rs
AmbientTea:
```rust
if raw_changes.len() > T::MaxChanges::get() as usize {
				return Err(TooManyChanges);
			}
```

line 256 above guards that

LGLO: Right

LGLO (approved):

PR #737, opened 2025-05-01
feat: allow ogmios url to be configured via env var.

gilescope:
# Description

Example usecase:

Allow running a partner chain's cli inside of docker. They may or may not be calling a command
that needs to use Ogmios. But if we can set up OGMIOS_URL=ws://host.docker.internal:1337 env var in the launch script then things will always work for them. (We could put various containers in a common network but it would be nice to have this option).

# Checklist

- [x] Self-reviewed the diff


AmbientTea (approved):

AmbientTea:
🫡

LGLO (approved):

PR #738, opened 2025-05-05
ETCM-9687 committee selection pallet docs

kpinter-iohk:
# Description

.

# Checklist

- [ ] Commit sequence broadly makes sense and commits have useful messages.
- [ ] The size limit of 400 LOC isn't needlessly exceeded
- [ ] The PR refers to a JIRA ticket (if one exists)
- [ ] New tests are added if needed and existing tests are updated.
- [ ] New code is documented and existing documentation is updated.
- [ ] Relevant logging and metrics added
- [ ] Any changes are noted in the `changelog.md` for affected crate
- [ ] Self-reviewed the diff

PR #741, opened 2025-05-05
remove: unused module `pallet_session_support`

AmbientTea:
# Description

Turns out this module has not been in use anymore.

# Checklist

- [ ] Self-reviewed the diff


kpinter-iohk (changes requested):
I have also noticed that this module is not wired into the demo node, but investigating I found that it never has been since it was added in #171 and seems to be included as an example, which I documented in #738. @LGLO please confirm.

AmbientTea:
> I have also noticed that this module is not wired into the demo node, but investigating I found that it never has been since it was added in #171 and seems to be included as an example, which I documented in #738. @LGLO please confirm.

I got confirmation from Lech that this is indeed supposed to stay.

PR #745, opened 2025-05-06
doc: ETCM-9687 Domain crate documentation

AmbientTea:
# Description

Documents the domain crate

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/toolkit/sidechain/domain/src/byte_string.rs
+++ b/toolkit/sidechain/domain/src/byte_string.rs
@@ -1,9 +1,10 @@
-use core::fmt::{Debug, Display};
-use core::ops::Deref;
+//! Module providing various byte sequence wrapper types for use by the Partner Chains Toolkit crates
 
 use alloc::string::ToString;
 use alloc::vec::Vec;
 use byte_string_derive::byte_string;
+use core::fmt::{Debug, Display};
+use core::ops::Deref;
 use derive_where::derive_where;
 use parity_scale_codec::{Decode, DecodeWithMemTracking, Encode, MaxEncodedLen};
 use scale_info::TypeInfo;
@@ -14,8 +15,7 @@ use sp_core::Get;
 use sp_core::bounded::BoundedVec;
 
 /// Wrapper for bytes that is serialized as hex string
-/// To be used for binary data that we want to display nicely but
-/// don't have a specific type for
+/// To be used for binary data that we want to display nicely but don't have a specific type for
 #[derive(Eq, Clone, PartialEq, TypeInfo, Default, Encode, Decode, PartialOrd, Ord)]
 #[byte_string(debug)]
 #[cfg_attr(feature = "std", byte_string(to_hex_string, decode_hex))]
@@ -42,7 +42,7 @@ impl Deref for ByteString {
 	}
 }
 
-// Constant size variant of `ByteString` that's usable as a runtime type
+/// Constant size variant of `ByteString` that's usable in a FRAME runtime
 #[derive(Eq, Clone, PartialEq, TypeInfo, MaxEncodedLen, Encode, Decode, DecodeWithMemTracking)]
 #[byte_string(debug)]
 #[byte_string(to_hex_string)]
@@ -70,25 +70,18 @@ impl<const N: usize> Default for SizedByteString<N> {
 	}
 }
 
-/// Byte-encoded text string with bounded length
+/// Bounded-length bytes representing a UTF-8 text string
 #[derive(TypeInfo, Encode, Decode, DecodeWithMemTracking, MaxEncodedLen)]
 #[scale_info(skip_type_params(T))]
 #[derive_where(Clone, PartialEq, Eq, Default, PartialOrd, Ord)]
-pub struct BoundedString<T: Get<u32>>(pub BoundedVec<u8, T>);
+pub struct BoundedString<T: Get<u32>>(BoundedVec<u8, T>);
 
+/// Macro that creates a [BoundedString], supporting string interpolation like [alloc::format].
 #[macro_export]
 macro_rules! bounded_str {
     ($($arg:expr)+) => {
 		sidechain_domain::byte_string::BoundedString::try_from(format!($($arg)+).as_str()).unwrap()
     };
-}
-
-impl<T: Get<u32>> TryFrom<Vec<u8>> for BoundedString<T> {
-	type Error = <BoundedVec<u8, T> as TryFrom<Vec<u8>>>::Error;
-
-	fn try_from(value: Vec<u8>) -> Result<Self, Self::Error> {
-		Ok(Self(value.try_into()?))
-	}
 }
 
 impl<'a, T: Get<u32>> Deserialize<'a> for BoundedString<T> {
--- a/toolkit/sidechain/domain/src/crypto.rs
+++ b/toolkit/sidechain/domain/src/crypto.rs
@@ -1,3 +1,6 @@
+//! Cryptographic utilities for use by the domain types
+
+/// Computes a BLAKE2b hash of arbitrary static length for given `data`
 pub fn blake2b<const N: usize>(data: &[u8]) -> [u8; N] {
 	blake2b_simd::Params::new()
 		.hash_length(N)
@@ -17,16 +20,19 @@ mod full_crypto {
 	use plutus::to_datum_cbor_bytes;
 	use secp256k1::{Message, PublicKey, SecretKey, ecdsa::Signature};
 
+	/// Crates a 32 bytes (256 bits) BLAKE2b hash of the Plutus datum representation of `msg`
 	pub fn hash<T: ToDatum>(msg: T) -> [u8; 32] {
 		blake2b(to_datum_cbor_bytes(msg).as_slice())
 	}
 
+	/// Calculates the public key and signature for given private key and hashed data.
 	pub fn sc_public_key_and_signature(key: SecretKey, hashed: [u8; 32]) -> (PublicKey, Signature) {
 		let public_key = PublicKey::from_secret_key_global(&key);
 		let signature = key.sign_ecdsa(Message::from_digest_slice(hashed.as_slice()).unwrap());
 		(public_key, signature)
 	}
 
+	/// Signs the Plutus data bytes of `datum_msg` using `key` and returns it together with the verification key.
 	pub fn sc_public_key_and_signature_for_datum<T: ToDatum>(
 		key: SecretKey,
 		datum_msg: T,
@@ -35,6 +41,7 @@ mod full_crypto {
 		sc_public_key_and_signature(key, hashed_msg)
 	}
 
+	/// Signs the Plutus data bytes of `datum_msg` using `key` and returns it together with the verification key.
 	pub fn cardano_spo_public_key_and_signature<T: ToDatum>(
 		key: ed25519_zebra::SigningKey,
 		datum_msg: T,
--- a/toolkit/sidechain/domain/src/lib.rs
+++ b/toolkit/sidechain/domain/src/lib.rs
@@ -1,6 +1,9 @@
-//! Types/Structs/Functions meant to be used by other modules / our Business Logic (i.e the Sidechain's node / runtime in our case)
+//! # Partner Chains domain types
+//!
+//! This crate defines common domain and utility types used in the Partner Chain Toolkit.
 
 #![cfg_attr(not(feature = "std"), no_std)]
+#![deny(missing_docs)]
 
 pub mod byte_string;
 pub mod crypto;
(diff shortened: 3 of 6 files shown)


toolkit/sidechain/domain/src/mainchain_epoch.rs
```diff
@@ -1,60 +1,90 @@
...
 	pub slot_duration_millis: Duration,
+	/// Starting timestamp of the Shelley era on the Cardano chain
+	///
+	/// This timestamp should be equal to the starting timestamp of the Cardano epoch and slot
+	/// indicated by `first_epoch_number` and `first_slot_number`.
+	pub first_epoch_timestamp_millis: Timestamp,
+	/// Number of the Cardano Epoch in which the chain entered the Shelley era.
+	pub first_epoch_number: u32,
+	/// Slot number at which the chain entered the Shelley era.
+	pub first_slot_number: u64,
```
AmbientTea, line 32: Can someone confirm that this is indeed the Shelley epoch?

toolkit/sidechain/domain/src/byte_string.rs
```diff
@@ -70,12 +70,13 @@ impl<const N: usize> Default for SizedByteString<N> {
 	}
 }
 
-/// Byte-encoded text string with bounded length
+/// Bounded-length bytes representing a UTF-8 text string
```
Klapeyron, line 73: this may be not a UTF-8 string as we discussed

toolkit/sidechain/domain/src/lib.rs
```diff
@@ -1084,9 +1231,17 @@ mod tests {
...
+	/// Hash of a staking key representing a user delegator
 	StakeKeyHash([u8; 28]),
-	ScriptKeyHash { hash_raw: [u8; 28], script_hash: [u8; 28] },
+	/// Script hash representing a non-user delegator
+	ScriptKeyHash {
+		/// Raw script hash
+		hash_raw: [u8; 28],
+		/// Script hash
+		script_hash: [u8; 28],
+	},
```
AmbientTea, line 1244: I have honestly no idea what that really means so I summon @jankun4  and @kpinter-iohk

toolkit/sidechain/domain/src/byte_string.rs
AmbientTea: In what circumstances? We treat it as UTF-8 in the code so it should be that.

Klapeyron: https://github.com/input-output-hk/partner-chains/pull/622

toolkit/sidechain/domain/src/byte_string.rs
```diff
@@ -1,9 +1,10 @@
-use core::fmt::{Debug, Display};
-use core::ops::Deref;
+//! Module providing various binary data types for use by the Partner Chains Toolkit crates
```
LGLO, line 1: What a "binary" data type is? Definitions in this file define Rust types and also their serialization.

toolkit/sidechain/domain/src/byte_string.rs
```diff
@@ -42,7 +42,7 @@ impl Deref for ByteString {
 	}
 }
 
-// Constant size variant of `ByteString` that's usable as a runtime type
+/// Constant size variant of `ByteString` that's usable in a Substrate runtime
```
LGLO, line 45:
```suggestion
/// Constant size variant of `ByteString` that's usable in a FRAME runtime
```

toolkit/sidechain/domain/src/byte_string.rs
LGLO: The thing here is that not every BoundedVec bytes is a valid UTF-8 string. In other words one can create `BoundedString` that cannot be transformed to `String`.

PS Please comment on the line that is changed.

toolkit/sidechain/domain/src/crypto.rs
```diff
@@ -17,16 +20,19 @@ mod full_crypto {
 	use plutus::to_datum_cbor_bytes;
 	use secp256k1::{Message, PublicKey, SecretKey, ecdsa::Signature};
 
+	/// Crates a 32-bit BLAKE2b hash of the Plutus datum representation of `msg`
```
LGLO, line 23:
```suggestion
	/// Crates a 32 bytes (256 bits) BLAKE2b hash of the Plutus datum representation of `msg`
```

toolkit/sidechain/domain/src/mainchain_epoch.rs
LGLO: It is at least Shelley era. It could be some later Era.
There is also an assumption that neither `epoch_duration_millis` nor `slot_duration_millis` are changed in any Era after this slot.
It is "first" because it is the first we support. 

To put it differently, if there is a Cardano chain that has changed epoch duration or slot duration **after** entering Shelley era, then we still can support it from this point of time.

toolkit/sidechain/domain/src/mainchain_epoch.rs
```diff
@@ -1,60 +1,90 @@
...
 use parity_scale_codec::Encode;
 pub use sp_core::offchain::{Duration, Timestamp};
 
+/// Parameters describing epoch configuration of a Cardano chain
+///
+/// A Partner Chain must be aware of slot and epoch configuration of its own particular Cardano
+/// main chain to be able to correctly observe its state. Additionally, The Partner Chains
+/// Toolkit does not support basing a Partner Chain on Cardano chains that are still on legacy
+/// Cardano Eras. Because of that, a Partner Chain must be aware of when its Cardano main chain
+/// entered the Shelley era.
```
LGLO, line 15: Please add that this toolkit does not support Cardano chains that changed epoch or slot duration after entering Shelley.

toolkit/sidechain/domain/src/lib.rs
kpinter-iohk: These come from [`stake_address`](https://github.com/IntersectMBO/cardano-db-sync/blob/master/doc/schema.md#stake_address) from DB sync.
* hash_raw: The raw bytes of the stake address hash.
* script_hash: The script hash, in case this address is locked by a script.
(Obviously there is logic so ScriptKeyHash always has script_hash)

toolkit/sidechain/domain/src/byte_string.rs
AmbientTea: Okay, I've made the inner `BoundedVec` and left only casts that ensure that the content is a valid UTF-8 string

toolkit/sidechain/domain/src/mainchain_epoch.rs
AmbientTea: very important, thank you

toolkit/sidechain/domain/src/byte_string.rs
Klapeyron: there is still an option to workaround it via `Decode` trait, but implementing custom `Decode` that will perform on-chain validation has performance overhead. We can replace `url` with something else, like `foot length` of the validator and stop using of `BoundedString` at all

toolkit/sidechain/domain/src/mainchain_epoch.rs
AmbientTea: I heavily reworded the whole documentation for this type, please take another look

toolkit/sidechain/domain/src/lib.rs
AmbientTea: Thank you, I updated all comments on this type

toolkit/sidechain/domain/src/byte_string.rs
Klapeyron: one place I see is governed map:
`	pub type MapKey<T> = BoundedString<<T as Config>::MaxKeyLength>;`
but do we have a requirement that a key needs to be a utf-8 encoded string? What about keeping it as a BoundedVec?

Right, it is part of the demo node as we discussed, but type itself is not aware of the validation, so can be accidentally misused, if we do not want to remove it, then we should ensure that it is always under constraints.
But if there is an option to review once again places that `BoundedString` is used and check if it is indeed needed, I think that will be better case. IMO if code is not making any opinions regarding what is inside this blob of bytes then better.

toolkit/sidechain/domain/src/lib.rs
```diff
@@ -655,12 +711,22 @@ pub struct EpochNonce(pub Vec<u8>);
 	MaxEncodedLen,
 	Hash,
 )]
+/// Identifies a Cardano UTxO (unspent transaction output)
+///
+/// A UTxO is uniquely identified by the hash of the transaction that produced it and its (zero-based)
+/// index in the transaction's output.
+///
+/// Standard semi-human-readable encoding of a UTxO id uses a hash sign to divide the two components:
+/// `0000000000000000000000000000000000000000000000000000000000000000#1`
```
kpinter-iohk, line 720:
```suggestion
/// `0000000000000000000000000000000000000000000000000000000000000000#0`
```
Would it make sense to use 0 as the ix to emphasize it's zero-based? Or does it make it even harder to find the `#`?

toolkit/sidechain/domain/src/lib.rs
```diff
@@ -797,27 +870,41 @@ impl Display for McBlockHash {
 	}
 }
 
+/// Extended information about an (unspent) transaction output in Cardano ledger
```
kpinter-iohk, line 873:
```suggestion
/// Extended information about a UTxO in Cardano ledger
```
UTxOs are explained above and should be assumed to be known by anyone using this library as it is a fundamental term.

toolkit/sidechain/domain/src/lib.rs
```diff
@@ -797,27 +870,41 @@ impl Display for McBlockHash {
...
+	/// Slot number in which the output was produced
 	pub slot_number: McSlotNumber,
+	/// Index in block of the transaction that produced the output
 	pub tx_index_within_block: McTxIndexInBlock,
 }
 
+/// Key type used for ordering transaction outputs
+///
+/// This ordering key is used in contexts where a common ordering of the data must be used
+/// by all nodes participating in a Partner Chain due to it being subject to consensus.
```
kpinter-iohk, line 895:
```suggestion
/// by all nodes participating in a partner chain due to it being subject to consensus.
```
I see we capitalize partner chain often but why? Isn't it a common noun?

toolkit/sidechain/domain/src/lib.rs
```diff
@@ -763,6 +830,9 @@ pub const TX_HASH_SIZE: usize = 32;
 )]
 #[byte_string(debug, from_bytes, decode_hex, hex_serialize, hex_deserialize)]
 #[constructor_datum]
+/// Cardano transaction hash
+///
+/// This hash uniquely identifies a transaction in the Cardano ledger.
```
kpinter-iohk, line 835: I think for computer ledgers it's "on the ledger". I've written about my reasoning [here](https://github.com/input-output-hk/partner-chains/pull/702#discussion_r2055888530). Would be nice to figure out which one it is.

toolkit/sidechain/domain/src/lib.rs
```diff
@@ -942,12 +1059,22 @@ impl From<ed25519::Public> for GrandpaPublicKey {
...
+/// selected into a Partner Chain committee, used by the Ariadne family of selection algorithms.
+///
+/// The core idea behind the D-Param is to enable a Partner Chain to bootstrap its operation by
+/// relying on a hand-picked set of trusted block producers for security, and to later incrementally
+/// shift block production onto trustless network participants as the chain grows and it becomes
+/// harder for malicious actors to manipulate the chain.
 pub struct DParameter {
+	/// Expected number of permissioned candidates selected to a committee
 	pub num_permissioned_candidates: u16,
+	/// Expected number of registered candidates selected to a committee
```
kpinter-iohk, line 1072:
```suggestion
	/// Expected number of permissioned candidates selected for a committee
	pub num_permissioned_candidates: u16,
	/// Expected number of registered candidates selected for a committee
```

toolkit/sidechain/domain/src/lib.rs
```diff
@@ -1113,11 +1269,18 @@ impl<T: Into<u64>> From<T> for DelegatorStakeAmount {
...
+///
+/// This mapping can be used to calculate relative share of the total delegation for the
+/// purpose of weighing during block producer selection.
 #[derive(Debug, Clone, Default)]
 pub struct StakeDistribution(pub BTreeMap<MainchainKeyHash, PoolDelegation>);
 
+/// ADA delegation data for a single Cardano SPO
 #[derive(Debug, Clone, Default, PartialEq)]
 pub struct PoolDelegation {
+	/// Total stake to the stake pool
```
kpinter-iohk, line 1282:
```suggestion
	/// Total amount delegated to the stake pool
```

kpinter-iohk (approved):

toolkit/sidechain/domain/src/lib.rs
gilligan: that's the preferred way of writing - I was correct on that before as well :)

toolkit/sidechain/domain/src/lib.rs
gilligan: "in the Cardano ledger" implies the transaction is **contained within** the ledger's data which makes sense. The ledger is a record of all transactions, each transaction exists _inside_ it. I would say let's keep it like this

gilligan (approved):

toolkit/sidechain/domain/src/lib.rs
```diff
@@ -221,6 +236,7 @@ impl FromStr for McTxIndexInBlock {
 	}
 }
 
+/// Maximum length of a Cardano address in UTF-8 bytes
```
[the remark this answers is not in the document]
AmbientTea: it's utf8 bytes of the address string in our case here

PR #749, opened 2025-05-07
change: ETCM-9816 prevent calling toolkit pallets' inherents twice in a block

AmbientTea:
# Description

This PR patches four of our pallets that didn't previously have logic that would prevent running the same (valid) inherent multiple times in the same block. An inherent having to run only once is not something that Substrate by itself enforces, as notes by @Klapeyron , so we need to guard against that. I added logic that prevents that to each pallet. I tried to do this in a way that makes sense in their contexts, instead of just adding a 'has run' flag storage everywhere.

I also squeezed some rust doc work into this PR in places where I touched code that was not properly documented.

# Checklist

- [ ] Self-reviewed the diff


toolkit/block-participation/pallet/src/lib.rs
```diff
@@ -109,6 +109,16 @@ pub mod pallet {
...
 
+	#[pallet::error]
+	pub enum Error<T> {
+		/// Indicates an attempt to process block participation data for already processed slot
+		UpToSlotNotIncreased,
+	}
+
+	/// Stores the slot number up to which block participation has already been processed
+	#[pallet::storage]
+	pub type ProcessedUpToSlot<T: Config> = StorageValue<_, Slot, ValueQuery>;
```
AmbientTea, line 120: the pallet guards against replays by requiring `up_to_slot` in each inherent invoked to increase

toolkit/committee-selection/pallet/src/lib.rs
```diff
@@ -278,6 +278,7 @@ pub mod pallet {
 			ensure_none(origin)?;
 			let expected_epoch_number = CurrentCommittee::<T>::get().epoch + One::one();
 			ensure!(for_epoch_number == expected_epoch_number, Error::<T>::InvalidEpoch);
+			ensure!(!NextCommittee::<T>::exists(), Error::<T>::NextCommitteeAlreadySet);
```
AmbientTea, line 281: Here we just check if the storage has not been set yet. The `create_inherent` function of the pallet already produces an inherent only when this storage is empty.

toolkit/native-token-management/pallet/src/lib.rs
```diff
@@ -51,31 +53,61 @@ pub mod pallet {
...
+		/// Event handler for incoming native token transfers
 		type TokenTransferHandler: TokenTransferHandler;
+
+		/// Weight information for this pallet's extrinsics
 		type WeightInfo: WeightInfo;
 	}
 
+	/// Events emitted by this pallet
 	#[pallet::event]
+	#[pallet::generate_deposit(pub(super) fn deposit_event)]
```
AmbientTea, line 71: It turns out we didn't even emit this event. I fixed that, but I can also just get rid of it completely.

toolkit/native-token-management/pallet/src/lib.rs
```diff
@@ -51,31 +53,61 @@ pub mod pallet {
...
+	/// The pallet is considered initialized if its inherent has been successfuly called at least once since
+	/// genesis or the last invocation of [set_main_chain_scripts][Pallet::set_main_chain_scripts].
 	#[pallet::storage]
 	pub type Initialized<T: Config> = StorageValue<_, bool, ValueQuery>;
 
+	/// Transient storage containing the amount of native token transfer registered in the current block.
+	///
+	/// Any value in this storage is only present during execution of a block and is wiped on block finalization.
+	#[pallet::storage]
+	pub type TransferedThisBlock<T: Config> = StorageValue<_, NativeTokenAmount, OptionQuery>;
```
AmbientTea, line 102: I save the number of tokens transferred this block so this storage has some potential usefulness aside from just being a set flag

toolkit/governed-map/pallet/src/lib.rs
```diff
@@ -351,6 +362,12 @@ pub mod pallet {
 		#[pallet::weight((T::WeightInfo::register_changes(changes.len() as u32), DispatchClass::Mandatory))]
 		pub fn register_changes(origin: OriginFor<T>, changes: Changes<T>) -> DispatchResult {
 			ensure_none(origin)?;
+			let current_block = frame_system::Pallet::<T>::block_number();
+			ensure!(
+				LastUpdateBlock::<T>::get().map_or(true, |last_block| last_block < current_block),
+				Error::<T>::InherentCalledTwice
+			);
+			LastUpdateBlock::<T>::put(current_block);
```
LGLO, line 370: Recalculate weights, please

toolkit/native-token-management/pallet/src/lib.rs
```diff
@@ -185,10 +225,19 @@ pub mod pallet {
 		}
 	}
 
+	#[pallet::hooks]
+	impl<T: Config> Hooks<BlockNumberFor<T>> for Pallet<T> {
+		fn on_finalize(_block: BlockNumberFor<T>) {
+			TransferedThisBlock::<T>::set(None);
```
LGLO, line 231:
```suggestion
			TransferedThisBlock::<T>::kill();
```
Would it work?

IDK why is it called `kill` instead of `empty` or `clear`.

LGLO (approved):

toolkit/native-token-management/pallet/src/lib.rs
AmbientTea: it does, ty

PR #755, opened 2025-05-08
tests: added GovernedMap observability tests

rsporny:
Refs: ETCM-8633

### DIFF
--- a/e2e-tests/src/substrate_api.py
+++ b/e2e-tests/src/substrate_api.py
@@ -695,9 +695,9 @@ class SubstrateApi(BlockchainApi):
         initial_epoch = epoch_result.value - session_index_result.value
         return initial_epoch
 
-    def get_governed_map_main_chain_scripts(self):
-        result = self.substrate.query("GovernedMap", "MainChainScripts")
-        logger.debug(f"Governed map main chain scripts: {result}")
+    def get_governed_map_initialization_state(self):
+        result = self.substrate.query("GovernedMap", "Initialized")
+        logger.debug(f"Governed map initialization state: {result}")
         return result.value
 
     @long_running_function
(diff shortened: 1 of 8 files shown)


rsporny:
TODO: skip on preview testnet

e2e-tests/src/substrate_api.py
```diff
@@ -694,3 +694,93 @@ def get_initial_pc_epoch(self):
...
+
+    def subscribe_governed_map_change(self, key=None, key_value=None):
+        current_main_chain_block = self.get_mc_block()
+        max_main_chain_block = current_main_chain_block + self.config.main_chain.security_param
+
+        def subscribed_change_handler(registered_changes):
+            if key_value:
+                return next((change for change in registered_changes if key_value in registered_changes), None)
+            elif key:
+                return next((change for change in registered_changes if change[0] == key), None)
```
AmbientTea, line 744: why use iterator here? Isn't
```python
next((change for change in registered_changes if key_value in registered_changes), None)
```
equivalent of
```python
registered_changes[0] if registered_changes and key_value in registered_changes else None
```
?

e2e-tests/src/substrate_api.py
```diff
@@ -694,3 +694,93 @@ def get_initial_pc_epoch(self):
...
+            else:
+                if registered_changes:
+                    return registered_changes
+                else:
+                    return True  # e.g. governed map reinitialization with 0 changes
```
AmbientTea, line 749:
```suggestion
            elif registered_changes:
                return registered_changes
            else:
                return True  # e.g. governed map reinitialization with 0 changes
```

e2e-tests/src/substrate_api.py
```diff
@@ -694,3 +694,93 @@ def get_initial_pc_epoch(self):
...
+                logger.debug(f"# {idx}: {extrinsic.value}")
+                if (
+                    extrinsic.value["call"]["call_module"] == "GovernedMap"
+                    and extrinsic.value["call"]["call_function"] == "register_changes"
+                ):
+                    registered_changes = extrinsic.value["call"]["call_args"][0]["value"]
+                    subscribed_change = subscribed_change_handler(registered_changes)
+                    break
+            if subscribed_change:
+                self.substrate.rpc_request("chain_unsubscribeNewHeads", [subscription_id])
```
AmbientTea, line 767: TIL 👍

AmbientTea (approved):

e2e-tests/src/substrate_api.py
rsporny: it is, and it's wrong, I wanted it to return searched key_value 🫤

PR #765, opened 2025-05-12
docs: Add section on running pc dependencies

gilligan:
# Description

Small addition to docs/intro.md providing info on pc runtime dependencies. I try to make small incremental improvements to the docs. Please have a look. The file should gradually replace the guides we have. 

The idea is to provide some relevant information on those externals but mostly refer to the respective existing documentation. We cannot explain details of how to run cardano-node or db-sync.

**But** if there are any particularities about how cardano-node needs to be run or any way in which db-sync needs to be configured that is out-of-the-ordinary and specific to PC it would be great to add it here. Please comment if you can think of something :)   

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/docs/intro.md
+++ b/docs/intro.md
@@ -214,7 +214,7 @@ indexer persists ledger state and events in the database. The partner chain node
 access the database directly.
 
 :warning: Make sure to create a database called `cexplorer` and make it accessible to the user which
-will execute your Partner Chain node executable.
+will execute your partner chain node executable.
 
 
 #### Running ogmios
--- a/docs/intro.md
+++ b/docs/intro.md
@@ -189,7 +189,7 @@ hardware characteristics but below are some approximations:
 | mainnet       | ~2 days|
 
 #### Running db-sync
-In order to observe the state of the Cardano ledger, the Partner Chain relies on
+In order to observe the state of the Cardano ledger, the partner chain relies on
 [db-sync](https://github.com/IntersectMBO/cardano-db-sync) as chain-indexer. The default
 configuration is sufficient. Make sure to refer to the project
 [documentation](https://github.com/IntersectMBO/cardano-db-sync/blob/master/doc/Readme.md) for
--- a/docs/intro.md
+++ b/docs/intro.md
@@ -210,7 +210,7 @@ depends on the network conditions and hardware characteristics but below are som
 #### Running PostgreSQL
 
 A [PostgreSQL](https://www.postgresql.org/) database is a runtime requirement via db-sync as the
-indexer persists ledger state and events in the database. The Partner Chain node also needs to
+indexer persists ledger state and events in the database. The partner chain node also needs to
 access the database directly.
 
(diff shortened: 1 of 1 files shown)


docs/intro.md
```diff
@@ -160,6 +166,67 @@ Finally, many features expose **Cli Commands** that support their operation. The
...
+| mainnet       | ~2 days|
+
+
+#### Running PostgreSQL
+
+A [PostgreSQL](https://www.postgresql.org/) database is a runtime requirement via db-sync as the
+indexer persists ledger state and events in the database. The Partner Chain node also needs to
+access the database directly.
+
+:warning: Make sure to create a database called `cexplorer` and make it accessible to the user which
```
gilligan, line 216: @skylar-simoncelli if we find out that this isn't hardcoded at all then I should perhaps I should adjust this section to reflect that and mention the environment variable instead.

docs/intro.md
```diff
@@ -160,6 +166,67 @@ Finally, many features expose **Cli Commands** that support their operation. The
...
+website](https://github.com/IntersectMBO/cardano-node) for detailed instructions on configuring and
+running cardano-node.
+
+:warning: Please note that your cardano-node instance needs to be fully synchronized before you can
+start to crate a Partner Chain setup. The synchronization time depends on the network conditions and
+hardware characteristics but below are some approximations:
+
+| network  | approximated sync time |
+| ------------- | ------------- |
+| preview       | hours  |
```
gilligan, line 187: I took these timings from the chain builder guide. They seemed.. reasonable??

nstanford5 (changes requested):

gilligan:
@nstanford5 d'oh sorry i somehow made an effort to consistently use the wrong capitalization for partner chains 😂

docs/intro.md
LGLO: It isn't hardcoded. It is default.

docs/intro.md
```diff
@@ -160,6 +166,67 @@ Finally, many features expose **Cli Commands** that support their operation. The
...
+access the database directly.
+
+:warning: Make sure to create a database called `cexplorer` and make it accessible to the user which
+will execute your Partner Chain node executable.
+
+
+#### Running ogmios
+[ogmios](https://github.com/CardanoSolutions/ogmios) is a lightweight bridge-interface providing a
+http/websocket interface for communicating with a local cardano node. Unlike _db-sync_, which is
+always mandatory, ogmios is **only required by the chain builder** when interacting with the Partner
```
LGLO, line 223: An SPO who wants to register needs it as well.

docs/intro.md
gilligan: Yes, sorry.. indeed. I don't know why I chose to say "hardcoded". I did mean "default"

docs/intro.md
gilligan: ah, right!! good catch. thanks, i'll adjust the text

LGLO (approved):

PR #770, opened 2025-05-13
docs: add C4 architectural diagrams

LGLO:
# Description

Another take on Architecture diagrams


AmbientTea (approved):
i like this, even though the C4 graphical style is hard to read

gilligan (approved):
Looks good to me - succinct.

PR #771, opened 2025-05-13
docs: add configuration section

gilligan:
# Description

This adds a Configuration section to intro.md

**Note**: This is really very much a DRAFT in that it requires quite some additional support filling out the blanks / TODOs. As per usual the goal is not perfection but ending up with something better than what we have now.

The goal is to add a **Configuration** section that provides info on:
1. All fields in chain-spec.json and their meaning
2. All ENV processed by the PC node
3. The keys used by the node

I used all keys found in `dev/envs/devnet/.envrc` as a starting point for the variables to consider. I also verified that all variables in there are actually being read by our code somewhere (see bash hackery below).

----

```bash
$ for e in $(cat dev/envs/devnet/.envrc \
| grep export \
| sed -n 's/^export \([A-Z_][A-Z0-9_]*\)=.*/\1/p'); \
do echo -e "\n\n\n\n$e\n\n\n\n" && git --no-pager  grep  $e ; done |less
```

# Checklist

- [ ] Self-reviewed the diff


docs/intro.md
```diff
@@ -227,6 +231,107 @@ Please refer to the [project
 documentation](https://ogmios.dev/getting-started/building/#-documentation) for details on how to
 run and configure ogmios.
 
+### Configuration
```
gilligan, line 234: I am aware that there are some additional config files of relevance when dealing with wizards, but I wanted to keep this those out of scope here. They can (and maybe should?) be documented elsewhere in sections about the respective wizards

docs/intro.md
```diff
@@ -227,6 +231,107 @@ Please refer to the [project
...
+Partner chains are configured through a combination of configuration files and environment
+variables. On top of the files outlined below, the command line wizard provides a convenient layer
+for initializing configurations.
+
+#### `chain-spec.json`
+The `chain-spec.json` file combines generic Substrate configuration values with partner chain
+specific settings. All partner chain specific settings are stored below
+`.genesis.runtimeGenesis.config`:
+
+##### `aura`
```
LGLO, line 244: > 1. All fields in chain-spec.json and their meaning
and
> All partner chain specific settings are stored below

Why? Grandpa is not our business. Aura isn't strict requirement as well. `sudo` ? I think we don't have to document anything that is not coming from our toolkit.

docs/intro.md
```diff
@@ -227,6 +231,107 @@ Please refer to the [project
...
+TODO
+##### `PERMISSIONED_CANDIDATES_POLICY_ID`
+TODO
+##### `NATIVE_TOKEN_POLICY_ID`
+TODO
+##### `NATIVE_TOKEN_ASSET_NAME`
+TODO
+##### `ILLIQUID_SUPPLY_VALIDATOR_ADDRESS`
+TODO
+##### `CARDANO_SECURITY_PARAMETER`
```
LGLO, line 304: It has to be set to save value that Cardano is using. Copy it from shelley-genesis `securityParam`.

docs/intro.md
```diff
@@ -227,6 +231,107 @@ Please refer to the [project
...
+TODO
+##### `CARDANO_ACTIVE_SLOTS_COEFF`
+TODO
+##### `DB_SYNC_POSTGRES_CONNECTION_STRING`
+TODO
+##### `BLOCK_STABILITY_MARGIN`
+TODO
+##### `MC__FIRST_EPOCH_TIMESTAMP_MILLIS`
```
LGLO, line 312: Timestamp in milliseconds of the first Shelley era epoch of main chain

docs/intro.md
```diff
@@ -227,6 +231,107 @@ Please refer to the [project
...
+TODO
+##### `MC__SLOT_DURATION_MILLIS`
+TODO
+##### `MC__FIRST_EPOCH_NUMBER`
```
LGLO, line 316: Number of the first Shelley era main chain epoch

docs/intro.md
```diff
@@ -227,6 +231,107 @@ Please refer to the [project
...
+##### `transactionPayment`
+##### `transactionPayment.multiplier`
+
+
+
+#### Environment Variables
+The following environment values will be read by the partner chain node process.
+##### `GENESIS_UTXO`
+TODO
+##### `SIDECHAIN_BLOCK_BENEFICIARY`
```
LGLO, line 290: obsolete #775

docs/intro.md
```diff
@@ -227,6 +231,107 @@ Please refer to the [project
...
+TODO
+##### `CARDANO_ACTIVE_SLOTS_COEFF`
+TODO
+##### `DB_SYNC_POSTGRES_CONNECTION_STRING`
+TODO
+##### `BLOCK_STABILITY_MARGIN`
+TODO
+##### `MC__FIRST_EPOCH_TIMESTAMP_MILLIS`
+TODO
+##### `MC__SLOT_DURATION_MILLIS`
```
LGLO, line 314: Cardano slot duration in milliseconds. Public networks use 1000ms slots. Defaults to 1000ms if not set.

docs/intro.md
```diff
@@ -227,6 +231,107 @@ Please refer to the [project
...
+TODO
+##### `MC__FIRST_EPOCH_NUMBER`
+TODO
+##### `MC__EPOCH_DURATION_MILLIS`
+TODO
+##### `MC__FIRST_SLOT_NUMBER`
```
LGLO, line 320: Number of the first slot of the first epoch of Shelley era

docs/intro.md
```diff
@@ -227,6 +231,107 @@ Please refer to the [project
...
+TODO
+##### `BLOCK_STABILITY_MARGIN`
+TODO
+##### `MC__FIRST_EPOCH_TIMESTAMP_MILLIS`
+TODO
+##### `MC__SLOT_DURATION_MILLIS`
+TODO
+##### `MC__FIRST_EPOCH_NUMBER`
+TODO
+##### `MC__EPOCH_DURATION_MILLIS`
```
LGLO, line 318: Duration of Cardano epoch in millis

docs/intro.md
```diff
@@ -227,6 +231,107 @@ Please refer to the [project
...
+TODO
+##### `MC__FIRST_SLOT_NUMBER`
+TODO
+##### `USE_MAIN_CHAIN_FOLLOWER_MOCK`
```
LGLO, line 322: Usage: `unset USE_MAIN_CHAIN_FOLLOWER_MOCK`
Thank you.

docs/intro.md
```diff
@@ -227,6 +231,303 @@ Please refer to the [project
...
+        }
+      ]
+    ]
+  }
+}
+```
+
+Keep in mind that for compatibility with Grandpa, a Partner Chain needs to also use
+the stock Session pallet provided by Substrate. As this pallet does is no longer
+responsible for managing session committee anymore, its genesis configuration
```
gilligan, line 336:
```suggestion
the stock Session pallet provided by Substrate. As this pallet is no longer
responsible for managing session committee, its genesis configuration
```

docs/intro.md
```diff
@@ -227,6 +231,107 @@ Please refer to the [project
...
+##### `sudo.key`
+##### `testHelperPallet`
+##### `testHelperPallet.participationDataReleasePeriod`
+##### `testHelperPallet.phantom`
+##### `transactionPayment`
+##### `transactionPayment.multiplier`
+
+
+
+#### Environment Variables
```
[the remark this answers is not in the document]
LGLO: Perhaps not, because to run the node I source the env file and run it.

docs/intro.md
```diff
@@ -227,6 +231,303 @@ Please refer to the [project
...
+  of the Shelley era, but for chains that modified their slot or epoch duration
+  after that, the start of any Cardano epoch after the last change could be used,
+  as long as it is before any Partner Chains smart contracts have been deployed
+
+Lastly, every Partner Chain node needs the `BLOCK_STABILITY_MARGIN` variable to
+be set. This variable determines an additional margin added to
+`CARDANO_SECURITY_PARAMETER` when selecting the latest stable Cardano block to be
+included in the block header. This value can be set to `0` by default, but higher
+values if the network experiences a high number of blocks rejected because of
+Db-Sync lag.
```
LGLO, line 500: No, `BLOCK_STABILITY_MARGIN` is not needed. 0 is used when it is absent AFAIK.

Also, it is documented in some other place that using anything other than 0 is detrimental in v1.6.1 and before.

I would write that 0 or 1 are endorsed values. If block producer uses 1 and his block is rejected it means that other validators have serious problems with following Cardano.

docs/intro.md
```diff
@@ -227,6 +231,303 @@ Please refer to the [project
...
+- `ILLIQUID_SUPPLY_VALIDATOR_ADDRESS`
+
+#### Keys
+The partner chain node needs three different keys to be present in a Substrate key store accessible
+by the partner chain node.
+1. `sidechain_pub_key`: The sidechain public key (ECDSA)
+1. `aura_pub_key`: The aura public key (sr25519)
+1. `grandpa_pub_key`: The grandpa public key (Ed25519)
+
+By default the partner chain node process will look for key stores in the base path directory.
```
LGLO, line 530: Not really. This is not our toolkit business. As far as I'm concerned logic of resolving keystore location is hardcoded in substrate and it looks in `<base_dir>/chains/<chain_id>`
where `chain_id` is taken from the used chain spec.

docs/intro.md
```diff
@@ -227,6 +231,303 @@ Please refer to the [project
...
+    }
+  }
+}
+```
+
+#### Environment Variables
+
+Some of the configuration values used by the Partner Chain toolkit components are
+read from the local environment. These generally come in two categories:
+operational data used by the data sources; and some of the genesis configuration
```
gilligan, line 458: Can you rephrase this a bit? The sentence describing the second category is a bit long-winded. I also like using enumerations when a text introduces two distinct things to differentiate between:

>These generally come in two categories:
>1. such and such
>2. blabla

docs/intro.md
AmbientTea: `BLOCK_STABILITY_MARGIN` does not have a default. The node fails if it's not present.

I updated the description to advise use of 1 only

docs/intro.md
AmbientTea: reworded

docs/intro.md
LGLO: Isn't it too early for this? Before we release the next version?

LGLO (approved):

PR #772, opened 2025-05-13
change: Sidechain primitives crates cleanup

AmbientTea:
# Description

- Moves the logic in `query` module out of the crate completely. It contained a json-string-returning function used on only by a CLI command, which the primitives should not know anything about, and an output struct shared between rpc and cli, which they should not share
- Deprecates the `GetSidechainStatus` runtime API and its returned type `SidechainStatus` (and removes some unnecessary uses of it). This API is mis-named (doesn't serve any status), returns a structure out of which each user code needs only one field, and one field is not used at all. Ideally, each functionality should define the runtime API it consumes instead of sharing one with other unrelated functionalities.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/toolkit/cli/commands/src/lib.rs
+++ b/toolkit/cli/commands/src/lib.rs
@@ -1,4 +1,5 @@
 pub mod address_association_signatures;
 pub mod block_producer_metadata_signatures;
+pub mod get_genesis_utxo;
 pub mod key_params;
 pub mod registration_signatures;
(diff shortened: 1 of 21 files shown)


toolkit/cli/commands/src/lib.rs
```diff
@@ -1,4 +1,5 @@
 pub mod address_association_signatures;
 pub mod block_producer_metadata_signatures;
+pub mod get_genesis_utxo;
```
LGLO, line 3: I can't see the file.

kpinter-iohk (approved):

PR #777, opened 2025-05-14
feat: change default multisig threshold to 1

rsporny:


toolkit/partner-chains-cli/src/config.rs
```diff
@@ -630,7 +630,7 @@ pub mod config_fields {
 			config_file: CHAIN_CONFIG_FILE_PATH,
 			path: &["initial_governance", "threshold"],
 			name: "Initial Multisig Governance Threshold",
-			default: Some("0"),
+			default: Some("1"),
```
LGLO, line 633: I was thinking about it when doing recent changes for default governance value and decided against it.

Setting it to 1 makes it easier to accidently set to 1 when real multisig is used.

If users accidently enters 0, then wizards will fail. If user sets multiple GA key hashes and accidently enters 1, there will be "1 of N" governance set on Cardano.

I only explain my perspective on having 1 instead of 0. I'm happy to accept opinion of others.

rsporny: Let's see what others have to say, but sure, it makes sense.

AmbientTea: I don't have a strong opinion here, but accidentally setting 1 of N which can easily be changed is not the end of the world. On the other hand, it's good to prevent that if we can, but defaulting to 0 is kinda hacky. Maybe we shouldn't have a default at all?

gilligan: I literally just walked into this while working on documentation
```
> Initial Multisig Governance Threshold 0
Error: Application(Initial Goveranance data is invalid: 'threshold has to be a positive number'. Please r
un the wizard again and provide correct value or edit values in 'pc-chain-config.json' and the run the w$
zard again. Example: '{"initial_governance":{"authorities":["0x000000000000000000000000000000000000000000
00000000000000","0x01010101010101010101010101010101010101010101010101010101","0x0202020202020202020202020
2020202020202020202020202020202"],"threshold":2}}')
```

 IMHO the two valid options for this are:
 
 1. Setting the default to 1
 2. Providing no default at all and forcing the user to specify a value. Note that the whole wizard shouldn't fail if the user enters 0. It should just keep on requesting input until the user provides a non-zero value

gilligan (approved):

PR #787, opened 2025-05-16
add: helper crate for wiring Partner Chain nodes

AmbientTea:
# Description

I'm leaving this PR as draft to first have a discussion if this is something we want.

* Moves logic wiring data sources and RPCs out of our demo node into a new `partner-chains-node` crate. This crate provides:
    - `PartnerChainDataSource` - a wrapper around all our data sources and time source
    - `PartnerChainNodeConfig` - wraps the configuration that we read from environment + makes the logic to read it accessible to users
    - `partner_chain_rpc!` - macro that adds our RPCs into the rpc module. accepts the data source + config and hides the wiring inside
  The crate provides data source and RPC for the sidechain and mc-hash functionality, and hides the rest behind feature flags. The idea is that after this crate is wired into the node, enabling a new feature should more or less work out of the box without additional integration by the node developer
* Removes the "unit tests" from the demo node. These tests weren't really useful and forced us to expose mocks from other crates to implement them, and forces the node to depend on internal data sources.

Unfortunately feature flags can't be applied to bounds in `where` clauses, which means that it's not possible to define helper traits that would free the node from needing to specify exactly which runtime Apis etc. are needed from each type, ie. I couldn't get this to work:
```
fn create_rpc<C>(..)
where
    #[cfg(feature = "block-participation")]
    C: BlockParticipationApi
    #[cfg(feature = "governed-map")]
    C: GovernedMapApi
{...}
```

# Checklist

- [ ] Self-reviewed the diff


demo/node/src/inherent_data.rs
AmbientTea, line 1: I have some ideas how to move wiring the IDP out of here too, but I don't want this draft to be too big.

toolkit/node/src/data_source.rs
```diff
@@ -0,0 +1,146 @@
...
+	pub async fn new_db_sync_from_env(
+		metrics_registry_opt: Option<&Registry>,
+		candidates_for_epoch_cache_size: usize,
+		#[cfg(feature = "governed-map")] governed_map_cache_size: u16,
+		#[cfg(feature = "block-participation")] stake_cache_size: usize,
+	) -> Result<Self, Box<dyn std::error::Error + Send + Sync + 'static>> {
+		let metrics_opt = register_metrics_warn_errors(metrics_registry_opt);
+		let pool =
+			partner_chains_db_sync_data_sources::data_sources::get_connection_from_env().await?;
+		// block data source is reused between mc_hash and sidechain_rpc to share cache
```
LGLO, line 106: and governed map

toolkit/node/src/rpc.rs
```diff
@@ -0,0 +1,99 @@
...
+	B: BlockT,
+	Metadata: Clone + Decode + Serialize + Send + Sync + 'static,
+	AccountId: Clone + Decode + Serialize + Send + Sync + 'static,
+	C: Send + Sync + 'static,
+	C: ProvideRuntimeApi<B>,
+	C: HeaderBackend<B>,
+	C::Api: BlockProducerMetadataApi<B, Metadata>,
+	C::Api: BlockProducerFeesApi<B, AccountId>,
+{
+	module.merge(BlockProducerFeesRpc::new(client.clone()).into_rpc())?;
```
LGLO, line 85: BlockProducerFees are separate to block participation.

PR #789, opened 2025-05-16
Extended error description in case node does not have enough history

Klapeyron:
# Description

Slightly more descriptive error message in case that `sidechain_getEpochCommittee` is executed on non-archive node.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/toolkit/committee-selection/query/src/lib.rs
+++ b/toolkit/committee-selection/query/src/lib.rs
@@ -163,7 +163,9 @@ where
 		let first_epoch = {
 			let second_block = (self.client)
 				.hash(1.into())
-				.map_err(err_debug)?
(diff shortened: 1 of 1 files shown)


toolkit/committee-selection/query/src/lib.rs
```diff
@@ -163,7 +163,9 @@ where
 		let first_epoch = {
 			let second_block = (self.client)
 				.hash(1.into())
-				.map_err(err_debug)?
+				.map_err(|err| {
+					format!("Node is not in archive mode, not able to fetch first block: {err:?}")
```
chrispalaskas, line 167: Node is not in archive mode therefore not able to fetch first block:
(just a suggestion, no fret)

PR #806, opened 2025-05-21
Inject GovernedMap genesis config to chain-spec file, remove building preprod chain-spec

LGLO:
# Injecting values into chain-spec file

Currently, we tend to:
* set env variables
* Have a code that uses these env vars in code that builds `RuntimeGenesisConfig`
* Wire this code to be used by `<node-executable> chain-spec` command.

At the same time we do not set all of the fields, requiring some to be externally injected into the chain-spec JSON file.

Example:
```rust
session_committee_management: SessionCommitteeManagementConfig {
    initial_authorities: vec![],
    main_chain_scripts: sp_session_validator_management::MainChainScripts::read_from_env()?,
},
```

In this PR I propose to not rely on env variables in the Rust code. `chain-spec` command would build only empty/default values.
CI and local-env would use shell tooling to update chain-spec.json files instead.

# Removal of staging-preprod chain-spec

I've checked that `dev/envs/staging-preprod` files were using GENESIS_UTXO of a transaction that was never executed on Preprod and it was using Preview Cardano configuration. In other words this setup was never correct. It is not used currently.
`kubectl get pods -n staging-preprod` shows this env is not running.


dev/envs/ci-preview/.envrc
```diff
@@ -1,36 +1,19 @@
...
 export D_PARAMETER_POLICY_ID=$(jq -r '.policyIds.DParameter' dev/envs/ci-preview/addresses.json)
 export PERMISSIONED_CANDIDATES_POLICY_ID=$(jq -r '.policyIds.PermissionedCandidates' dev/envs/ci-preview/addresses.json)
+export ILLIQUID_SUPPLY_VALIDATOR_ADDRESS=$(jq -r '.addresses.IlliquidCirculationSupplyValidator' dev/envs/ci-preview/addresses.json)
+export GOVERNED_MAP_VALIDATOR_ADDRESS=$(jq -r '.addresses.GovernedMapValidator' dev/envs/ci-preview/addresses.json)
+export GOVERNED_MAP_POLICY_ID=$(jq -r '.policyIds.GovernedMap' dev/envs/ci-preview/addresses.json)
 
 # native token observability
 export NATIVE_TOKEN_POLICY_ID='ada83ddd029614381f00e28de0922ab0dec6983ea9dd29ae20eef9b4'
 export NATIVE_TOKEN_ASSET_NAME='5043546f6b656e44656d6f'
-export ILLIQUID_SUPPLY_VALIDATOR_ADDRESS='addr_test1wrhvtvx3f0g9wv9rx8kfqc60jva3e07nqujk2cspekv4mqs9rjdvz'
```
LGLO, line 12: I have no idea why this one is different to one from addresses. We'll see if it requires ci-preview PC restart.

changelog.md
```diff
@@ -17,7 +17,8 @@ This changelog is based on [Keep A Changelog](https://keepachangelog.com/en/1.1.
 
 ## Changed
 
-
+* Added injecting GovernedMap MainChainScripts configuration into chain-spec file in local-environment and CI.
```
LGLO, line 20: I do not touch a line of `toolkit` or `demo` code in this PR. Should anyone in public care about these changes?

AmbientTea: I don't think so

AmbientTea (approved):

changelog.md
gilligan: Agreed, i see no need to mention anything in the changelog

PR #814, opened 2025-05-23
Simplify nix setup to single flake.nix

gilligan:
# Description
Reduces the nix setup to a **single flake.nix** containing the devShell definition that works for linux and darwin

- Drops process-compose setup which has been deprecated by local-env
- Drops packages which were only provided for process-compose setup
- Drops flake-modules and uses flake-utils instead

This by the way also reduces the time it takes to evaluate this meaning entering the shell / direnv is faster.

# Checklist

- [x] Self-reviewed the diff


AmbientTea (approved):

kpinter-iohk (commented):
works on linux, tested cargo check, cargo clippy, setting up a local-env with wizards, loading python venv for e2e tests

PR #825, opened 2025-05-28
fix: use echo -n to avoid having line feed byte in the hex of validator address

LGLO:
# Description

`echo $SOME_VAR` with `-n` output this var with line feed at the end. `xxd` was picking it up and adding to the hex representation.


gilligan (approved):
This is fine because the file has a `#!/bin/bash` shebang but just as a comment:

Using `printf %s $FOOBAR` is a fully POSIX compatible alternative to `echo -e` which works fine in bash but not with /bin/sh or such.

But completely fine to use `echo -e` in this script :)

AmbientTea (approved):

PR #828, opened 2025-05-28
Replace earthly

gilligan:
# Description
This PR replaces earthly by instead relying on a more simplistic setup where the build time dependencies are provided via Nix from our flake.nix

Some additional notes:
- The PR uses the GitHub cache action and the overall runtime is roughly the same
- The runtime could be improved through parallelization but any optimizations should come as a follow-up
- The `flake.nix` has been reduced a lot in size recently and is now just 114 lines
- With this PR we get a consistent setup where the CI builds are easy to reproduce locally

# Checklist

- [x] Self-reviewed the diff


gilligan:
@skylar-simoncelli All pre-merge steps seem to work. I haven't tested post-merge so I'm gonna assume that's where all my typos are and everything breaks :)

blabla
AmbientTea, line 1: sir step out of the car please

deny.toml
```diff
@@ -0,0 +1,268 @@
...
+    "BSD-3-Clause",
+    "Zlib",
+    "BSL-1.0",
+    "MPL-2.0",
+    "Unlicense",
+    "CC0-1.0",
+    "Unicode-3.0",
+    "CDLA-Permissive-2.0",
+    "Apache-2.0 WITH LLVM-exception"
+]
```
AmbientTea, line 99: We use only GPLv3+CPE and Apache 2. I would allow those and deny everything else.

gilligan: Skylar was getting into some annoyances with the license checking. Moving forward we'll use https://github.com/EmbarkStudios/cargo-deny-action but I said we can solve this separately as a follow-up.

skylar-simoncelli (approved):

PR #870, opened 2025-06-18
move configs to genesis

neotheprogramist:


Klapeyron:
I like the idea of moving configuration on-chain, but I am against implementing it from scratch. We can try to utilize `pallet-parameters` for that, [here](https://github.com/input-output-hk/partner-chains/commit/40a3ea79d02d397d7b15c2d78221aee849fc2a9b) I made some draft with a parameters you used. Feel free to use that and adjust to your needs

PR #874, opened 2025-06-24
change: ETCM-12049 extrinsic to delete SPO metadata

AmbientTea:
# Description

* Adds logic to save depositor when creating SPO metadata
* Changes the signed message to make the metadata field optional, so it can be reused for deletion
* Adds a `delete_metadata` extrinsic to allow SPOs to clear their metadata and reclaim deposit
* Splits the metadata signing CLI command into `upsert` and `delete` subcommands; adds logic to sign deletion message

# Checklist

- [ ] Self-reviewed the diff


toolkit/block-producer-metadata/pallet/src/lib.rs
```diff
@@ -170,6 +171,15 @@ pub mod pallet {
 		QueryKind = OptionQuery,
 	>;
 
+	/// Storage mapping from block producers to their metadata depositor accounts
+	#[pallet::storage]
+	type BlockProducerMetadataOwners<T: Config> = StorageMap<
```
Klapeyron, line 176: we can try to check if using of StorageDoubleMap will be easier here, to keep data in one structure

toolkit/block-producer-metadata/pallet/src/benchmarking.rs
```diff
@@ -91,5 +100,18 @@ mod benchmarks {
 		_(RawOrigin::Signed(caller), metadata, cross_chain_signature, cross_chain_pub_key);
 	}
 
+	#[benchmark]
+	fn delete_metadata() {
+		let cross_chain_pub_key = T::BenchmarkHelper::cross_chain_pub_key();
+		let cross_chain_signature = T::BenchmarkHelper::delete_cross_chain_signature();
+
+		// Create an account and fund it with sufficient balance
```
LGLO, line 108: What is sufficient balance for delete? Is this required?

LGLO: Anyway I think that the setup phase should insert the data instead of just removing something that doesn't exist.

AmbientTea: it's not required, that was copy-paste. I added writing the data and regenerated weights

toolkit/block-producer-metadata/pallet/src/lib.rs
```diff
@@ -170,6 +175,15 @@ pub mod pallet {
...
 	>;
 
+	/// Storage mapping from block producers to their metadata depositor accounts
+	#[pallet::storage]
+	pub(crate) type BlockProducerMetadataOwners<T: Config> = StorageMap<
+		Hasher = Blake2_128Concat,
+		Key = CrossChainKeyHash,
+		Value = T::AccountId,
+		QueryKind = OptionQuery,
+	>;
```
Klapeyron, line 185: We can try to store this data together in one map, should be at least one lookup less and more consistent. Maybe storing some structure with origin and metadata will be ok

Klapeyron: right, but my point is that we have let's say 1 read 1/2 writes depending on the path, which we can narrow to always have 1 read 1 write. Even we will `get` some value and do the same later in the same extrinsic, it is still the same overlay layer so this value is already fetched in a cache from key-value storage. Same applies to a write, we can perform multiple writes to the storage, but the real commit to the underlying db happens on the end of the scope of the transactional layer.

toolkit/block-producer-metadata/pallet/src/lib.rs
```diff
@@ -221,21 +235,73 @@ pub mod pallet {
...
 			ensure!(is_valid_signature, Error::<T>::InvalidMainchainSignature);
 
-			if BlockProducerMetadataStorage::<T>::get(cross_chain_key_hash).is_none() {
-				T::Currency::hold(
-					&HoldReason::MetadataDeposit.into(),
-					&origin_account,
-					T::HoldAmount::get(),
-				)
-				.map_err(|_| Error::<T>::InsufficientBalance)?;
+			if BlockProducerMetadataOwners::<T>::get(cross_chain_key_hash).is_none() {
```
Klapeyron, line 238: there is [contains_key](https://paritytech.github.io/polkadot-sdk/master/frame_support/storage/trait.StorageMap.html#tymethod.contains_key) just in case in case we do not use the value, but I will recommend to aggregate the storages into one entity.

LGLO (approved):
I agree with Tomasz, that it will be simpler code and easier to understand when everything is in one map, but difference is so insignificant that I don't want to restrict PRs author freedom.

toolkit/block-producer-metadata/pallet/src/lib.rs
Klapeyron: keeping it all together allows also to reduce a risk that we will have some leftovers, if we have a case that `cross_chain_key_hash` always needs to be stored with associated `origin` or both have to be removed at the same time. We have some conditional inserting/removal, now it should work but accidentally in the future it can be changed

toolkit/block-producer-metadata/pallet/src/lib.rs
```diff
@@ -221,21 +235,73 @@ pub mod pallet {
...
+		fn hold_deposit(account: &T::AccountId) -> DispatchResult {
+			T::Currency::hold(&HoldReason::MetadataDeposit.into(), account, T::HoldAmount::get())
+				.map_err(|_| Error::<T>::InsufficientBalance)?;
+			Ok(())
+		}
+
+		fn release_deposit(depositor: &T::AccountId, beneficiary: &T::AccountId) -> DispatchResult {
+			use frame_support::traits::tokens::*;
+
+			<T::Currency as MutateHold<_>>::transfer_on_hold(
```
Klapeyron, line 293: If I understand correctly, `transfer_on_hold` makes:
> Transfer held funds into a destination account.

Does it mean that deposit of a user that originally made it will be moved to the next user? Shouldn't we rather release allocated funds of first user and allocate for new one?

Klapeyron, line 293:
```suggestion
			<T::Currency as Mutate<_>>::transfer_on_hold(
```

Klapeyron: if it will require additional import then please resolve this comment

toolkit/block-producer-metadata/pallet/src/lib.rs
```diff
@@ -221,21 +235,73 @@ pub mod pallet {
...
 			}
 
 			BlockProducerMetadataStorage::<T>::insert(cross_chain_key_hash, metadata);
+
+			Ok(())
+		}
+
+		/// Deletes metadata for the block producer identified by `cross_chain_pub_key`.
+		///
+		/// The deposit funds will be returned to the submitter of the extrinsic.
```
Klapeyron, line 250: there is some MEV risk associated with that. In the situation that `MetadataSignedMessage` became public, anybody can submit this transaction and if will be faster than person that signed it, will be able to unlock the funds. In theory, we can assume that signer of the metadata will protect metadata and not share unless will submit an extrinsic, but at that moment this metadata can be also copied once they will start to fly over network. Also at the end of the message trip, malicious SPO can filter-out user extrinsic and apply its own with replaced origin, which will be accepted by state transition function. The only option to safely add this extrinsic will be if origin will be in the same time an block producer in given round and will share this as part of `InBlock` transaction, which hopefully will be not rejected or forked (like with VRF), due otherwise other block producer can take advantage of it.

toolkit/block-producer-metadata/pallet/src/lib.rs
AmbientTea: okay, changed to single storage

toolkit/block-producer-metadata/pallet/src/lib.rs
```diff
@@ -221,21 +235,73 @@ pub mod pallet {
...
+				genesis_utxo,
+			};
+			let cross_chain_key_hash = cross_chain_pub_key.hash();
+			let is_valid_signature =
+				signature.verify(&cross_chain_pub_key, &metadata_message.encode()).is_ok();
+
+			ensure!(is_valid_signature, Error::<T>::InvalidMainchainSignature);
+
+			if let Some(owner) = BlockProducerMetadataOwners::<T>::get(cross_chain_key_hash) {
+				Self::release_deposit(&owner, &origin_account)?;
```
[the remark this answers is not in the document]
AmbientTea: We're not allocating new deposit, we're unlocking it. I added saving the deposit amount together with the depositor account

toolkit/block-producer-metadata/pallet/src/lib.rs
AmbientTea: No, we're releasing the funds. I changed it so that the original depositor gets the deposit back and not the origin.

toolkit/block-producer-metadata/pallet/src/lib.rs
```diff
@@ -161,12 +165,12 @@ pub mod pallet {
...
 	}
 
-	/// Storage mapping from block producers to their metadata
+	/// Storage mapping from block producers to their metadata and owner Substrate account
 	#[pallet::storage]
 	pub type BlockProducerMetadataStorage<T: Config> = StorageMap<
 		Hasher = Blake2_128Concat,
 		Key = CrossChainKeyHash,
-		Value = T::BlockProducerMetadata,
+		Value = (T::BlockProducerMetadata, T::AccountId, BalanceOf<T>),
```
Klapeyron, line 173: what do you think about making some structure with field docs out of it?

toolkit/block-producer-metadata/pallet/src/lib.rs
```diff
@@ -161,12 +165,12 @@ pub mod pallet {
 		type BenchmarkHelper: benchmarking::BenchmarkHelper<Self::BlockProducerMetadata>;
 	}
 
-	/// Storage mapping from block producers to their metadata
+	/// Storage mapping from block producers to their metadata and owner Substrate account
```
Klapeyron, line 168:
```suggestion
	/// Storage mapping from block producers to their metadata, owner Substrate account and hold funds.
```
but maybe we can just say about metadata and content describe in the Value type definition

toolkit/block-producer-metadata/pallet/src/lib.rs
AmbientTea: just changed docs

toolkit/block-producer-metadata/pallet/src/lib.rs
AmbientTea: This will be adressed in a separate PR

Klapeyron: I am afraid that there is still is a risk, even if we will release deposit to the person that put it. It may be still in someone interest to apply the same deregistration extrinsic (either block producer or normal user), that will be in a race with the others, but just by looking on the extrinsic runtime will be not able to categorize them as performing the same operation. It can cause, that in case that there will be at least one extrinsic, only the first one (chosen by block producer) will succeed, causing that all the others may be added to the block with status `ExtrinsicFailed`, causing that funds allocated for that execution will be slashed. It may cause, that honest owner of the metadata who wants to deregister them, can be slashed for trying to do so.

toolkit/block-producer-metadata/pallet/src/lib.rs
```diff
@@ -221,21 +230,81 @@ pub mod pallet {
...
+				None => {
+					let deposit = Self::hold_deposit(&origin_account)?;
+					BlockProducerMetadataStorage::<T>::insert(
+						cross_chain_key_hash,
+						(metadata, origin_account, deposit),
+					);
+				},
+				Some((_old_data, owner, deposit)) => {
+					ensure!(owner == origin_account, Error::<T>::NotTheOwner);
+					BlockProducerMetadataStorage::<T>::insert(
```
Klapeyron, line 243: shall we adjust cost of storing updated data to current configuration?

toolkit/block-producer-metadata/pallet/src/lib.rs
```diff
@@ -221,21 +230,81 @@ pub mod pallet {
...
+
+		fn release_deposit(depositor: &T::AccountId, amount: BalanceOf<T>) -> DispatchResult {
+			use frame_support::traits::tokens::*;
+
+			<T::Currency as MutateHold<_>>::release(
+				&HoldReason::MetadataDeposit.into(),
+				depositor,
+				amount,
+				Precision::BestEffort,
+			)?;
```
Klapeyron, line 304:
```suggestion
			T::Currency::release(
				&HoldReason::MetadataDeposit.into(),
				depositor,
				amount,
				Precision::BestEffort,
			)?;
```

toolkit/block-producer-metadata/pallet/src/benchmarking.rs
```diff
@@ -91,5 +100,22 @@ mod benchmarks {
...
+		let metadata = T::BenchmarkHelper::metadata();
+		let cross_chain_pub_key = T::BenchmarkHelper::cross_chain_pub_key();
+		let cross_chain_signature = T::BenchmarkHelper::delete_cross_chain_signature();
+
+		let caller: T::AccountId = account("caller", 0, 0);
+
+		BlockProducerMetadataStorage::<T>::insert(
+			cross_chain_pub_key.hash(),
+			(metadata, caller.clone(), T::HoldAmount::get()),
+		);
```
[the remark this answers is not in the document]
AmbientTea: added holding to the preparation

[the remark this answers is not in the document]
Klapeyron: I think it will still be better to invoke real `upsert_metadata`. In case that someone in the future will make any adjustments to the code, will have not to remember to make modifications in the benchmark code

PR #911, opened 2025-07-11
feat: use generic account/session keys in authority selection, committee management and runtime

LGLO:
# Description

Removes AURA, and Grandpa from toolkit crates and require that user provides `TryFrom<CandidateKeys>` for his specific keys.

# Checklist

- [x] Self-reviewed the diff


demo/runtime/src/mock.rs
```diff
@@ -123,6 +125,19 @@ impl From<(sr25519::Public, ed25519::Public)> for TestSessionKeys {
 	}
 }
 
+impl TryFrom<CandidateKeys> for TestSessionKeys {
```
Klapeyron, line 128: I just wanted to check — are we not planning to use OpaqueKeys and Decode?

LGLO: Please see what I've done in commit https://github.com/input-output-hk/partner-chains/pull/911/commits/c477bf5e9afd2aa93283adc54b5a5ab1a686dc1d

toolkit/committee-selection/authority-selection-inherents/src/lib.rs
```diff
@@ -108,3 +109,26 @@ impl<AuthorityId: Clone, AuthorityKeys: Clone> CommitteeMemberT
 		}
 	}
 }
+
+/// Trait to try extract [T] from [CandidateKeys].
+pub trait MaybeFromCandidateKeys<T> {
```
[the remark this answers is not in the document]
Klapeyron: if that will work, then in theory user will not need to implement this trait on his side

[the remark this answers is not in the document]
LGLO: User doesn't have to implement this trait on his side. I've done it for him in `ConvertForImplOpaqueKeys`.

We can't have two instances in the way you show. So if it works, then users can't have other implementation.
As I've said before, users can have `OpaqueKeys` that are not implemented with the `impl_opaque_keys!`.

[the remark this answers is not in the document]
Klapeyron: Are there any specific cases where users wouldn't use impl_opaque_keys? I'm not sure it's worth adding extra complexity to handle such scenarios.
The macro follows the same convention as construct_runtime and others, which are widely adopted across chains. Even in cases where someone decides not to use impl_opaque_keys, they would still need to manually implement all the associated traits that the macro would generate — meaning relying solely on Decode and OpaqueKeys should still be sufficient.

toolkit/committee-selection/authority-selection-inherents/src/filter_invalid_candidates.rs
```diff
@@ -652,12 +650,12 @@ mod tests {
...
 		);
-		let valid_aura_key = AuraPublicKey(
-			hex!("d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d").to_vec(),
-		);
-		let valid_grandpa_key = GrandpaPublicKey(
-			hex!("88dc3417d5058ec4b4503e0c12ea1a0a89be200fe98922423d4334014fa6b0ee").to_vec(),
-		);
+		let aura_key_bytes =
+			hex!("d43593c715fdd31c61141abd04a99fd6822c8558854ccde39a5684e7a56da27d");
+		let valid_aura_key = CandidateKey { id: *b"sr25", bytes: aura_key_bytes.to_vec() };
```
Klapeyron, line 655: sr25 and ed25 are better suited as values of [CryptoTypeId](https://paritytech.github.io/polkadot-sdk/master/sp_core/crypto/struct.CryptoTypeId.html). For [KeyTypeId](https://paritytech.github.io/polkadot-sdk/master/sp_core/crypto/struct.KeyTypeId.html), it would be more appropriate to use identifiers that reflect the component or module that utilizes the key—something like test, dumm, or more specific names such as aura or gran might be a better fit.

toolkit/committee-selection/authority-selection-inherents/src/lib.rs
```diff
@@ -108,3 +109,18 @@ impl<AuthorityId: Clone, AuthorityKeys: Clone> CommitteeMemberT
...
+/// Trait to try extract [T] from [CandidateKeys].
+pub trait MaybeFromCandidateKeys: OpaqueKeys + Decode + Sized {
+	/// Depends on `Encode` that is derived by `impl_opaque_keys!`
+	fn maybe_from(keys: &CandidateKeys) -> Option<Self> {
+		let required_keys = Self::key_ids();
+
+		let mut encoded_keys = sp_runtime::BoundedVec::<u8, ConstU32<1024>>::new();
+		for key_id in required_keys {
+			let key = keys.0.iter().find(|key| key.id == key_id.0)?;
+			encoded_keys.try_append(&mut key.bytes.clone()).ok()?;
```
Klapeyron, line 122: Please double-check whether pushing the raw bytes directly works as expected. I'm currently unsure whether the keys will be interpreted as Vec<u8> or fixed-size arrays during decoding.

If they are treated as Vec<u8>, then we need to prepend the length of each key (or use .encode() directly), since:

> A collection of same-typed values is encoded, prefixed with a compact encoding of the number of items, followed by each item's encoding concatenated in turn.

On the other hand, if they're treated as arrays of fixed length, the current approach should work fine, as:

> For structures, the values are named, but that is irrelevant for the encoding (names are ignored - only order matters).

It would be great to add a test case in this file that explicitly confirms which scenario applies.

LGLO: There are many tests check scenario where there is sr25519 and/or ed25519. The most direct is the added one in tests.rs

toolkit/committee-selection/authority-selection-inherents/src/filter_invalid_candidates.rs
Klapeyron: my intention is not to introduce aura or grandpa pallets, but rather to use `KeyTypeId` to reflect the module that utilizes it. This helps avoid confusion, as identifiers like `sr25` or `ed25` might misleadingly suggest a `CryptoTypeId` instead.

toolkit/committee-selection/authority-selection-inherents/src/lib.rs
Klapeyron: My point here is that we're relying on a very specific behavior: we assume the keys are provided as-is, and we simply sort them before using them to decode the outer type. Given how specific this approach is, I think it's worth adding a few validation checks to the proposed algorithm, such as:

(_Incorrect key order_): the algorithm will sort them.
(_Missing keys_): the algorithm should fail.
(_Duplicate keys_): we need to decide whether to take the first occurrence or fail.
(_Unexpected (extra) keys_): we should decide whether to ignore them or fail.

I would like to ensure, that serialization of the bytes as they are in correct order is sufficient.

LGLO: Okay, now I get what you mean.
Incorrect key order => there is no assumption on ordering, I'll add a test for having different ordering in CandidateKeys,

Missing key => return None,

Duplicate key => I don't want to write spec, if there are duplicated key, this means we have written it and it is internal, programatic error. This is not handling of user input

Extra key => I also don't want to write specification here (however I see that algorithm won't fail)

toolkit/committee-selection/authority-selection-inherents/src/lib.rs
```diff
@@ -108,3 +109,18 @@ impl<AuthorityId: Clone, AuthorityKeys: Clone> CommitteeMemberT
...
+		}
+		Self::decode(&mut &encoded_keys[..]).ok()
+	}
+}
```
AmbientTea, line 126: We could leave `MaybeFromCandidateKeys` unconstrained and add an `impl` for types that implement `OpaqueKeys`. That way chain builders can if needed implement their own `MaybeFromCandidateKeys` if for whatever reason they're not using `OpaqueKeys`

toolkit/committee-selection/authority-selection-inherents/src/lib.rs
```diff
@@ -108,3 +109,18 @@ impl<AuthorityId: Clone, AuthorityKeys: Clone> CommitteeMemberT
 		}
 	}
 }
+
+/// Trait to try extract [T] from [CandidateKeys].
+pub trait MaybeFromCandidateKeys: OpaqueKeys + Decode + Sized {
+	/// Depends on `Encode` that is derived by `impl_opaque_keys!`
```
AmbientTea, line 115:
```suggestion
	/// Depends on `Decode` that is derived by `impl_opaque_keys!`
```

AmbientTea: Or `Codec`, because we're emulating the generated `Encode` and depending on the generated `Decode` to correctly read our bytes.

AmbientTea (approved):

toolkit/committee-selection/authority-selection-inherents/src/filter_invalid_candidates.rs
Klapeyron: I am not against `sp_runtime::app_crypto::sr25519::AppPublic`, my point is to rename the identifiers, due Candidate keys have 4 character ident of KeyTypeId, not CryptoTypeId. Keeping it as `sr25` or something can be misleading, so I suggest to rename it to whatever else, even "test"/"dumm" or "aura"/"gran", that is not associated with underlying crytpography beeing used

toolkit/committee-selection/authority-selection-inherents/src/lib.rs
Klapeyron: this emaulation we have to check, because we take the key bytes and identifiers as they are in the byte form already. In theory, if we will then involve encode on them, then we can unintentionally prefix them with the length of the bytes.

toolkit/committee-selection/authority-selection-inherents/src/filter_invalid_candidates.rs
LGLO: In production code, when our SessionKeys are used, there are `aura` and `gran` used, not `sr25` and `ed25`.

Klapeyron: I understand the difference, but I am suprised that you have such identifier in paritular case. I expected that there will be something like [here](https://paritytech.github.io/polkadot-sdk/master/sp_application_crypto/macro.app_crypto.html):
```rust
// Declare a new set of crypto types using ed25519 logic that identifies as `KeyTypeId`
// of value `b"fuba"`.
app_crypto!(ed25519, KeyTypeId(*b"fuba"));
```
So we have indeed crypto associated with `sr25519`/`ed25519`, but the app (like `AuraLikeModule`) has some identifier like `fuba`, which we later use as an identifier. Maybe it is a case, that we are using some traits more directly, for instance without `app_crypto!` macro.

PR #914, opened 2025-07-16
[DO NOT MERGE] ETCM-9835 tx out consumed query

AmbientTea:
# Description

Showcase of changes to the queries required to make use of the `tx_out.value = "consumed"` option in db-sync configuration. When this option is set:
1. the `tx_out` table has an additional column `consumed_by_tx_id` which references the transaction in `tx` table that consumed given output
2. the `tx_in` does not exist
In practice this means we can use a single join between `tx` and `tx_out` when we need to determine inputs to a transaction instead of two joins going through `tx_in`, which also makes the queries easier to understand.

### Next steps

If we decide to support this option, I see three ways to go about it:
1. update the queries and strictly require db-sync to be configured with `tx_out.value = "consumed"`
2. preserve the old queries, supporting both old and new configurations, and either:
    a. let the user control which queries are run by some `TX_OUT_STYLE` environment variable
    b. automatically detect the setting by checking for existence of `tx_in`

I lean towards option 1

### Testing

I run two nodes compiled from this branch and `master` and they synced each other's blocks without issues.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/dev/local-environment/configurations/db-sync/config.json
+++ b/dev/local-environment/configurations/db-sync/config.json
@@ -1,4 +1,9 @@
 {
+  "insert_options": {
+    "tx_output": {
+      "value": "consumed"
+    }
+  },
   "EnableLogMetrics": false,
   "EnableLogging": true,
   "NetworkName": "Testnet",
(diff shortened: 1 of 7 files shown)


gilligan:
> I lean towards option 1

yeah, for sure. I think option 2 just introduces too much complication..

PR #927, opened 2025-07-25
fix: metadata extrinsic front-run attack, rework benchmarks

AmbientTea:
# Description

* Adds `AccountId` of the extrinsic submitter to the signed message for updating block producer metadata to prevent front-run attacks
* Reworks the tests and benchmarks for the metadata pallet, so that signatures are now automatically signed instead of requiring manual updating using the CLI

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/toolkit/cli/commands/src/block_producer_metadata_signatures.rs
+++ b/toolkit/cli/commands/src/block_producer_metadata_signatures.rs
@@ -6,13 +6,13 @@ use serde::de::DeserializeOwned;
 use serde_json::{self, json};
 use sidechain_domain::*;
 use sp_block_producer_metadata::MetadataSignedMessage;
-use std::io::BufReader;
+use std::{io::BufReader, str::FromStr};
 use time_source::{SystemTimeSource, TimeSource};
 
 /// Generates ECDSA signatures for block producer metadata using cross-chain keys.
 #[derive(Clone, Debug, clap::Subcommand)]
 #[command(author, version, about, long_about = None)]
-pub enum BlockProducerMetadataSignatureCmd {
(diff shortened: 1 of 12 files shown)


toolkit/cli/commands/src/block_producer_metadata_signatures.rs
```diff
@@ -27,6 +27,9 @@ pub enum BlockProducerMetadataSignatureCmd {
 		/// Time-to-live of the signature in seconds.
 		#[arg(long, default_value = "3600")]
 		ttl: u64,
+		/// Partner Chain Account that will be used to upsert the metadata and will own it on-chain
+		#[arg(long, value_parser=parse_partner_chain_accounts::<AccountId>)]
```
kpinter-iohk, line 31: You can omit `value_parser` and get rid of the `parse_partner_chain_accounts` function if you change the `FromStr` constraint on `AccountId` to `FromStr<Err : Into<Box<(dyn std::error::Error + Send + Sync + 'static)>>>`.
Since the `FromStr` constraint appears multiple times you could create a trait with the constraint on `Err`:
```rust
pub trait FromStrStdErr: FromStr<Err : Into<Box<(dyn std::error::Error + Send + Sync + 'static)>>> {}
impl<T: FromStr<Err : Into<Box<(dyn std::error::Error + Send + Sync + 'static)>>>> FromStrStdErr for T {}
```

AmbientTea: eh, I'll leave it like that. this type looks ugly and brittle

PR #935, opened 2025-07-29
Remove custom clap parser functions

kpinter-iohk:
# Description

`clap` allows use of `value_parser` to specify what function should be used over the `FromStr` instance of a type when parsing command line arguments. In many cases we unnecessarily use this feature when a `FromStr` instance exists.

When the `FromStr` instance for a type has an `Err` type defined that does not convert to StdErr + etc as expected by clap it generates a confusing macro compile error which seems to have been worked around with custom `value_parser`s. To fix this a `FromStrStdErr` trait (quasi trait alias) is now provided that can be used when the struct used with `clap::Parser` has type params.

The parser for `PermissionedCandidateData` has been moved into a `FromStr` instance.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/toolkit/sidechain/domain/src/lib.rs
+++ b/toolkit/sidechain/domain/src/lib.rs
@@ -1265,6 +1265,76 @@ pub struct PermissionedCandidateData {
 	pub keys: CandidateKeys,
 }
 
+/// Parses public keys formatted as PARTNER_CHAINS_KEY:AURA_KEY:GRANDPA_KEY or PARTNER_CHAINS_KEY,KEY_ID_1:KEY_1,...,KEY_ID_N:KEY_N
+#[cfg(feature = "std")]
+impl FromStr for PermissionedCandidateData {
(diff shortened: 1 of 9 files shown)


toolkit/sidechain/domain/src/lib.rs
```diff
@@ -1265,6 +1265,76 @@ pub struct PermissionedCandidateData {
...
+
+	fn from_str(partner_chain_public_keys: &str) -> Result<Self, Self::Err> {
+		fn is_legacy_format(line: &str) -> bool {
+			line.contains(':') && !line.contains(',')
+		}
+
+		fn parse_legacy_format(
+			line: &str,
+		) -> Result<
+			PermissionedCandidateData,
```
LGLO, line 1281: Cosmetic: `Self` could be used.

kpinter-iohk: Unfortunately: ```can't use `Self` from outer item refer to the type directly here instead```

PR #961, opened 2025-08-20
local-environment upgrades and tests fixes

skylar-simoncelli:


dev/local-environment-dynamic/configurations/cardano/entrypoint.sh
```diff
@@ -372,8 +372,8 @@ echo "[LOG] Waiting 45 seconds for the main transaction to process and be confir
 sleep 45
 
 echo "[LOG] Querying and saving the first UTXO details for new address to /shared/genesis.utxo:"
-# Query UTXOs and extract the first UTXO key from JSON format
-cardano-cli latest query utxo --testnet-magic 42 --address "${new_address}" --out-file /dev/stdout | /busybox grep -o '"[a-f0-9]\{64\}#[0-9]\+":' | head -1 | /busybox sed 's/"//g' | /busybox sed 's/://g' > /shared/genesis.utxo
+cardano-cli latest query utxo --testnet-magic 42 --address "${new_address}" | /busybox awk -F'"' '/#/ {gsub(":", "", $2); print $2; exit}' > /shared/genesis.utxo
```
LGLO, line 375: As output is a JSON, I think it is proper to use `jq` instead of `awk`.

PR #964, opened 2025-08-25
Update cardano-node to v10.5.1 in local-environment

LGLO:
Updates cardano-node to v10.5.1 in local-environment (-dynamic is left for another time, as it requires manual testing).
Follow comments in the diff for explanation.


dev/local-environment-dynamic/modules/cardano.txt
```diff
@@ -19,7 +19,7 @@
       - ./configurations/cardano/keys/op.cert:/keys/node.cert
       - ./configurations/cardano/reward_token_policy.script:/shared/reward_token_policy.script
       - ./configurations/cardano/v-function.script:/shared/v-function.script
-      - ./configurations/cardano/config-pool1.json:/shared/node-1-config.json.base
```
LGLO, line 22: Updating genesis hashes in this file is gone in this PR, so we have only final version without .base suffix

dev/local-environment/configurations/genesis/byron/genesis.json
```diff
@@ -62,6 +62,6 @@
         { "summand": "155381000000000" , "multiplier": "43946000000" }
     , "unlockStakeEpoch": "18446744073709551615"
     }
-, "protocolConsts": { "k": 10 , "protocolMagic": 42 }
```
LGLO, line 65: This was problematic. We had an inconsistency that is fatal in v10.5.1. k=5 is what we want as security parameter.

dev/local-environment/configurations/cardano/entrypoint.sh
```diff
@@ -170,17 +170,17 @@ echo $new_address > /shared/FUNDED_ADDRESS
 echo "Created /shared/FUNDED_ADDRESS with value: $new_address"
 
 echo "Querying and saving the first UTXO details for node-4 address to /shared/node4.utxo:"
-cardano-cli latest query utxo --testnet-magic 42 --address "${node4_address}" | /busybox awk 'NR>2 { print $1 "#" $2; exit }' > /shared/node4.utxo
+cardano-cli latest query utxo --testnet-magic 42 --address "${node4_address}" | jq -r 'keys|.[0]' > /shared/node4.utxo
```
LGLO, line 173: cardano-node 10.5.1 default output is JSON. I take the first key here.

PR #967, opened 2025-08-26
feat: upgrade cardano in local-environment-dynamic to 10.5.1

skylar-simoncelli:


dev/local-environment-dynamic/configurations/cardano/config-pool1.json
```diff
@@ -28,46 +28,51 @@
...
+    "TraceMux": { "isOn": false },
+    "TracePeerSelection": { "isOn": true },
+    "TracePeerSelectionActions": { "isOn": true },
+    "TracePublicRootPeers": { "isOn": true },
+    "TraceServer": { "isOn": true },
+    "TraceTxInbound": { "isOn": false },
+    "TraceTxOutbound": { "isOn": false },
+    "TraceTxSubmissionProtocol": { "isOn": false }
+  },
+
```
LGLO, line 72: There are lot of options that are added to the object above but are not removed from the "old keys"

skylar-simoncelli: Yes this is now based off your config file instead (with some minor changes to suppress log verbosity)

AmbientTea (approved):

PR #983, opened 2025-09-03
add: ETCM-12255 bridge data source

AmbientTea:
# Description

this is a wip, get out of here

# Checklist

- [ ] Self-reviewed the diff


toolkit/data-sources/db-sync/src/bridge/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+				&main_chain_scripts.illiquid_supply_validator_address.into(),
+				asset,
+				(block_number, tx_ix, tx_out_ix),
+				current_mc_block.block_no,
+				max_transfers,
+			)
+			.await?;
+
+			let new_checkpoint = match utxos.last() {
+				None => data_checkpoint,
```
LGLO, line 78: I would move the the checkpoint in this case to at least current block, but the problem is that `BridgeDataCheckpoint` is an `UtxoId` instead of the triplet we use in query for new utxos.

There could be a problem with range of this query ever increasing.

I understand that with the triplet, the problem is that we can't use _genesis_utxo_ as the initial value?

toolkit/data-sources/db-sync/src/bridge/mod.rs
```diff
@@ -0,0 +1,119 @@
...
+	}
+);
+
+fn utxo_to_transfer<RecipientAddress>(
+	utxo: BridgeUtxo,
+) -> Option<BridgeTransferV1<RecipientAddress>>
+where
+	RecipientAddress: for<'a> TryFrom<&'a [u8]>,
+{
+	let token_delta = (utxo.tokens_out.0 as i128) - (utxo.tokens_in.0 as i128);
```
LGLO, line 95: This works with assumption of that Deposit transaction spending from ICS has exactly one ICS input and one ICS output. This is enforced by smart-contract rules (ICS tokens not leaking AND unique output AND 1 ICS token in UTXO IMPLIES there is one input as well).
I think we should document it in this repo, because without these assumptions there are scenarios where this logic will not work. Basically, without unique output there are a lot of scenarios we can't handle.

toolkit/data-sources/db-sync/testdata/bridge/migrations-tx-in-consumed/6_insert_transactions.sql
```diff
@@ -0,0 +1,102 @@
...
+;
+
+
+INSERT INTO tx_out ( id, tx_id                 , index, address     , address_raw, address_has_script, payment_cred, stake_address_id, value, data_hash                    , consumed_by_tx_id     )
+            VALUES ( 11, init_ics_tx           , 0    , 'ics address', ''         , TRUE              , NULL        , NULL            , 0    , NULL                        , reserve_transfer_tx   ) -- ICS initial utxo 1
+                  ,( 12, init_ics_tx           , 1    , 'ics address', ''         , TRUE              , NULL        , NULL            , 0    , NULL                        , NULL                  ) -- ICS initial utxo 2
+                  ,( 13, init_ics_tx           , 2    , 'ics address', ''         , TRUE              , NULL        , NULL            , 0    , NULL                        , NULL                  ) -- ICS initial utxo 3
+                  ,( 14, init_ics_tx           , 3    , 'ics address', ''         , TRUE              , NULL        , NULL            , 0    , NULL                        , reserve_transfer_tx   ) -- ICS initial utxo 4
```
LGLO, line 70: `reserve_transfer_tx` consumes two UTXOs, why? In fact none of Deposit transfers can do that.

toolkit/data-sources/db-sync/src/db_model.rs
```diff
@@ -893,4 +893,217 @@ mod tests {
...
+    , datum.value                             AS datum
+
+FROM tx_out      outputs
+JOIN tx                        ON outputs.tx_id = tx.id
+JOIN block                     ON tx.block_id = block.id
+JOIN ma_tx_out   output_tokens ON output_tokens.tx_out_id = outputs.id
+JOIN multi_asset native_token  ON native_token.id = output_tokens.ident
+JOIN datum                     ON datum.tx_id = tx.id
+
+LEFT JOIN tx_out     inputs        ON inputs.consumed_by_tx_id = tx.id
```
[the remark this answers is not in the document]
AmbientTea: good catch. I fixed this in both queries and added a utxo from outside that covers this case in the test

toolkit/data-sources/db-sync/testdata/bridge/migrations-tx-in-consumed/6_insert_transactions.sql
AmbientTea: Mistake on my part, this as you wrote can't happed in the wild

toolkit/data-sources/db-sync/src/bridge/tests.rs
```diff
@@ -0,0 +1,148 @@
+use super::TokenBridgeDataSourceImpl;
+use hex_literal::hex;
+use sidechain_domain::byte_string::ByteString;
+use sidechain_domain::{AssetName, MainchainAddress, McBlockHash, McTxHash, PolicyId, UtxoId};
+use sp_partner_chains_bridge::MainChainScripts;
+use sp_partner_chains_bridge::TokenBridgeDataSource;
+use sp_partner_chains_bridge::{BridgeDataCheckpoint, BridgeTransferV1};
```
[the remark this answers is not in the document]
AmbientTea: merged the imports

toolkit/data-sources/db-sync/src/bridge/mod.rs
AmbientTea: I added a section in this module docs (which I leave here for later to write the rest of)

PR #989, opened 2025-09-08
chain-spec for bridge pallet and few touches to chain-spec building

LGLO:
# Description

* wizard is updated in place and it sets bridge pallet values instead of native-token pallet values
* in other places bridge configuration is only added
* After 4 months the idea of updating chain-spec after building one with Default values wasn't picked up, so I reverted this approach for governed-map

Ref: ETCM-12269
# Checklist

- [x] Any changes are noted in the `changelog.md` for affected crate <- changelog already mentions big chain-spec changes in wizards, bridge part should be in the bridge section
- [x] Self-reviewed the diff


toolkit/bridge/primitives/src/lib.rs
```diff
@@ -40,6 +40,36 @@ pub struct MainChainScripts {
...
+	/// It expects the following variables to be set:
+	/// - `BRIDGE_TOKEN_POLICY_ID`
+	/// - `BRIDGE_TOKEN_ASSET_NAME`
+	/// - `ILLIQUID_CIRCULATION_SUPPLY_VALIDATOR_ADDRESS`
+	pub fn read_from_env() -> Result<Self, envy::Error> {
+		#[derive(serde::Serialize, serde::Deserialize)]
+		pub struct MainChainScriptsEnvConfig {
+			pub bridge_token_policy_id: PolicyId,
+			pub bridge_token_asset_name: AssetName,
+			pub illiquid_circulation_supply_validator_address: MainchainAddress,
```
LGLO, line 56: The other envy formats used a `String`. Was it because we used to serialize address as hexstring in the past?

dev/local-environment/configurations/partner-chains-setup/entrypoint.sh
```diff
@@ -54,9 +54,12 @@ echo "Permissioned candidates policy ID: $PERMISSIONED_CANDIDATES_POLICY_ID"
 
 echo "Setting values for NATIVE_TOKEN_POLICY_ID, NATIVE_TOKEN_ASSET_NAME, and ILLIQUID_SUPPLY_VALIDATOR_ADDRESS for chain-spec creation"
 export ILLIQUID_SUPPLY_VALIDATOR_ADDRESS=$(jq -r '.addresses.IlliquidCirculationSupplyValidator' addresses.json)
+export ILLIQUID_CIRCULATION_SUPPLY_VALIDATOR_ADDRESS=$(echo $ILLIQUID_SUPPLY_VALIDATOR_ADDRESS)
 echo "Illiquid Circulation Supply Validator address: $ILLIQUID_SUPPLY_VALIDATOR_ADDRESS"
 export NATIVE_TOKEN_POLICY_ID="1fab25f376bc49a181d03a869ee8eaa3157a3a3d242a619ca7995b2b"
 export NATIVE_TOKEN_ASSET_NAME="52657761726420746f6b656e"
```
kpinter-iohk, line 60: Are these still needed?

LGLO: Yes and they will be required until we have native-token-management. Its related read_from_env is using these values. We need native-token-management pallet as long as we have e2e-tests for them.

kpinter-iohk (approved):

PR #991, opened 2025-09-09
change: Use numeric traits and helper functions for domain types

AmbientTea:
# Description

Adds traits and helper functions to avoid using `.0` in multiple places in the code when handling block numbers and other numeric values

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/toolkit/data-sources/db-sync/src/bridge/mod.rs
+++ b/toolkit/data-sources/db-sync/src/bridge/mod.rs
@@ -116,13 +116,13 @@ fn utxo_to_transfer<RecipientAddress>(
 where
 	RecipientAddress: for<'a> TryFrom<&'a [u8]>,
 {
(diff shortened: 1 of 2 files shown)


toolkit/data-sources/db-sync/src/bridge/mod.rs
```diff
@@ -117,7 +117,7 @@ fn utxo_to_transfer<RecipientAddress>(
 where
 	RecipientAddress: for<'a> TryFrom<&'a [u8]>,
 {
-	let token_delta = (utxo.tokens_out.0 as i128) - (utxo.tokens_in.0 as i128);
+	let token_delta = utxo.tokens_out.0 - utxo.tokens_in.0;
```
LGLO, line 120: For this one I would rather use `checked_sub`. This function is already `Option`.

PR #995, opened 2025-09-10
feat: use uv for e2e-tests and add to devshell

tgunnoe:
* replace requirements.txt with pyproject.toml. this brings the e2e-tests python environment in line with modern python projects. It also allows specifying others useful properties like linters and formatters

* add uv and uv.lock. uv is a modern python package manager that achieves both better speeds and reproducability through locking the exact dependencies required, much like Cargo.lock

* add these to the nix devshell

* bump omegaconf to the latest preview, because 2.3.0 included an older version of antlr4-python3-binary that used an older manifest version from the rest of the toolchain
# Checklist

- [ ] Self-reviewed the diff


ladamesny (approved):
LGTM

PR #1000, opened 2025-09-16
feat: parrallel builds for stable + nightly

skylar-simoncelli:


.github/workflows/ci.yml
```diff
@@ -251,6 +240,177 @@ jobs:
...
+            mkdir -p /tmp/sccache
+            nix develop -c sccache --stop-server || true
+            nix develop -c sccache --start-server || true
+            nix develop -c sccache --version || true
+          else
+            echo "sccache not in devshell; skipping wrapper."
+          fi
+
+      - name: Build (Nightly)
+        run: nix develop -c bash -c "RUSTUP_TOOLCHAIN=nightly cargo build --locked"
```
LGLO, line 361: Lets see how it works. I think we could also have another toolchain TOML file and direct to it.

skylar-simoncelli: Merged, let's see it running and can revert if needed. Has worked on last few runs, but there was one run where it didn't:
image

https://github.com/input-output-hk/partner-chains/actions/runs/17795340606

skylar-simoncelli: No thoughts on the separate toolchain file, just targetting nightly seems cleaner to me but whatever is preferable

PR #1028, opened 2025-09-30
Make `prompt_multi_option` more typey

kpinter-iohk:
# Description

`prompt_multi_option` in `IOContext` takes a vec of strings to present to the user and returns the selected one. This leads to situations where values have to be converted to strings, and the result of the input parsed, converted back to the original type, or the vec of strings searched for the position of the returned string.
To avoid this, `prompt_multi_option` now takes a vec of `T` and returns the index of the item the user chose. This gives much more flexibility eg. a vec of strings can still be passed if one wants to use custom formatting for their prompt items, and then use the returned index on the original typed vec.

# Checklist

- [ ] Self-reviewed the diff

### DIFF
--- a/toolkit/partner-chains-cli/src/prepare_configuration/mod.rs
+++ b/toolkit/partner-chains-cli/src/prepare_configuration/mod.rs
@@ -1,5 +1,6 @@
 use crate::CmdRun;
 use crate::cmd_traits::{GetScriptsData, InitGovernance};
+use crate::config::SelectOptions;
 use crate::config::config_fields::{BOOTNODES, SUBSTRATE_NODE_DATA_BASE_PATH};
 use crate::generate_keys::network_key_path;
 use crate::io::IOContext;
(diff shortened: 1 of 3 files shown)


toolkit/partner-chains-cli/src/prepare_configuration/mod.rs
```diff
@@ -195,6 +196,12 @@ impl FromStr for Protocol {
 	}
 }
 
+impl Display for Protocol {
+	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+		write!(f, "{}", ToString::to_string(self))
```
LGLO, line 201: `ToString` uses `Display`, so I suspect this can be the cause of stack overflow:
https://doc.rust-lang.org/std/string/trait.ToString.html

toolkit/partner-chains-cli/src/prepare_configuration/mod.rs
```diff
@@ -160,7 +161,7 @@ fn peer_id_from_config(context: &impl IOContext) -> anyhow::Result<Option<String
 }
 
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Hash)]
-enum Protocol {
+pub enum Protocol {
```
LGLO, line 164: would be enough to make it and function in `scenarios` pub crate

PR #1045, opened 2025-10-08
change: Benchmark selection pallet

AmbientTea:
# Description

Turns out the benchmarks for the `session-validator-management-pallet` were really out of date and were missing one extrinsic. This PR:
- Moves them from a separate crate to a sub-module like in the other crates
- Simplifies the benchmarks (I think they were written very early when we were clueless)
- Adds a benchmark for `set_main_chain_scripts` and uses it for the extrinsic (it reused weights for `set` for some reason)

# Checklist

- [x] Self-reviewed the diff

### DIFF
--- a/toolkit/committee-selection/pallet/src/benchmarking.rs
+++ b/toolkit/committee-selection/pallet/src/benchmarking.rs
@@ -2,10 +2,8 @@ use crate::*;
 use frame_benchmarking::v2::*;
 use frame_support::BoundedVec;
 use frame_system::RawOrigin;
-use sidechain_domain::byte_string::SizedByteString;
 use sp_core::Get;
 use sp_runtime::traits::One;
-use sp_session_validator_management::MainChainScripts;
 use sp_std::vec::Vec;
 
 const SEED: u32 = 0;


toolkit/committee-selection/pallet/src/benchmarking.rs
AmbientTea, line 1: This was really moved to `benchmarking.rs` but I was able to strip enough unnecessary code that git doesn't recognize it as the same file any longer even after `git mv`.

toolkit/committee-selection/pallet/src/benchmarking.rs
```diff
@@ -0,0 +1,91 @@
...
+use sp_core::Get;
+use sp_runtime::traits::One;
+use sp_session_validator_management::MainChainScripts;
+use sp_std::vec::Vec;
+
+const SEED: u32 = 0;
+
+/// Helper for creating mock data used by benchmarks.
+///
+/// Only functions without default implementation need to be implemented by chain builders
```
LGLO, line 15: In such a case, do we need `create_inputs_hash()` and `create_main_chain_scripts()` ? Couldn't we inline them?
Less code and less comments.

AmbientTea: image

AmbientTea: Thanks for keeping my over-engineering impulses in check I guess.

PR #1109, opened 2025-11-21
add: ETCM-12351 dolos governed map data source

ladamesny:
# Description

Implemented governed map dolos data source and tweaked the stake distribution code. Confirmed that governed map, bridge and stake distribution sources from dolos is working on the local environment.

# Checklist

- [x] Self-reviewed the diff


chrispalaskas (changes requested):

toolkit/data-sources/dolos/src/client/minibf.rs
```diff
@@ -114,6 +114,16 @@ impl MiniBFApi for MiniBFClient {
 		self.paginated_request_all(&format!("addresses/{address}/utxos")).await
 	}
 
+	async fn addresses_utxos_asset(
```
AmbientTea: This naming convention is kinda unwieldy IMO. I understand that it mirrors the request path, but it tells me nothing without looking into implementation. what about `utxos_at_address_with_asset`?

toolkit/data-sources/dolos/src/bridge.rs
```diff
@@ -138,48 +137,63 @@ pub(crate) async fn get_bridge_utxos_tx(
...
 ) -> Result<Vec<BridgeUtxo>> {
-	let asset_txs = client.assets_transactions(native_token.clone()).await?;
-	let futs = asset_txs.into_iter().map(|a| async move {
-		if (a.block_height as u32) <= to_block.0 {
-			let tx_hash = McTxHash::from_hex_unsafe(&a.tx_hash);
-			let utxo = client.transactions_utxos(tx_hash).await?;
+	// Use the optimized endpoint to get UTXOs at ICS address filtered by the bridge token
+	// This is much more efficient than querying all asset transactions
+	let address_utxos =
+		client.addresses_utxos_asset(icp_address.clone(), native_token.clone()).await?;
```
AmbientTea: Am I understanding it right, that we're fetching **all** utxos at this address with this asses **ever**, and then filter out the ones before `to_block` and after `checkpoint`? If that's the best we can do with what API Dolos provides us with, I don't see a way this can scale on mainnet, especially for a **bridge**.

AmbientTea (commented):
Reading this PR, I can't help but think that all the code here is struggling very hard to do what a couple of SQL queries do relatively easily for Db-Sync data sources. It doesn't look like the API exposed by Dolos is well suited to our use case, as long as we can't narrow down our requests to a specific block range and are forced to run separate queries for each utxo.

Also a minor thing - the title mentions only governed map data sources, but implements 2 others as well. I would stick to 1 PR - 1 data source, unless they're trivial (they're not).

PR #1123, opened 2025-12-11
change: move data epoch offset calculation out of data sources

AmbientTea:
# Description

Removes the `data_epoch` from the `AuthoritySelectionDataSource` trait. This method was used internally to apply the 2-epoch offset within each data source method, but was also exposed outside because `AriadneInherentDataProvider` needed it too in its own logic.

Instead, the data source now simply serves data for the epoch it receives and it's solely `AriadneInherentDataProvider`'s responsibility to calculate it.

This change is backwards compatible.

# Checklist

- [x] Self-reviewed the diff


kpinter-iohk (approved):

LGLO (approved):
Thank you!

PR #1153, opened 2026-04-20
fix(session-validator-management): don't panic when validator IDP returns None

realdecimalist:
## Summary

\`pallet-session-validator-management\`'s \`inherent_data_to_authority_selection_inputs\` uses \`.expect()\` on both the decode result and the \`Option\` returned by the IDP:

\`\`\`rust
// toolkit/pallets/session-validator-management/src/lib.rs:333-343 (as of v1.5.1 / 20e607f)
fn inherent_data_to_authority_selection_inputs(
    data: &InherentData,
) -> (T::AuthoritySelectionInputs, SizedByteString<32>) {
    let decoded_data = data
        .get_data::<T::AuthoritySelectionInputs>(&INHERENT_IDENTIFIER)
        .expect("Validator inherent data not correctly encoded")
        .expect("Validator inherent data must be provided"); // <-- panics
    let data_hash = SizedByteString(blake2_256(&decoded_data.encode()));
    (decoded_data, data_hash)
}
\`\`\`

In practice the authority-selection IDP (fed by the db-sync follower) can transiently return \`None\` — e.g. when db-sync falls behind, or when the Cardano chain crosses an epoch boundary before the next committee's selection inputs are stable. When that happens every block author on the partner chain panics during \`BlockBuilder_inherent_extrinsics\`, so no block is ever produced and the chain wedges.

We hit this on the Materios preprod partner chain (based on v1.5.1): a transient IDP outage at block #14661 stalled the chain for ~17h with exactly this panic trace.

## The fix

- \`inherent_data_to_authority_selection_inputs\` now returns \`Option\` and never panics.
- \`create_inherent\` returns \`None\` when data is absent (don't emit a \`set\` call; the next successful author will).
- \`check_inherent\` accepts a \`set\` call without reverifying when data is absent (proposer had the same IDP view).
- \`is_inherent_required\` only demands the inherent when the IDP has data AND \`NextCommittee\` is missing — otherwise verifiers would reject blocks that don't include the inherent for the same reason the authors didn't emit one.

The happy path (IDP returns \`Some(...)\`) is byte-identical: same \`set\` call, same committee hash verification, same \`InvalidValidatorsMatchingHash\` / \`InvalidValidatorsHashMismatch\` errors on mismatch. Only the panic-and-wedge failure mode becomes a skip-and-resume.

## Test plan

- [x] Applied via \`--wasm-runtime-overrides\` to a running partner chain stuck at block #14661 for 17h
- [x] Block production resumed immediately at #14662; finalization advanced within ~30s
- [x] Committee rotation at the next epoch boundary (with healthy IDP data) worked normally
- [ ] Add unit test covering the \`None\` branch in create_inherent / check_inherent / is_inherent_required (happy to add in this PR if desired)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


realdecimalist:
Closing this PR — it was opened prematurely as part of an automated recovery operation. The underlying bug report is real (we hit the `.expect("Validator inherent data must be provided")` panic on Materios preprod running v1.5.1, chain stalled for 17h until we applied a local patch via `--wasm-runtime-overrides`), but this PR is bloated with 80 unrelated files from v1.5.1→master drift in our fork, which makes it unreviewable.

Plan:
1. We'll open an **issue first** to discuss the design question: the panic appears intentional per the pallet docstring ("Setting incorrect main chain script values will result in stalling block production indefinitely, requiring a network-wide roll-back") but we hit it from a *transient* IDP-None at a sc-epoch-vs-mc-epoch boundary (Materios uses 1h sc-epochs against Cardano preprod's 5-day mc-epochs). The transient vs. misconfig distinction seems worth discussing before we slap on a behavioral change.
2. If there's appetite for a fix, we'll rebase a **minimal PR** against current master (`toolkit/committee-selection/pallet/src/lib.rs` — the file has moved since v1.5.1). The actual logical diff is ~4 functions; the other 80 files here are noise.

Apologies for the noise in the meantime. Full root-cause analysis + patch details available on request.

